PR #4, opened 2022-01-19
Add CircleCI setup

MerelTheisenQB:
This PR is the complete set of changes required to have CCI builds for all plugins in `kedro-plugins`. 

It has the changes from https://github.com/kedro-org/kedro-plugins/pull/2 with the addition of [`path-filtering`](https://circleci.com/developer/orbs/orb/circleci/path-filtering) which allows you to only run builds for the parts of the repo that have changed. I followed the steps [here](https://circleci.com/docs/2.0/configuration-cookbook/#execute-specific-workflows-or-steps-based-on-which-files-are-modified) to get this to work. 

I'll mark https://github.com/kedro-org/kedro-plugins/pull/2 as `draft` so you can use it as a reference how I came to the changes here, but I will only be merging be this PR.


.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+
+  setup_requirements:
+    parameters:
+      plugin:
+        type: string
+    steps:
+      - run:
+          name: Install pip setuptools
+          command: |
+            make install-pip-setuptools
```
AntonyMilneQB, line 49: Very minor, applies in several places: I'd just put this inline when there's only one line in command.
```suggestion
          command: install-pip-setuptools
```

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - run:
+          name: Install kedro and test requirements
+          command: |
+            cd <<parameters.plugin>>
+            pip install git+https://github.com/kedro-org/kedro
```
AntonyMilneQB, line 54: Maybe just to be explicit?
```suggestion
            pip install git+https://github.com/kedro-org/kedro@main
```

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+            pip install -r test_requirements.txt
+      - run:
+          name: Install pre-commit hooks
+          command: |
+            cd <<parameters.plugin>>
+            pre-commit install --install-hooks
+            pre-commit install --hook-type pre-push
```
AntonyMilneQB, line 61: In the Makefile we have
```
install-test-requirements:
	pip install -r test_requirements.txt

install-pre-commit: install-test-requirements
	pre-commit install --install-hooks
```

So you could do this just by modifying the Makefile so it takes `$(plugin)` and then calling `make install-pre-commit` instead of writing out the stages again here (and we don't need the `pre-push` one anywhere I think). Similar to how you are doing with `lint`.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+    parameters:
+      plugin:
+        type: string
+    steps:
+      - run:
+          name: Install Kedro plugins dependencies
+          command: |
+            conda activate kedro_plugins
+            cd <<parameters.plugin>>
+            python -m pip install -U pip setuptools wheel
```
AntonyMilneQB, line 104: This is just `make install-pip-setuptools`.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - run:
+          name: Install pip setuptools
+          command: |
+            make install-pip-setuptools
+      - run:
+          name: Install kedro and test requirements
+          command: |
+            cd <<parameters.plugin>>
+            pip install git+https://github.com/kedro-org/kedro
+            pip install -r test_requirements.txt
```
AntonyMilneQB, line 55: Do we want `-U` here like on the windows setup?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - win_setup_conda:
+          python_version: <<parameters.python_version>>
+      - win_setup_requirements:
+          plugin: <<parameters.plugin>>
+      - run:
+          # e2e tests are not currently runnable on CircleCI on Windows as
+          # those require the ability to run Linux containers:
+          # "The Windows executor currently only supports Windows containers.
+          # Running Linux containers on Windows is not possible for now"
+          # (from https://circleci.com/docs/2.0/hello-world-windows/)
```
AntonyMilneQB, line 126: Which tests don't work on windows currently? 🤔  I guess `kedro-telemetry` should work?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+          name: Run unit tests
+          command: |
+            conda activate kedro_plugins
+            pytest <<parameters.plugin>>\tests
```
AntonyMilneQB, line 130: Maybe `make test`?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - run:
+          name: Install Kedro plugins dependencies
+          command: |
+            conda activate kedro_plugins
+            cd <<parameters.plugin>>
+            python -m pip install -U pip setuptools wheel
+            pip install git+https://github.com/kedro-org/kedro@main
+            pip install -r test_requirements.txt -U
+
+  win_tests:
```
AntonyMilneQB, line 108: I think this should go in `jobs` rather than `commands`.

AntonyMilneQB (commented):
Still looking through the PR, will continue reviewing tomorrow! So far looks really great though 👍  👍

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      python_version:
+        type: string
+      plugin:
+        type: string
+    machine:
+      # Don't use 2018 image: https://discuss.circleci.com/t/24639/18
+      image: circleci/classic:201711-01
+      docker_layer_caching: true
+    environment:
+      python_version: <<parameters.python_version>>
```
AntonyMilneQB, line 144: What is this needed for?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+    steps:
+      - setup:
+          python_version: <<parameters.python_version>>
+          plugin: <<parameters.plugin>>
+      - run:
+          name: Run all end to end tests
```
AntonyMilneQB, line 171:
```suggestion
          name: Run e2e tests
```
(just to match the unit tests description)

.circleci/continue_config.yml
AntonyMilneQB: Now that I see what's in `jobs`, I think this should just replace `win_build`.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - setup_conda:
+          python_version: <<parameters.python_version>>
+      - setup_requirements:
+          plugin: <<parameters.plugin>>
+
+  # Windows specific commands
+  win_setup_conda:
+    # Miniconda3 is pre-installed on the machine:
+    # https://circleci.com/docs/2.0/hello-world-windows
+    description: Setup conda
```
AntonyMilneQB, line 80:
```suggestion
```

Not a biggy, but personally I'd remove these `description`s throughout. They don't actually appear anywhere in CircleCI, and all the commands have descriptive names already (which do appear in CircleCI).

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+  # when pipeline parameter, run-build-kedro-telemetry is true, the
+  # kedro-telemetry job is triggered.
+  kedro-telemetry:
+    when: << pipeline.parameters.run-build-kedro-telemetry >>
+    jobs:
+      - unit_tests:
+          matrix:
+            parameters:
+              python_version: [ "3.6", "3.7", "3.8" ]
+              plugin: [ "kedro-telemetry" ]
```
AntonyMilneQB, line 224:
```suggestion
              python_version: ["3.6", "3.7", "3.8"]
              plugin: ["kedro-telemetry"]
```
Don't know why but I find all all these spaces upsetting 😅  Obviously just nitpicking though...

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - win_tests:
+          python_version: <<parameters.python_version>>
+          plugin: <<parameters.plugin>>
+
+
+workflows:
+  # when pipeline parameter, run-build-kedro-telemetry is true, the
+  # kedro-telemetry job is triggered.
+  kedro-telemetry:
+    when: << pipeline.parameters.run-build-kedro-telemetry >>
```
AntonyMilneQB, line 218:
```suggestion
    when: <<pipeline.parameters.run-build-kedro-telemetry>>
```

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+      - unit_tests:
+          matrix:
+            parameters:
+              python_version: [ "3.6", "3.7", "3.8" ]
+              plugin: [ "kedro-telemetry" ]
+      - win_build:
+          matrix:
+            parameters:
+              python_version: [ "3.6", "3.7", "3.8" ]
+              plugin: [ "kedro-telemetry" ]
```
AntonyMilneQB, line 229: Given that `plugin` isn't in the matrix here I wonder if you could just do
```suggestion
          matrix:
            parameters:
              python_version: [ "3.6", "3.7", "3.8" ]
           parameters:
             plugin: kedro-telemetry
```

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+    environment:
+      python_version: <<parameters.python_version>>
+    steps:
+      - setup:
+          python_version: <<parameters.python_version>>
+          plugin: <<parameters.plugin>>
+      - run:
+          name: Run pylint and flake8
+          command: |
+            make plugin=<<parameters.plugin>> lint
```
AntonyMilneQB, line 194: Since we only ever use `python_version = 3.8` here, I'd actually just hard code it in rather than parametrising it. Up to you though.

```suggestion
  lint:
    parameters:
      plugin:
        type: string
    machine:
      # Don't use 2018 image: https://discuss.circleci.com/t/24639/18
      image: circleci/classic:201711-01
      docker_layer_caching: true
    environment:
      python_version: 3.8
    steps:
      - setup:
          python_version: 3.8
          plugin: <<parameters.plugin>>
      - run:
          name: Run pylint and flake8
          command: |
            make plugin=<<parameters.plugin>> lint
```

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+              plugin: [ "kedro-telemetry" ]
+      - win_build:
+          matrix:
+            parameters:
+              python_version: [ "3.6", "3.7", "3.8" ]
+              plugin: [ "kedro-telemetry" ]
+      - lint:
+          matrix:
+            parameters:
+              python_version: [ "3.8" ]
```
AntonyMilneQB, line 233: Again I don't think this single value needs a matrix. In fact if `lint` was hardcoded with `python_version = 3.8` you wouldn't need anything beyond just `- lint` here.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,284 @@
...
+  # Windows jobs
+  win_build:
+    parameters:
+      python_version:
+        type: string
+      plugin:
+        type: string
+    executor:
+      name: win/default
+    working_directory: ~/repo
```
AntonyMilneQB, line 205: Is this `working_directory` line actually needed? It looks a bit weird 🤔

AntonyMilneQB:
Overall this looks awesome! I was expecting something that was an MVP but this looks pretty close to an ideal solution, amazing work! Love the use of the matrix and parameters and all that - really pleased to see we can do per-directory selective workflows without too much faff. 🎉  In addition to my many minor comments, there's a few other things I think we should still consider.

### pip freeze
Might be nice to add a `pip freeze` step like [this](https://github.com/kedro-org/kedro/blob/main/.circleci/config.yml#L61-L63) so we can easily debug things.

### Testing
By pushing to this branch and making sure only the right workflows run, making sure that the environment has the right requirements file installed (e.g. using `pip freeze` step), etc.

### How to deal with pre-commit
Do the pre-commit hooks actually still work without the file being at the root of the repo? My guess based on [this](https://stackoverflow.com/questions/64773189/can-the-python-pre-commit-tool-consider-a-subdirectory-as-a-base-of-the-project) is maybe not 🤔 

### How to deal with Makefile
I'm not sure whether the Makefile should be per-project or at the repo root. It's definitely less repetitive the way you've done it, but when I'm working on a single plugin I would sort of expect the Makefile to live inside the directory for that individual plugin. Kind of like how each plugin has its own .pre-commit-config.yaml, pyproject.toml etc.

### How to deal with parameters
3 boolean parameters like `run-build-kedro-telemetry` is how the CircleCI guide suggested doing this but actually I wonder if there's a neater way. Why can't we just use the name of the plugin directory as the parameter?
```
### config.yml
# 3-column, whitespace-delimited mapping. One mapping per
# line:
# <regex path-to-test> <parameter-to-set> <value-of-pipeline-parameter>
mapping: |
  kedro-telemetry/.* plugin-name telemetry
  kedro-docker/.* plugin-name docker
  kedro-airflow/.* plugin-name airflow

### continue_config.yml
parameters:
  plugin-name:
    type: enum
    enum: [telemetry, docker, airflow]
```

The advantage of this is you might then need a only single workflow that works for all 3 plugins instead of needing to switch between 3 different workflows dependent on the boolean:

```
workflow_name:
    jobs:
      - unit_tests:
          matrix:
            parameters:
              python_version: [ "3.6", "3.7", "3.8" ]
              plugin: <<pipeline.parameters.plugin-name>>
      - e2e_tests:
          matrix:
            parameters:
              python_version: [ "3.6", "3.7", "3.8" ]
              plugin: <<pipeline.parameters.plugin-name>>
      - win_build:
          matrix:
            parameters:
              python_version: [ "3.6", "3.7", "3.8" ]
              plugin: <<pipeline.parameters.plugin-name>>
      - lint:
          matrix:
            parameters:
              python_version: [ "3.8" ]
              plugin: <<pipeline.parameters.plugin-name>>
```

Related to this we are doing `make plugin=$(plugin_name) command` quite a bit. I wonder if this would be neater as an environment variable:
```
environment:
   plugin: <pipeline.parameters.plugin-name>>
```
And then you can call `make command` directly without needing to inject the plugin name into every `make` command individually. Or if the Makefiles do move into each individual directory maybe you'd just need to `cd <<pipeline.parameters.plugin-name>>` or change the `working_directory` to the plugin name parameter and then just run the `make command`.

.circleci/continue_config.yml
MerelTheisenQB: Nope not needed!

.circleci/continue_config.yml
AntonyMilneQB: Yeah, definitely not one for now. Was just curious 🙂

.circleci/continue_config.yml
MerelTheisenQB: ahh I forgot that you need to install `chocolatey` to run `make` on Windows. If we only call `make` in this step, I'd suggest just leaving at is and not add any more installs.

.circleci/continue_config.yml
MerelTheisenQB: This is sadly not as straightforward as it seems. The `install-pre-commit` step references another `make` command `install-test-requirements` and so when you add `cd $(plugin)` it complains about not being able to find `install-pre-commit`. I'm not entirely sure why but I'm guessing it has to do with interchanging bash commands and referencing other `make` steps. 

I'm leaving this as is for now.

MerelTheisenQB:
> ### How to deal with pre-commit
> Do the pre-commit hooks actually still work without the file being at the root of the repo? My guess based on [this](https://stackoverflow.com/questions/64773189/can-the-python-pre-commit-tool-consider-a-subdirectory-as-a-base-of-the-project) is maybe not 🤔

Yeah.. I fiddled quite a bit with all the `.pre-commit-config.yaml` files in the plugins to make them run properly. There might be a way to have just one pre-commit setup to work with all repos. 

> ### How to deal with Makefile
> I'm not sure whether the Makefile should be per-project or at the repo root. It's definitely less repetitive the way you've done it, but when I'm working on a single plugin I would sort of expect the Makefile to live inside the directory for that individual plugin. Kind of like how each plugin has its own .pre-commit-config.yaml, pyproject.toml etc.

I totally agree! I was struggling to find an optimal solution for keeping the setup clean but also having it all run correctly for the different plugins. I looked at the [PmPx](https://github.com/McK-Internal/asset-pmpx) and [Brix](https://github.com/McK-Internal/brix) codebases for inspiration but they each did it slightly differently. I'll add a follow up ticket to look at whether we need to unify or split the config for each of the plugins. 

> ### How to deal with parameters
> 3 boolean parameters like `run-build-kedro-telemetry` is how the CircleCI guide suggested doing this but actually I wonder if there's a neater way. Why can't we just use the name of the plugin directory as the parameter?
> > Edit: just thought of a problem with this suggestion... What happens if I modify both telemetry and docker at the same time? I guess we need the 3 boolean variables so we can trigger workflows for multiple plugins simultaneously if required 🤔

I don't think there's a way of achieving the same setup with just one parameter, exactly because of what you said in your later comment: how could you trigger multiple plugins if there's just one enum parameter. 


> Related to this we are doing `make plugin=$(plugin_name) command` quite a bit. I wonder if this would be neater as an environment variable:
> 
> ```
> environment:
>    plugin: <pipeline.parameters.plugin-name>>
> ```
> 
> And then you can call `make command` directly without needing to inject the plugin name into every `make` command individually. Or if the Makefiles do move into each individual directory maybe you'd just need to `cd <<pipeline.parameters.plugin-name>>` or change the `working_directory` to the plugin name parameter and then just run the `make command`.

I'll leave this as is for now and add it as something to look into on the follow up ticket.

SajidAlamQB (approved):
Amazing work! 💥

.circleci/continue_config.yml
```diff
@@ -0,0 +1,249 @@
...
+            sudo rm -rf .pyenv/ /opt/circleci/.pyenv/
+            # Download and install miniconda
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh  > miniconda.sh
+            bash miniconda.sh -b -p $HOME/miniconda
+            # Create an anaconda virtualenv for python_version and make that the default python interpreter
+            echo ". /home/circleci/miniconda/etc/profile.d/conda.sh" >> $BASH_ENV
+            echo "conda deactivate; conda activate kedro_plugins" >> $BASH_ENV
+            . /home/circleci/miniconda/etc/profile.d/conda.sh
+            conda create --name kedro_plugins python=<<parameters.python_version>> -y
+            source $BASH_ENV
```
idanov, line 39:
```suggestion
            echo ". /home/circleci/miniconda/etc/profile.d/conda.sh" >> $BASH_ENV
            echo "conda deactivate; conda activate kedro_plugins" >> $BASH_ENV
            . /home/circleci/miniconda/etc/profile.d/conda.sh
            conda create --name kedro_plugins python=<<parameters.python_version>> -y
            source $BASH_ENV
```

This one seems mixed up. Why is `$BASH_ENV` overwritten before being used at all? And then why do we not just call `conda deactivate; conda activate kedro_plugins` directly instead of echoing it to a variable?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,249 @@
...
+    parameters:
+      plugin:
+        type: string
+    steps:
+      - run:
+          name: Install Kedro plugins dependencies
+          command: |
+            conda activate kedro_plugins
+            cd <<parameters.plugin>>
+            make install-pip-setuptools
```
idanov, line 103: Does this work after we move one directory down? Didn't you move the `Makefile` at the root directory?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,249 @@
...
+          plugin: "kedro-docker"
+          matrix:
+            parameters:
+              python_version: ["3.6", "3.7", "3.8"]
+      - win_unit_tests:
+          plugin: "kedro-docker"
+          matrix:
+            parameters:
+              python_version: ["3.6", "3.7", "3.8"]
+      - lint:
```
idanov, line 226: Does it make sense to reorder this a bit and put the `lint` first? I believe it should be the fastest one, so if it fails, we shouldn't wait for all the tests to run to fail the whole job.

kedro-airflow/.pre-commit-config.yaml
```diff
@@ -32,21 +36,21 @@ repos:
         name: "Quick PyLint on kedro_airflow/*"
         language: system
         types: [file, python]
-        files: ^kedro_airflow/
+        files: ^kedro-airflow/kedro_airflow/
```
idanov, line 39: Why is this one needed? Isn't the directory already changed to `kedro-airflow` when we call the `pre-commit` hook?

idanov (commented):
When calling the `pre-commit` hook from the project folder, is it really needed to prepend all paths in the `.pre-commit-config.yaml` with the folder containing the project?

.circleci/continue_config.yml
MerelTheisenQB: All builds will just kick-off at the same time, so the `lint` build can still finish/fail first.

.circleci/continue_config.yml
MerelTheisenQB: I haven't written the whole setup from scratch, so in all honesty I've been a bit lazy and left some things as they were and focussed more and getting the builds to run properly. If it's okay with you I'd like to get this merged and will then focus on clean up, because it's already quite a lot to review.

MerelTheisenQB:
> When calling the `pre-commit` hook from the project folder, is it really needed to prepend all paths in the `.pre-commit-config.yaml` with the folder containing the project?

As I mentioned to Antony as well, I fiddled quite a bit with the `.pre-commit-config.yaml` files to get them all working and agree that this is probably not the ultimate setup we'd like. I already created a [ticket](https://jira.quantumblack.com/browse/KED-3056) to improve this setup and focus on configuration in general (e.g. should we have just one `Makefile` or one per plugin?). The aim was to have this setup done before the LF launch, so that's why I didn't immediately tackle everything.

.circleci/continue_config.yml
MerelTheisenQB: good catch!! 🦅  Turns out this doesn't work at all because `make` doesn't work on windows unless you install chocolately

Makefile
```diff
@@ -9,10 +9,13 @@ install-pip-setuptools:
 	python -m pip install -U pip setuptools wheel
 
 lint:
-	pre-commit run -a --hook-stage manual
+	cd $(plugin) && pre-commit run -a --hook-stage manual
 
 test:
-	pytest -vv tests
+	pytest -vv $(plugin)/tests
```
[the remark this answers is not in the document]
MerelTheisenQB: Nope no particular reason. I'll make it consistent with the other commands.

.circleci/continue_config.yml
MerelTheisenQB: I have split this chunk in separate run steps so hopefully it's more clear now what's happening.

PR #6, opened 2022-02-25
Fix connection error issue in telemetry

datajoely:
Two related GH issues:

https://github.com/kedro-org/kedro/issues/1289
https://github.com/kedro-org/kedro/issues/1249

Today - if consent provided, but the user has no access to the internet  an error is raised. This PR introduces a small exception handler that simply raises warnings whenever a `requests.exceptions.RequestException` base exception is raised.

### DIFF
--- a/kedro-telemetry/tests/test_masking.py
+++ b/kedro-telemetry/tests/test_masking.py
@@ -29,6 +29,7 @@ DEFAULT_KEDRO_COMMANDS = [
     "ipython",
     "jupyter",
     "lint",
+    "micropkg",
     "new",
     "package",
     "pipeline",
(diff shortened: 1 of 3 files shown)


SajidAlamQB (approved):
LGTM 👍

kedro-telemetry/tests/test_masking.py
```diff
@@ -29,6 +29,7 @@
     "ipython",
     "jupyter",
     "lint",
+    "micropkg",
```
lorenabalan, line 32: Mm have you merged latest `main` in your branch? I updated this in the last PR. 🤔

datajoely: yes it's picked up now!

PR #8, opened 2022-03-09
Modify Makefile per plugin

lorenabalan:
Signed-off-by: lorenabalan <lorena.balan@quantumblack.com>

Update make instructions to make plugin release easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
 package:
-	rm -Rf dist
+	cd $(plugin);\
+	rm -Rf dist;\
 	python setup.py sdist bdist_wheel
 
 install: package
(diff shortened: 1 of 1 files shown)


AntonyMilneQB (approved):
👍 

Just curious, why do we sometimes need `;\` instead of linebreaks just working by themselves?

MerelTheisenQB (approved):
👍

lorenabalan:
> Just curious, why do we sometimes need `;\` instead of linebreaks just working by themselves?

So that the `cd <here>` propagates to the next instructions (and avoid long `&&` chains). https://stackoverflow.com/a/1789616

Issue #11, opened 2022-03-24
Telemetry breaks everything if there's no write permissions


AntonyMilneQB:
## Description
As seen on databricks repos, which currently do not have write access, running any kedro command in a directory with no write access will raise an error with kedro-telemetry which then stops kedro from running everything.

This isn't the first time that an exception raised by kedro-telemetry has stopped kedro commands from working rather than just raising a warning and carrying on. I wonder if we should wrap the whole plugin in a `try/except` to catch any exceptions we don't anticipate. A broken kedro-telemetry shouldn't prevent users from using kedro.

## Steps to reproduce
1. Go to https://gitpod.io/#https://github.com/kedro-org/kedro
2. `pip install kedro-telemetry`
3. `chmod a=rx /workspace/project`
4. `kedro info`

This will give the following exception and not execute the kedro command at all:
```
PermissionError: [Errno 13] Permission denied: '/workspace/project/.telemetry'
```

MerelTheisenQB:
"A broken kedro-telemetry shouldn't prevent users from using kedro." -> No error in `kedro-telemetry` should prevent a user from running Kedro, so as part of this fix ensure that other errors won't break Kedro either.

PR #21, opened 2022-04-13
Prevent Telemetry breaking Kedro when no write permissions

SajidAlamQB:
## Description

Related issue: https://github.com/kedro-org/kedro-plugins/issues/11

When having no write access in a project, running any kedro command will raise an error with kedro-telemetry which then stops kedro from running everything. A broken kedro-telemetry shouldn't prevent users from using kedro.

## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,7 +1,10 @@
 # Release 0.2.0
+
+## Bug fixes and other changes
 * Add compatibility with `kedro` 0.18.0
 * Add compatibility with Python 3.9 and 3.10
 * Remove compatibility with Python 3.6
+* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines. 
 
 # Release 0.1.4
 
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -4,7 +4,7 @@
 * Add compatibility with `kedro` 0.18.0
 * Add compatibility with Python 3.9 and 3.10
 * Remove compatibility with Python 3.6
-* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines. 
+* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines.
 
 # Release 0.1.4
 
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,10 +1,14 @@
+# Release 0.2.1
+
+## Bug fixes and other changes
+* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines.
+
 # Release 0.2.0
 
 ## Bug fixes and other changes
 * Add compatibility with `kedro` 0.18.0
 * Add compatibility with Python 3.9 and 3.10
 * Remove compatibility with Python 3.6
-* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines.
 
 # Release 0.1.4
 
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -180,21 +180,28 @@ def _is_valid_syntax(telemetry: Any) -> bool:
 
 
 def _confirm_consent(telemetry_file_path: Path) -> bool:
-    with telemetry_file_path.open("w") as telemetry_file:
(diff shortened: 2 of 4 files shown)


AntonyMilneQB (commented):
This is good, but I think we should err on the side of caution here and also wrap everything inside `before_command_run` in a `try/except` which logs a warning if something goes wrong.

It's a bit horrible to wrap so much code in `try`, but potentially there will be another exception raised by something we didn't anticipate (e.g. in the masking code), so best to be on the safe side and just catch all errors now. Otherwise we'll just have to more in future if anything breaks.

It would be nice to have a couple of tests for this also:
1. if `yaml.dump` (used inside `_config_consent`) raises an error, we get a warning message (mock out `yaml.dump` and give it a `side_effect`)
2. if something else in `before_command_run` (e.g. `_mask_kedro_cli`) raises an error, we get a warning message

Currently the first of these tests should pass but the second would fail (write the tests first here to make sure they're working right!).

noklam (commented):
Agree with Antony that we should wrap the entrypoint `before_command_run` for being more conservative here. I don't have much to add.

noklam (approved):
LGTM!

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -180,20 +186,28 @@ def _is_valid_syntax(telemetry: Any) -> bool:
...
+            if click.confirm(confirm_msg):
+                yaml.dump({"consent": True}, telemetry_file)
+                click.secho("You have opted into product usage analytics.", fg="green")
+                return True
+            yaml.dump({"consent": False}, telemetry_file)
+            return False
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning(
+            "Failed to confirm consent. Exception: %s",
+            exc,
```
MerelTheisenQB, line 209: It might be useful to add that no data was sent to heap when this exception happens.

kedro-telemetry/tests/test_plugin.py
```diff
@@ -192,6 +196,19 @@ def test_before_command_run_anonymous(self, mocker, fake_metadata):
...
+            "kedro_telemetry.plugin._check_for_telemetry_consent", return_value=True
+        )
+        mocked_heap_call = mocker.patch(
+            "kedro_telemetry.plugin._send_heap_event", side_effect=Exception
+        )
+        telemetry_hook = KedroTelemetryCLIHooks()
+        command_args = ["--version"]
+
+        telemetry_hook.before_command_run(fake_metadata, command_args)
+        mocked_heap_call.assert_called()
```
MerelTheisenQB, line 210: I would explicitly check that the right exception message gets thrown. Also in the test below.

MerelTheisenQB (commented):
Don't forget to update the release notes.

kedro-telemetry/RELEASE.md
```diff
@@ -1,7 +1,10 @@
 # Release 0.2.0
+
+## Bug fixes and other changes
 * Add compatibility with `kedro` 0.18.0
 * Add compatibility with Python 3.9 and 3.10
 * Remove compatibility with Python 3.6
+* `kedro telemetry` raising errors will no longer stop `kedro` running pipelines.
```
MerelTheisenQB, line 7: This will be part of a new release, so you'll have to add a new section for `0.2.1` and add it there.

kedro-telemetry/tests/test_plugin.py
```diff
@@ -258,3 +280,16 @@ def test_check_for_telemetry_consent_file_invalid_yaml(self, mocker, fake_metada
...
+    def test_confirm_consent_yaml_dump_error(self, mocker, fake_metadata, caplog):
+        Path(fake_metadata.project_path, "conf").mkdir(parents=True)
+        telemetry_file_path = fake_metadata.project_path / ".telemetry"
+        mocker.patch("yaml.dump", side_effect=Exception)
+
+        assert not _confirm_consent(telemetry_file_path)
+
+        msg = (
+            "Failed to confirm consent. No data was sent to Heap. Exception: "
+            "pytest: reading from stdin while output is captured!  Consider using `-s`."
```
MerelTheisenQB, line 293: It should be fine just checking this part "Failed to confirm consent. No data was sent to Heap." is included in the message.

MerelTheisenQB (approved):
LGTM! 👍

Issue #26, opened 2022-05-15
Extend Kedro-Airflow CLI functionality with optional jinja template


fdroessler:
## Description

The default Jinja2-Template works well as a base to get started on using kedro with Airflow. However, the fact that a few things are hardcoded in the plugin/template makes it hard to modify a variety of things. I would propose a couple of adjustments in order to allow for more customisability:
- Provide CLI Arguments for folder and name of jinja2 Templates
- Prefix/Postfix the name of the resulting dag files with the pipeline name if a pipeline was given

I have already implemented those in a branch and was curious if this would be a welcomed PR.

## Context
In a recent project I found myself constantly having to adjust the resulting dag files as I was experimenting with a variety of things. I ended up replacing the template file in the `site-packages` folder at some point which struck me as too hacky. The above changes would allow users to provide their own Jinja2 Templates as well as generating Airflow Dags for multiple pipelines without having to manually rename the output files.

## Possible Implementation
Add two more click options (names just examples):
```
@click.option("-jd", "--jinja-dir", "jinja_template_dir", default=str(Path(__file__).parent))
@click.option("-j", "--jinja-file", "jinja_template", default="airflow_dag_template.j2")
```
Modify the loader and template respectively:
```
loader = jinja2.FileSystemLoader(jinja_template_dir)
jinja_env = jinja2.Environment(autoescape=True, loader=loader, lstrip_blocks=True)
jinja_env.filters["slugify"] = slugify
template = jinja_env.get_template(jinja_template)
```
Adjust the dag_filename if necessary:
```
if pipeline_name != "__default__":
    dag_filename = f"{package_name}_{pipeline_name}_dag.py"
```

Looking forward to some feedback on this suggestions.

PR #30, opened 2022-05-17
Feature/additional cli args

fdroessler:
## Description
This PR address #26 

## Development notes

- Added two new cli arguments (`--jinja-file` and `--jinja-dir`) to provide an alternate Jinja directory and name of the jinja file.
- Renaming the dag output file to include the name of the pipeline if the pipeline is not the default one
- Added tests

## Checklist


kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -33,18 +33,30 @@ def airflow_commands():
...
     jinja_env.filters["slugify"] = slugify
-    template = jinja_env.get_template("airflow_dag_template.j2")
+    template = jinja_env.get_template(jinja_template)
 
     package_name = metadata.package_name
-    dag_filename = f"{package_name}_dag.py"
+    if pipeline_name != "__default__":
+        dag_filename = f"{package_name}_{pipeline_name}_dag.py"
+    else:
+        dag_filename = f"{package_name}_dag.py"
```
noklam, line 59: I think it is a good idea to have the flexibility to the generate dag file name. I have two comments here.

1. It breaks the backward compatibility.
2. I don't have strong preference about the file name, but may as well just let user define whatever name they find meaningful?

kedro-airflow/tests/test_plugin.py
```diff
@@ -1,26 +1,45 @@
...
+        ("hello_world_ds_dag", "ds", ["airflow", "create", "--pipeline", "ds"]),
+        # Test execution with different dir and filename for Jinja2 Template
+        (
+            "hello_world_dag",
+            "__default__",
+            ["airflow", "create", "-jd", ".", "-j", "airflow_dag.j2"],
+        ),
+    ],
+)
+def test_create_airflow_dag(
```
noklam, line 29: Great that you have the tests covered!

noklam (commented):
Thank you for your contribution! Awesome tests, I have left some comments.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -33,18 +33,30 @@ def airflow_commands():
     type=click.Path(writable=True, resolve_path=True, file_okay=False),
     default="./airflow_dags/",
 )
+@click.option(
+    "-jd", "--jinja-dir", "jinja_template_dir", default=str(Path(__file__).parent)
+)
+@click.option("-j", "--jinja-file", "jinja_template", default="airflow_dag_template.j2")
```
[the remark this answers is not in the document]
fdroessler: Actually I was thinking the same.. not sure why I did this in the first place :D

kedro-airflow/tests/conftest.py
```diff
@@ -15,7 +16,10 @@
 @fixture(name="cli_runner")
 def cli_runner():
     runner = CliRunner()
+    owd = Path().cwd()
```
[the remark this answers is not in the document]
fdroessler: I think `Path().cwd()` just returns the current working directory. I use this here to the copy the jinja2 template into the test dir under a different name so I can use it to test the cli option (`--jinja-file`) ensuring that it works with a copy of the original file under a different path.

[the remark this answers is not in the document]
noklam: I wonder if it would just work with Path("kedro_airflow/trmplate.j2")? Since the owd is only used once anyway.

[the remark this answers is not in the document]
fdroessler: Isn't that similar to what is done in the metadata fixture?

fdroessler:
@noklam couple of observations:

- e2e tests have a dependency on python-starters (pandas-iris) therefore I had to adjust the behave tests. It might be worth triggering the whole test suit in some cases?
- Changing the `-t` option now kind of breaks backwards compatibility if people use it, is that a concern for you?
- unit tests under windows are run differently compared to linux. I.e. Linux runs using the makefile which first cds into the plugin dir (e.g. kedro-airflow) and then executes the tests while under windows they are run from the kedro-plugin dir instead. This caused me to have a slightly ugly fix in the `conftest.py` maybe the circleci can be adjusted so that unit tests are run in the same way?

kedro-airflow/features/airflow.feature
```diff
@@ -8,21 +8,20 @@ Feature: Airflow
...
     And I should get a message including "split"
-    And I should get a message including "train"
 
   Scenario: Run Airflow task locally with latest Kedro
     Given I have installed kedro version "latest"
     And I have prepared a config file
     And I have run a non-interactive kedro new
     And I have prepared a data catalog
-    And I have executed the kedro command "airflow create -t ../airflow/dags/"
+    And I have executed the kedro command "airflow create -t ../airflow/dags/project_dummy_dag.py"
```
noklam, line 23: It isn't immediately obvious to me, why does it has to break, shouldn't the default work as the same?

kedro-airflow/features/airflow.feature
```diff
@@ -8,21 +8,20 @@ Feature: Airflow
...
     And I have installed the kedro project package
     When I execute the airflow command "tasks test project-dummy split 2016-06-01T00:00:00+00:00"
     Then I should get a successful exit code
-    And I should get a message including "Loading data from `params:"
+    And I should get a message including "Loading data from `parameters"
```
noklam, line 27: This is fine - as we did updated the starter, thank you for figuring this out!

kedro-airflow/features/airflow.feature
fdroessler: Default is fine but if there is previous code that uses the -t option to provide a directory instead of dir+filepath than this would break now. I could check if its a dir or a file and handle those cases in code 🤔

noklam: I am slightly inclined to just have --target, whether it is path/dir should just work. But I would love to hear comments from others. And definitely less breaking changes are better.

kedro-airflow/README.md
```diff
@@ -47,3 +47,11 @@ Please visit the guide to [deploy Kedro as a Python package](https://kedro.readt
 #### What if my DAG file is in a different directory to my project folder?
 
 By default the generated DAG file is configured to live in the same directory as your project as per this [template](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-airflow/kedro_airflow/airflow_dag_template.j2#L44). If your DAG file is located in a different directory to your project, you will need to tweak this  manually after running the `kedro airflow create` command.
+
+#### What if I want to use another Jinja2 Template in a different location
+
+You can use the additional command line argument (`--jinja-file/-j`) that allows the user to provide an alternative path to the Jinja2 template. Note that these files have to accept the same variables as those used in the original Jinja2-Template.
```
AntonyMilneQB, line 53:
```suggestion

#### What if I want to use a different Jinja2 template?

You can use the additional command line argument `--jinja-file` (alias `-j`) to provide an alternative path to a Jinja2 template. Note that these files have to accept the same variables as those used in the [default Jinja2 template](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-airflow/kedro_airflow/airflow_dag_template.j2).
```

kedro-airflow/README.md
```diff
@@ -47,3 +47,11 @@ Please visit the guide to [deploy Kedro as a Python package](https://kedro.readt
...
+
+```bash
+kedro airflow create --jinja-file ./custom/template.j2
```
AntonyMilneQB, line 56:
```suggestion
kedro airflow create --jinja-file=./custom/template.j2
```
(just a small convention that we very roughly follow: usually `=` when using `--flag` and no equals sign for a short alias `-f`)

kedro-airflow/RELEASE.md
```diff
@@ -1,3 +1,8 @@
+# Release 0.5.1
+* Add additional cli argument `--jinja-file` to provide custom Jinja2 Templates
```
AntonyMilneQB, line 2:
```suggestion
* Add additional CLI argument `--jinja-file` to provide a path to a custom Jinja2 template
```

kedro-airflow/RELEASE.md
```diff
@@ -1,3 +1,8 @@
...
+* Change the `--target-dir` to `--target-path` so a users can define directory and filename of the output
+* Change e2e behave tests to adjust for changes in pandas-iris starter (no more train node and loading from parameters instead of params:)
```
AntonyMilneQB, line 4:
```suggestion
```
We don't typically put changes to things that aren't user-facing in the changelog.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -28,26 +28,43 @@ def airflow_commands():
...
     "target_path",
-    type=click.Path(writable=True, resolve_path=True, file_okay=False),
-    default="./airflow_dags/",
+    default="./airflow_dags/{package_name}_dag.py",
+)
+@click.option(
+    "-j",
+    "--jinja-file",
+    "jinja_template",
+    type=click.Path(exists=True, readable=True, resolve_path=True, file_okay=True),
```
AntonyMilneQB, line 39:
```suggestion
    type=click.Path(exists=True, readable=True, resolve_path=True, file_okay=True, dir_okay=False),
```
And then you should be able to get rid of the `if jinja_template.is_file` conditional also, since it will already be checked by `click.Path`.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -28,26 +28,43 @@ def airflow_commands():
...
+    "--target-path",
     "target_path",
-    type=click.Path(writable=True, resolve_path=True, file_okay=False),
-    default="./airflow_dags/",
+    default="./airflow_dags/{package_name}_dag.py",
+)
+@click.option(
+    "-j",
+    "--jinja-file",
+    "jinja_template",
```
AntonyMilneQB, line 38:
```suggestion
```
Small simplification: you don't need to name this variable explicitly if you just change to using `jinja_file` in the function.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -28,26 +28,43 @@ def airflow_commands():
...
 )
 @click.pass_obj
 def create(
-    metadata: ProjectMetadata, pipeline_name, env, target_path
-):  # pylint: disable=too-many-locals
+    metadata: ProjectMetadata,
+    pipeline_name,
+    env,
+    target_path,
+    jinja_template,
```
AntonyMilneQB, line 48:
```suggestion
    jinja_file,
```
(and then same renaming throughout the function)

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -28,26 +28,43 @@ def airflow_commands():
...
+):  # pylint: disable=too-many-locals,too-many-arguments
     """Create an Airflow DAG for a project"""
-    loader = jinja2.FileSystemLoader(str(Path(__file__).parent))
-    jinja_env = jinja2.Environment(autoescape=True, loader=loader, lstrip_blocks=True)
-    jinja_env.filters["slugify"] = slugify
-    template = jinja_env.get_template("airflow_dag_template.j2")
+    jinja_template = Path(jinja_template)
```
AntonyMilneQB, line 51:
```suggestion
    jinja_file = Path(jinja_file).resolve()
```
See https://stackoverflow.com/questions/42513056/how-to-get-absolute-path-of-a-pathlib-path-object for why `resolve` is preferred to `absolute`.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -28,26 +28,43 @@ def airflow_commands():
...
+    if jinja_template.is_file():
+        jinja_template = jinja_template.absolute()
+        loader = jinja2.FileSystemLoader(jinja_template.parent)
```
AntonyMilneQB, line 54: Does this work ok with a `Path` as argument? Just checking because previously it was converted to `str`.

kedro-airflow/tests/conftest.py
```diff
@@ -15,7 +16,10 @@
 @fixture(name="cli_runner")
 def cli_runner():
     runner = CliRunner()
+    owd = Path.cwd()
     with runner.isolated_filesystem():
+        fp = owd / "kedro_airflow/airflow_dag_template.j2"
+        copyfile(fp.resolve(), Path("./airflow_dag.j2").resolve())
```
AntonyMilneQB, line 22: What is this for? 🤔 

Edit: is this the thing you're trying to fix by altering the CircleCI config instead? If it is possible to do so that way then it would definitely be preferable I think.

AntonyMilneQB (commented):
This is great, thank you so much for the contribution! I've left a lot of small suggestions, but broadly speaking:
1. The `jinja-file` option is great, I just have some fussy comments to tidy the code a bit
1. The new `target_path` functionality makes less sense to me, but let me comment on that separately in the relevant thread
1. Thank you for fixing the tests! Much appreciated ⭐

kedro-airflow/features/airflow.feature
AntonyMilneQB: My thoughts on this are:
* being able to change output filename is a relatively small new feature, since it can already be achieved by running `kedro airflow create && mv ...` (unlike the `jinja-file` option, which can't currently be achieved so easily)
* as per @noklam, it's nice to do non-breaking changes where possible, especially when it's just to add a relatively minor feature. So we should not release in 0.5.1 a version that changes the name of the `--target-dir` flag to something else or which breaks current behaviour that this argument can take a directory

In an ideal world, Nok's idea of having a `--target` option that can take a directory or a filepath sounds like the right solution to me. The only breaking part of that is renaming the `--target-dir` flag since it's not just a directory any more. Technically we should do it as follows:
1. introduce in 0.5.1 a warning that the argument is changing name
2. change the argument name _in 0.6.0_

But this is a bit of a faff  for such a minor change to a plugin that not that many people use. (Can we check how many people use `kedro airflow create` with the `--target-dir` flag on telemetry? I think it's going to be very small... @datajoely) So I'd be tempted to just skip step 1 above.

Then `--target` can be a directory or filename. The default behaviour when no target is specified should remain as it currently is. If `--target` supplies a directory, the output file should be in that directory with the default `{package_name}_dag.py` name. If `--target` supplies a file, then save to that file (creating any parent directories that are required).

Having an (optional) user-specified filename also makes more sense to me than the initial proposal where the output file was labelled according to the pipeline name (e.g. why don't we also label it according to `env`?).

All this is becoming a bit more complicated a change so I'd also advise splitting it off into a separate PR and just focussing on the `jinja-file` addition here 🙂

fdroessler: I don't mind so much, happy to fork it off into a separate PR. I found being able to either provide a dir or filepath more intuitive than just a dir with a "default" filename and then using mv afterwards. @AntonyMilneQB @noklam would you still be happy to merge that PR relatively quickly or is it a too minor change?

fdroessler: "quickly" in terms of the PR not necessarily as a release :)

kedro-airflow/kedro_airflow/plugin.py
fdroessler: I think so, at least the checks are working and generating a valid DAG.

kedro-airflow/tests/conftest.py
fdroessler: No this is different. My reasoning here was that I wanted to provide a "real" different filepath to test `--jinja-file`. Therefore I am copying the template file into the test directory under a different filepath. I could of course just test the `--jinja-file` option with the original template file in the kedro_airflow directory but I wanted to be a bit more thorough. Maybe its not necessary 🤷

kedro-airflow/features/airflow.feature
AntonyMilneQB: It's a minor change but it does make sense as a new argument, so if you think it's useful then I'd be very happy to accept such a PR. @noklam is my above description of a `--target` flag that takes a file or a directory what you were thinking?

Also I suspect that once the PRs are merged we'll do a release pretty soon after anyway, unless we're waiting for any other particular changes to go in.

kedro-airflow/tests/test_plugin.py
```diff
@@ -9,18 +10,35 @@ def identity(arg):
     return arg
 
 
-def test_create_airflow_dag(mocker, cli_runner, metadata):
+@pytest.mark.parametrize(
+    "dag_name,pipeline_name,command",
+    [
+        # Test normal execution
+        ("hello_world_dag", "__default__", ["airflow", "create"]),
+        # Test execution with alternate pipeline and output name
```
AntonyMilneQB, line 18:
```suggestion
        # Test execution with alternate pipeline name
```

kedro-airflow/tests/conftest.py
AntonyMilneQB: Ahh I see, thanks for the explanation. This makes sense and I think is a good thing to do 👍  My only very minor comment would be that `owd` seems like a bit of an odd variable name - why not `cwd`?

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -33,15 +33,29 @@ def airflow_commands():
...
+    metadata: ProjectMetadata,
+    pipeline_name,
+    env,
+    target_path,
+    jinja_file,
+):  # pylint: disable=too-many-locals,too-many-arguments
     """Create an Airflow DAG for a project"""
-    loader = jinja2.FileSystemLoader(str(Path(__file__).parent))
+    jinja_file = Path(jinja_file).resolve()
+    jinja_file = jinja_file.absolute()
```
AntonyMilneQB:
```suggestion
```
No need for `absolute` if you're already doing `resolve` 🙂

AntonyMilneQB (approved):
This is great, thank you very much for doing it! Just a couple of very small comments but looks good.

noklam:
Sorry for the delay! I will review this tomorrow.

kedro-airflow/features/airflow.feature
noklam: I am thinking the same about the `--target` flag.

noklam (approved):
Nicely done! Thank you for your PR and sorry for blocking this PR.

PR #32, opened 2022-05-24
Change some of the logging message to `DEBUG`  in `kedro-telemetry`

noklam:
Signed-off-by: noklam <nok.lam.chan@quantumblack.com>

## Description

> There's two problems with how kedro-telemetry does logging at the moment:
> 
> 1. [Sometimes it uses `click.secho`](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-telemetry/kedro_telemetry/plugin.py#L57) and [sometimes it uses logging](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-telemetry/kedro_telemetry/plugin.py#L64). This is inconsistent.
> 2. Due to changes in [Improve logging setup in Kedro [parent issue] kedro#1461](https://github.com/kedro-org/kedro/issues/1461), as it stands, no kedro-telemetry logging message that is < WARNING will display now.
> 
> For point 1, technically speaking I think both these message should use `logging.debug` (not info) rather than `click.echo`. Since by default these messages should not show to a user (confirmed by @yetudada in sprint planning) then setting them to use `logging.debug` will make them disappear.
> 
> So all that should be needed here is modifying L64 and L57 to both use `logging.debug`, check that the messages disappear when you do `kedro run`, and update any tests affected.
> 
> Note, as per discussion in [kedro-org/kedro#1546](https://github.com/kedro-org/kedro/pull/1546), at the moment there's no way for a user to change whether or not the kedro-telemetry messages are shown by altering project-side logging.yml because `before_command_run` hooks are run before that logging configuration is set up. This is fine for now.

## Development notes

* Update the logging level to `debug`.
* Add a logging so the message should be shown when the user opt in/out for the first time.

Below are the screenshot for the user journey when they do `kedro run` for the 1st time.
| Opt-in      | Opt-out | 
| :---        |    :----:   | 
| image     | image       | 

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -54,14 +54,13 @@ class KedroTelemetryCLIHooks:
 
             consent = _check_for_telemetry_consent(project_metadata.project_path)
             if not consent:
-                click.secho(
+                logger.debug(
                     "Kedro-Telemetry is installed, but you have opted out of "
                     "sharing usage analytics so none will be collected.",
-                    fg="green",
                 )
                 return
 
-            logger.info("You have opted into product usage analytics.")
+            logger.debug("You have opted into product usage analytics.")
 
             try:
                 hashed_computer_name = hashlib.sha512(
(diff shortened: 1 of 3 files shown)


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -201,6 +200,10 @@ def _confirm_consent(telemetry_file_path: Path) -> bool:
                 yaml.dump({"consent": True}, telemetry_file)
                 click.secho("You have opted into product usage analytics.", fg="green")
                 return True
+            click.secho(
+                "You have opted out of product usage analytics, so none will be collected.",
+                fg="green",
+            )
```
noklam, line 206: This was not mentioned in the issue, but I think it needs to be printed once.

SajidAlamQB (approved):
LGTM! 👍

AntonyMilneQB (approved):
👍 Perfect!

PR #38, opened 2022-06-21
Create a new package `kedro-datasets` as part of `kedro-plugins`

noklam:

## Description

The full issues describe this https://github.com/kedro-org/kedro/issues/1492 

> Currently Kedro as a framework and kedro.extras.datasets are one package and this has caused a number of issues:
> * Kedro Framework was held back adding support for Python 3.9/3.10 by breaking changes in the datasets dependencies
> * Adding support for newer versions of the datasets was held back by the need for more conservative versioning of the framework
>  * The difference of breaking change velocity between the Framework (more stable) and the datasets (less stable) meant a painful accounting of what was new in 0.18 and what was merely a change in a dataset

This is the initial attempt to setup the `kedro-datasets` plugins package, which will enable subsequent related works and allow us to migrate all datasets code to this new package.

## Development notes

- [ ] Create the new package in `kedro-plugins`
- [ ] Copy all datasets to the new `kedro-datasets` package including the relevant `extras_require=`
https://github.com/kedro-org/kedro/blob/1b1558952c059eea5636d9ccf9a883f9cf4ef643/setup.py#L107-L148
- [ ] Move all dataset tests to `kedro-datasets`

## Checklist

### DIFF
--- a/kedro-datasets/setup.cfg
+++ b/kedro-datasets/setup.cfg
@@ -1,0 +1,10 @@
+[metadata]
+description-file=README.md
+
+[tool:pytest]
+addopts=--cov-report xml:coverage.xml
+        --cov-report term-missing
+        --cov kedro_airflow
+        --cov tests
+        --no-cov-on-fail
+        -ra
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -1,0 +1,47 @@
+import re
+from codecs import open
+from os import path
+
+from setuptools import setup
+
+name = "kedro-airflow"
+here = path.abspath(path.dirname(__file__))
+
+# get package version
+package_name = name.replace("-", "_")
+with open(path.join(here, package_name, "__init__.py"), encoding="utf-8") as f:
+    version = re.search(r'__version__ = ["\']([^"\']+)', f.read()).group(1)
+
+# get the dependencies and installs
+with open("requirements.txt", "r", encoding="utf-8") as f:
+    requires = [x.strip() for x in f if x.strip()]
+
+# get test dependencies and installs
+with open("test_requirements.txt", "r", encoding="utf-8") as f:
+    test_requires = [x.strip() for x in f if x.strip() and not x.startswith("-r")]
+
+
+# Get the long description from the README file
+with open(path.join(here, "README.md"), encoding="utf-8") as f:
+    readme = f.read()
+
+setup(
+    name=name,
+    version=version,
+    description="Kedro-Datasets is the core plugin that support Kedro's DataCatalog",
+    long_description=readme,
+    long_description_content_type="text/markdown",
+    url="https://github.com/kedro-org/kedro-plugins/tree/main/kedro-datasets",
+    author="Kedro",
+    python_requires=">=3.7, <3.11",
+    install_requires=requires,
+    tests_require=test_requires,
+    license="Apache Software License (Apache 2.0)",
+    packages=["kedro_datasets"],
+    # package_data={"kedro_airflow": ["kedro_airflow/airflow_dag_template.j2"]},
+    # include_package_data=True,
+    # zip_safe=False,
+    # entry_points={
+        # "kedro.project_commands": ["airflow = kedro_airflow.plugin:commands"]
+    # },
+)
--- a/kedro-datasets/.gitignore
+++ b/kedro-datasets/.gitignore
@@ -1,0 +1,147 @@
+# CMake
+cmake-build-debug/
+
+## File-based project format:
+*.iws
+
+## Plugin-specific files:
+
+# IntelliJ
+.idea/
+*.iml
+out/
+
+### macOS
+*.DS_Store
+.AppleDouble
+.LSOverride
+.Trashes
+
+# mpeltonen/sbt-idea plugin
+.idea_modules/
+
+# JIRA plugin
+atlassian-ide-plugin.xml
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+### Python template
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.hypothesis/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+.static_storage/
+.media/
+local_settings.py
+
+# Flask stuff:
+instance/
+.webassets-cache
+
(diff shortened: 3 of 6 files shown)


kedro-datasets/kedro_datasets/datasets/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""``kedro_datasets.datasets`` is where you can find all of Kedro's data connectors.
```
noklam, line 1: At a quick glance at [parent ticket](https://github.com/kedro-org/kedro/issues/1457), I don't see there are any ticket created for this. How are we going to link the documentation here to the `kedro` docs when we move dataset over here?

noklam:
Although CI belongs to this https://github.com/kedro-org/kedro/issues/1493 issue, but I think it's valuable to have the basic CI setup here just for PR purposes. 

The CI settings in `config.yml` missed `kedro-datasets` on purpose since we don't have anything in main.

kedro-datasets/kedro_datasets/datasets/README.md
```diff
@@ -0,0 +1,18 @@
+# Datasets
+
+Welcome to `kedro_datasets.datasets`, the home of Kedro's data connectors. Here you will find `AbstractDataSet` implementations created by QuantumBlack and external contributors.
```
MerelTheisenQB, line 3:
```suggestion
Welcome to `kedro_datasets.datasets`, the home of Kedro's data connectors. Here you will find `AbstractDataSet` implementations created by the core Kedro team and external contributors.
```

kedro-datasets/setup.cfg
```diff
@@ -0,0 +1,10 @@
+[metadata]
```
MerelTheisenQB, line 1: Do we need this file? We don't have it for core Kedro or `kedro-telemetry`

kedro-datasets/kedro_datasets/datasets/README.md
```diff
@@ -0,0 +1,18 @@
+# Datasets
```
MerelTheisenQB, line 1: I would just add this to the top-level readme instead of having two separate ones.

kedro-datasets/kedro_datasets/datasets/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Welcome to `kedro_datasets.datasets`, the home of Kedro's data connectors. Here you will find `AbstractDataSet` implementations created by QuantumBlack and external contributors.
+
+## What `AbstractDataSet` implementations are supported?
+
+We support a range of data descriptions, including CSV, Excel, Parquet, Feather, HDF5, JSON, Pickle, SQL Tables, SQL Queries, Spark DataFrames and more. We even allow support for working with images.
```
MerelTheisenQB, line 7: I don't think we need to specifically mention the support for images. 
```suggestion
We support a range of data descriptions, including CSV, Excel, Parquet, Feather, HDF5, JSON, Pickle, SQL Tables, SQL Queries, Spark DataFrames and more. 
```

kedro-datasets/kedro_datasets/datasets/README.md
MerelTheisenQB: Additionally I would keep the language consistent and either say data "connectors" or data "descriptions" but not both. My preference goes to data connectors, because that's used in the main Kedro docs as well.

kedro-datasets/kedro_datasets/datasets/__init__.py
MerelTheisenQB: I'm not sure I understand your question here? Do you mean we are missing a ticket to update all mentions in the docs from `kedro.extras.datasets` to `kedro_datasets.datasets`?

MerelTheisenQB (commented):
On first review this looks good! 🙂  I've mainly left comments around wording in the documentation for the plugin. IMO it's fine to completely ignore the tests + test setup for this PR because it's already huge and any changes will go unnoticed since all files are new in this repo.

kedro-datasets/kedro_datasets/datasets/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## What `AbstractDataSet` implementations are supported?
+
+We support a range of data descriptions, including CSV, Excel, Parquet, Feather, HDF5, JSON, Pickle, SQL Tables, SQL Queries, Spark DataFrames and more. 
+
+These data descriptions are supported with the APIs of `pandas`, `spark`, `networkx`, `matplotlib`, `yaml` and more.
+
+[The Data Catalog](https://kedro.readthedocs.io/en/stable/data/data_catalog.html) allows you to work with a range of file formats on local file systems, network file systems, cloud object stores, and Hadoop.
+
+Here is a full list of [supported data descriptions and APIs](https://kedro.readthedocs.io/en/stable/kedro.datasets.html).
```
SajidAlamQB, line 13: This link doesn't exist yet, is this a placeholder?

SajidAlamQB (approved):
Amazing job on this, I agree with what Merel has said on skipping test setup for another PR as this one is pretty big. Other than that only a few comments from me.

kedro-datasets/kedro_datasets/datasets/__init__.py
AntonyMilneQB: @deepyaman We didn't think of this but it's a very interesting idea. Where are the places where that would be less breaking than what we're doing here? I can think of the case where a user defines a custom dataset that inherits from one of our built in ones (and hence does `from kedro.extras.datasets import ...`). Are there others?

What I think @noklam was asking about here though is about the documentation, i.e. how do we get [these dataset docs in the right place](https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.html#module-kedro.extras.datasets). This is an excellent question and not something we'd thought of either. This actually sounds quite tricky and hard to do well 😬 @noklam please could you make a ticket for this if it's what you meant?

kedro-datasets/setup.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+# Get the long description from the README file
+with open(path.join(here, "README.md"), encoding="utf-8") as f:
+    readme = f.read()
+
+setup(
+    name=name,
+    version=version,
+    description="Kedro-Datasets is the core plugin that power Kedro's DataCatalog",
```
AntonyMilneQB, line 31: I think we should change this also. Maybe this?
```suggestion
    description="Kedro-Datasets is where you can find all of Kedro's data connectors.",
```

kedro-datasets/kedro_datasets/datasets/__init__.py
deepyaman: @AntonyMilneQB It would also be less breaking in cases where people specify the full path to the dataset--which some users inexplicably do. Finally, for better or for worse, new datasets would potentially work out of the box, with partial paths, with older versions of Kedro (that don't have an updated default lookup path).

I think another part of it just comes down to whether you all think sharing the `kedro` namespace would be nice--the user feels like they're using one cohesive product--or confusing.

AntonyMilneQB: @deepyaman All very good points I think. If we keep `kedro.extras.datasets` as a namespace then eventually I hope we could drop the superfluous `extras` part anyway rather than maintain it indefinitely for backwards compatibility.

AntonyMilneQB: I think sharing the `kedro` namespace would be nice, although I do see why it would be potentially confusing also. The main motivation for splitting off `kedro-datasets` is for release/distribution purposes. It's still a part of kedro I think.

deepyaman: @AntonyMilneQB @noklam I was just thinking about this more, in the context of something I'm working on--the alternative is to try and disassociate it from `kedro` as much as possible. Datasets don't actually use anything on the Kedro side, and I'm actually just now thinking about a case where I may just want the generic load/save abstraction that datasets provide. If I'm then leveraging this in a separate library, that really has nothing to do with `kedro`, it would feel awkward to be referencing the `kedro` namespace--I'd rather the top-level namespace just be called `datasets` (not even `kedro_datasets`).

Happy to talk this through, if it could be helpful for all involved; I think the tradeoffs are worth discussing. :)

PR #49, opened 2022-07-28
[PAUSED] Namespace kedro-datasets as kedro.datasets

AhdraMeraliQB:
Signed-off-by: Ahdra Merali <ahdra.merali@quantumblack.com>

## Description

This main change in this PR is in line with [this issue](https://github.com/kedro-org/kedro/issues/1693) on kedro about namespacing `kedro-dataset` as `kedro.datasets`. A subtask was to remove the datasets directory, so what was previously `kedro-datasets/kedro_datasets/datasets/...` is now `kedro-datasets/kedro/datasets/...`. There are a lot of "changed files" due to this so the key changes are broken down below:

Firstly (and perhaps most importantly), the changes to `kedro-datasets/setup.py`. These changes were made to allow for the package to be namespaced.

Secondly, mentions of `kedro_datasets.datasets` in the READMEs and some import statements in test files were updated to read as just `kedro.datasets`. This makes up the bulk

Lastly, files were renamed without being changed. This is again due to the restructuring of the directories.

#### *IMPORTANT: Development is not complete*

Tests on this branch are still broken. Linux tests currently *pass*, but code coverage reports as 0%. This is because in our tests calls to `kedro.datasets` find the package in site-packages instead of the local modules. Without installing `kedro.datasets` the import would look in the `kedro` package and return a ModuleNotFound error. Uninstalling `kedro` would fix the imports but is not an option due to dependencies in the code.

To resolve this there are two options:
- Point pytest coverage to evaluate the test coverage of the package `kedro.datasets` where it is installed.
- Rename the directories so that the import works locally. As a consequence, we would need to update the distribution name to `kedro.datasets` (it is currently still `kedro-datasets`) for imports from `kedro.datasets` to still work, which is undesirable.

The Windows tests are broken because ~it's Windows~ they are not able to install the local `kedro.datasets`.

---

To make sure kedro.datasets works as kedro.extras.datasets does, the following steps were taken:

First, `cd` into `kedro-plugins/kedro-datasets`
Run `pip install .`
This should show that `kedro.datasets-0.0.4` has been installed

From there, ensure the following commands can be run without any ImportError or ModuleNotFoundError
```
$ python3
>>> from kedro.datasets.pandas import CSVDataSet #should still work with any dataset
>>> from kedro.extras.datasets.pandas import CSVDataSet
>>> quit()
```

## Checklist


AhdraMeraliQB:
Checkpoint 📍 : `from kedro.datasets.[dummy_module] import [dummy_class]` will now work but only with directory restructuring. Still looking into achieving this without it so tests remain unchanged (and will fail) at this point.

AhdraMeraliQB (commented):
[This line](https://github.com/kedro-org/kedro-plugins/blob/111c83efb411db9955a44df26b117cc4f5b77e9e/kedro-datasets/tests/tensorflow/test_tensorflow_model_dataset.py#L78) is triggering the secret scan during linting (it hasn't been changed, unsure why it's only picking up now 🤔). Currently looking at workarounds - perhaps ignoring this file is the best route? 

Other failed tests are inherited from main 😬 and should be addressed in a separate PR.

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -15,7 +15,7 @@ class APIDataSet(AbstractDataSet):
     Example:
     ::
 
-        >>> from kedro_datasets.datasets.api import APIDataSet
+        >>> from kedro_datasets.api import APIDataSet
```
noklam, line 18: I was expecting this to be `kedro.datasets` that is mentioned in https://github.com/kedro-org/kedro/issues/1693.

So the `pypi` package name is still `kedro-datasets`, but the folder structure would be `kedro/datasets/xxx` etc.

kedro-datasets/setup.py
```diff
@@ -140,6 +140,6 @@ def _collect_requirements(requires):
     install_requires=requires,
     tests_require=test_requires,
     license="Apache Software License (Apache 2.0)",
-    packages=find_packages(exclude=["tests*"]),
+    packages=find_namespace_packages(include=["kedro_datasets*"], exclude=["tests*"]),
```
noklam, line 143: Nice!

noklam (commented):
It doesn't work for me even with package install. 
image

The Windows unit test are failing due to sparks test didn't skipped properly.
```
test-no-spark:
	cd kedro-datasets && pytest tests --no-cov --ignore tests/datasets/spark --numprocesses 4 --dist loadfile
```

AhdraMeraliQB:
@noklam interesting! I'm having trouble reproducing the error - here I'm running `ipython` through the terminal with the same `from kedro.datasets.pandas import CSVDataSet` line as above:

Screenshot 2022-08-02 at 11 45 13

I'll start looking at the failed skips for the spark test - thanks for finding that out!

kedro-datasets/kedro_datasets/api/api_dataset.py
AhdraMeraliQB: Good catch! I'll go through and replace the examples; should be good to make a note to update the kedro docs as well when the change is made!

noklam (commented):
I just realised this may break the CircleCI release too, but let's check at the end.

kedro-datasets/setup.py
```diff
@@ -140,6 +140,6 @@ def _collect_requirements(requires):
     install_requires=requires,
     tests_require=test_requires,
     license="Apache Software License (Apache 2.0)",
-    packages=find_packages(exclude=["tests*"]),
+    packages=find_namespace_packages(exclude=["tests*"]),
```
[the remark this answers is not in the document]
AhdraMeraliQB: That's odd, there should be no more mentions of `kedro_datasets` 🤔 , I'll check again. Maybe it's leftover from a previous build?

AhdraMeraliQB:
Adding on: We should not require package installation for tests to run. This requires further investigation into how pytest works with namespace packages.

AhdraMeraliQB:
@AntonyMilneQB @idanov @noklam 

Tried taking another look after our conversation yesterday:

Forcing relative imports works for the main code but it won't work for testing because Python does not record where the tests are loaded from, and the `tests` directory remains the top-level directory. [A fuller summary of the problem can be found on this thread](https://stackoverflow.com/questions/30669474/beyond-top-level-package-error-in-relative-import). There are some workarounds listed (that mostly seem to alter `sys.path` in every test file). Alternatively there are some other fixes:

1.  We could move `tests` one level lower into the `kedro` directory. 
 
In this case the imports would be from `datasets.[...]` instead of `kedro.datasets.[...]` which would solve the problem of first import win. However, we would need to restructure a lot of the automated testing and make file commands to support this new file structure. This presents a challenge as it would break the generic setup and necessitate `kedro-datasets` having its own bespoke setup. However, it may be worth noting that `kedro-datasets` does already have some bespoke make commands.

2. We could rename the `kedro-datasets` directory to `kedro_datasets`. 

This would allow us to declare relative imports from the topmost level. However, this would a) break consistency with other plugins and b) also have several ripple out effects in how we define the CI and testing for this sub-package, just like option 1.

3. We could wrap `kedro/datasets` in another directory, allowing us to import from `wrapper.kedro.datasets`. 

This is probably the least complex solution of getting around the issue. Note: We would need to explicitly declare `include(kedro/datasets)` in `find_namepace_packages` for the implementation to still work as intended.

4. Mystery option that solves everything with little-to-no drawbacks:

🦄 
 

I'm partial to option 3 myself, as it seems to be the most straightforward way to get the namespacing working, but I believe this warrants some further discussion before progressing in a given direction.

AhdraMeraliQB:
Linking @noklam's comment on possible approaches: https://github.com/kedro-org/kedro/issues/1758#issuecomment-1235232779

MerelTheisenQB:
Closing as we decided not to do namespacing: https://github.com/kedro-org/kedro/issues/1758#issuecomment-1248189590

PR #52, opened 2022-08-04
Add support for Airflow>=2.3.0 to the Kedro plugin

### LINKED ISSUE #60: Kedro-airflow python-slugify version pin is incompatible with latest version of airflow
## Description
kedro-airflow pins python-slugify with python-slugify~=4.0
airflow pins python-slugify with python-slugify>=5.0 and supports the latest version of python-slugify.
A pin of python-slugify>=4.0,<7 would be great if kedro-airflow can support it.

## Context
I was trying to pip install kedro-airflow with the latest version of airflow and got a deoendency resolution error.

## Steps to Reproduce
1. pip install apache-airflow==2.3.4 kedro-airflow
2. Receive error

## Expected Result
Ideally you can install both in the same environment.

## Actual Result
You receive an error: 
```
ERROR: Cannot install apache-airflow==2.3.4 and kedro-airflow==0.5.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    apache-airflow 2.3.4 depends on python-slugify>=5.0
    kedro-airflow 0.5.0 depends on python-slugify~=4.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

## Your Environment
Not applicable 

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

deepyaman:
## Description

Resolves part of kedro-org/kedro#1749 and resolves #60 

Also closes #80 

Also keep Kedro-Airflow plugin docstring from appearing in `kedro -h`

## Checklist


MerelTheisenQB (approved):
👍

kedro-docker/kedro_docker/plugin.py
```diff
@@ -87,8 +87,7 @@ def _make_docker_args_option(**kwargs):
 
 
 @click.group(name="Docker")
-def commands():
-    """Kedro plugin for packaging a project with Docker"""
+def commands():  # pylint: disable=missing-function-docstring
```
deepyaman, line 90:
```suggestion
def commands():
    """Kedro plugin for packaging a project with Docker"""
```

.circleci/continue_config.yml
```diff
@@ -451,6 +451,10 @@ workflows:
           matrix:
             parameters:
               python_version: ["3.7", "3.8", "3.9", "3.10"]
+          pre-steps:
+          - run:
+              name: Avoid GPL dependency (unidecode)
+              command: echo 'export SLUGIFY_USES_TEXT_UNIDECODE=yes' >> $BASH_ENV
```
deepyaman, line 457:
```suggestion
              name: Force install GPL dependency (unidecode)
              command: echo 'export AIRFLOW_GPL_UNIDECODE=yes' >> $BASH_ENV
```

.circleci/continue_config.yml
```diff
@@ -454,6 +447,10 @@ workflows:
           matrix:
             parameters:
               python_version: ["3.7", "3.8", "3.9", "3.10"]
+          pre-steps:
+          - run:
+              name: Force install GPL dependency (unidecode)
+              command: echo 'export AIRFLOW_GPL_UNIDECODE=yes' >> $BASH_ENV
```
jmholzer, line 453: Are we sure this is the right choice? I think this plugin meets the terms of the GPL license but I'm not certain. As I understand it we have an option here to use `text-unidecode` which is not distributed with a GPL license.

deepyaman: If you scroll up the commit history, I did try it in https://github.com/kedro-org/kedro-plugins/pull/52/commits/4bb506ac6d69c612e0ce9f307ab55130f045fc66, but it's quite possible something else was going on then. Let me see if it works to change it.

I don't think it's a big deal though, since we're only using it for testing, but guess we may as well avoid it if it works.

deepyaman: Ok, seems it works!

merelcht (approved):
Nice work!

noklam:
Related: https://github.com/kedro-org/kedro/pull/2030

.circleci/continue_config.yml
```diff
@@ -284,14 +279,12 @@ jobs:
...
-            or:
-              - and:
-                - equal: [ "3.10", <<parameters.python_version>> ]
-                - equal: [true, <<pipeline.parameters.run-build-kedro-datasets>>]
-              - or:
-                - equal: [false, <<pipeline.parameters.run-build-kedro-datasets>>]
+            and:
+              - not:
+                  equal: [ "3.10", <<parameters.python_version>> ]
+              - equal: [true, <<pipeline.parameters.run-build-kedro-datasets>>]
```
noklam, line 287: My brain always gets twisted when I read CCI config.

noklam (approved):
Nice work!

Side-notes:
I wonder if a YAML + Conditional  Kedro Pipeline will look exactly like this CCI config😅

.circleci/continue_config.yml
deepyaman: @noklam TBH I feel like some of this can be just done in bash. For example, whether or not it's parallel, the step can be named "Run unit tests without spark", and then whether sequential or parallel can just be decided in the job itself. Reduce the config complexity in a way, hopefully wouldn't make errors harder to debug?

noklam: I think this can be done with/without bash? This is a consequence that you want to run these tests locally too, if that logic is embedded with CI job parameters, you can't do it locally with the `Makefile`.

Issue #44, opened 2022-07-11
Understanding the average size of a Kedro project


yetudada:
# Description

One of the value propositions that Kedro supports is that it's suitable for large-scale analytics projects. A large-scale analytics project has many nodes, datasets and pipelines and might involve many team members. 

This task focuses on making it possible to collect data to support this value proposition via `kedro-telemetry`. Furthermore, I want an easy way to understand this data via Heap Analytics. 

## Context
Suppose we find this is not true in practice and the average Kedro project is small. In that case, it will affect some design choices and perhaps open up usability functionality that supports smaller projects, e.g. a YAML API for the pipelines and nodes. 

> We know that YAML is easier to work with; however, we have discouraged the YAML API pattern because it creates maintainability problems for large-scale projects. However, a YAML interface appears to be a pipeline creation pattern used by [Ploomber](https://docs.ploomber.io/en/latest/get-started/first-pipeline.html) and [MLFlow Pipelines](https://mlflow.org/docs/latest/pipelines.html), and it may suggest that projects are smaller, which is why users can get away with using it. This needs to be verified. 

## Possible Implementation

This task will require us being able to find out a few things: 

To work out the size of a Kedro project, probably tracked per `username` (following work on HEAP): 

- [ ]  Number of datasets
- [ ]  Number of nodes
- [ ]  Number of pipelines

 - To work out how many team members are on a project, to show that Kedro helps with collaboration: 
- [ ]    Number of projects, identified by `project_name` 
- [ ]    A count of `username` per project

## Possible Alternatives

Another wild idea is that we may be able to find this information by counting elements on Kedro-Viz. Kedro-Viz already has a way to find large-scale pipelines when the [pipeline warning graph appears](https://heapanalytics.com/app/env/2388822444/graph/report/preview/1999755). However, this information is incomplete because it assumes that users are also using Kedro-Viz and I'm also not sure what size of the pipeline will trigger the warning. 

![Screenshot 2022-07-11 at 11 38 55](https://user-images.githubusercontent.com/43755008/178246688-65b443f9-ab8c-4a4f-bca1-8bddc8790d8b.png)

MerelTheisenQB:
**If possible add:**
Size of projects by internal vs external users

**Nice to have:**
Could we figure out what parts of a pipeline are data engineering vs data science specific?

MerelTheisenQB:
This issue was discussed as part of a **Technical Design** session:

[...]

PR #62, opened 2022-09-29
Expand telemetry to better understand size of kedro projects in the wild

noklam:
## Description

See #44, quotes related to the actual tasks related to the scope of this PR.

> To work out the size of a Kedro project, probably tracked per username:

- Number of datasets
- Number of nodes
- Number of pipelines

To work out how many team members are on a project, to show that Kedro helps with collaboration:

- Number of projects, identified by project_name
- A count of username per project

> **If possible add:** Size of projects by internal vs external users
> 
> **Nice to have:** Could we figure out what parts of a pipeline are data engineering vs data science specific?

> This issue was discussed as part of a **Technical Design** session:
> 
> * Size should be determined by the number of nodes in the default pipeline and the number of dataset entries in the catalog
> * Currently we use the `before_command_run` hook, but to send the size info to heap we'd also need to either use the `before_pipeline_run` or `after_context_created` hook.
> * To be decided by the person implementing this: inside `kedro-telemetry` would we do a call to the Heap API for every Kedro hook that's called, or do we aggregate the data fetched in the hooks and then send it to Heap in one call?
> 
> 1. Need to implement the size calculation and sending it to heap
> 2. Make a "dashboard" or other view on Heap that shows the Kedro project size

## Development notes

* Number of pipelines
* Number of nodes in the `__default__` pipeline
* Number of datasets

### Setup for HEAP testing environment
* https://confluence.quantumblack.com/display/CAI/Heap+Analytics+for+Kedro-Viz?searchId=ZC01QX5OV

# demo
image

# How to review?
There are mainly 2 ways that users run a project
* `kedro run`
* Manually create a `KedroSessions` and do `session.run`

The difference is - the first one will always invoke the CLI first, and currently this is the only hook that captures the project metadata. `PROJECT_PROPERTIES` is a global variable that tracks this information. So the order is always CLI hook -> Kedro's run hook.

However, it's also common to skip the CLI (Databricks or any interactive workflow), and I think it's important to capture the statistics for these projects. Without the project metadata, the data on HEAP will be inconsistent, so a manual `bootstrap_project` call is added to retrieve the `metadata`. 

# How to test?
* `kedro new -s pandas-iris` # telemetry
* `cd telemetry`

1. `kedro run`
2. `kedro ipython` - `session.run()`

You should see events on Heap immediately (Better to use Dev environment) - check the `Live Data Feed` page on Heap. If you have trouble with the setup, feel free to DM me I can show you quickly how it's done.

## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,3 +1,7 @@
+# Release 0.2.3
+## Bug fixes and other changes
+* Expanding to track numbers of datasets, nodes, pipelines to better understand the size of kedro project.
+
 # Release 0.2.2
 
 ## Bug fixes and other changes
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,11 +1,8 @@
-# Release 0.2.3
-## Bug fixes and other changes
-* Expanding to track numbers of datasets, nodes, pipelines to better understand the size of kedro project.
-
 # Release 0.2.2
 
 ## Bug fixes and other changes
 * Changed the value associated with the `identity` key in the call to heap's /track endpoint from hashed computer name to hashed username.
+* Expanding to track numbers of datasets, nodes, pipelines to better understand the size of kedro project.
 
 # Release 0.2.1
 
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -115,7 +115,7 @@ class KedroTelemetryCLIHooks:
             )
 
 
-class KedroTelemetryHooks:  # pylint: disable=too-few-public-methods
+class KedroTelemetryProjectHooks:  # pylint: disable=too-few-public-methods
     """Hook to send proejct statistics data to Heap"""
 
     project_properties = PROJECT_PROPERTIES if PROJECT_PROPERTIES else {}
@@ -125,7 +125,7 @@ class KedroTelemetryHooks:  # pylint: disable=too-few-public-methods
         """Hook implementation to send proejct statistics data to Heap"""
 
         catalog = context.catalog
-        default_pipeline = pipelines.get("__default__")
+        default_pipeline = pipelines.get("__default__")  # __default__
         hashed_username = _get_hashed_username()
 
         if not self.project_properties:  # `KedroSession.run` without invoking CLI
@@ -186,17 +186,10 @@ def _format_project_statistics_data(
 ):
     """Add project staitsitcs to send to Heap."""
     project_statistics_properties = properties.copy()
-    # Initialize default value
-    project_statistics_properties["number_of_datasets"] = 0
-    project_statistics_properties["number_of_nodes"] = 0
-    project_statistics_properties["number_of_pipelines"] = 0
-
-    # Assign the project statistics
     project_statistics_properties["number_of_datasets"] = len(
         catalog.datasets.__dict__.keys()
     )
-    if default_pipeline:
-        project_statistics_properties["number_of_nodes"] = len(default_pipeline.nodes)
+    project_statistics_properties["number_of_nodes"] = len(default_pipeline.nodes) if default_pipeline else None
     project_statistics_properties["number_of_pipelines"] = len(project_pipelines.keys())
     return project_statistics_properties
 
@@ -287,4 +280,4 @@ def _confirm_consent(telemetry_file_path: Path) -> bool:
 
 
 cli_hooks = KedroTelemetryCLIHooks()
-project_telemetry_hooks = KedroTelemetryHooks()
+project_hooks = KedroTelemetryProjectHooks()
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -17,6 +17,7 @@ import yaml
 from kedro.framework.cli.cli import KedroCLI
 from kedro.framework.cli.hooks import cli_hook_impl
 from kedro.framework.hooks import hook_impl
+from kedro.framework.project import pipelines
 from kedro.framework.startup import ProjectMetadata
 from kedro.io.data_catalog import DataCatalog
 from kedro.pipeline import Pipeline
@@ -117,7 +118,6 @@ class KedroTelemetryCLIHooks:
 class KedroTelemetryHooks:
     @hook_impl
     def after_context_created(self, context):
-        from kedro.framework.project import pipelines
 
         catalog = context.catalog
         default_pipeline = pipelines.get("__default__")
@@ -171,10 +171,17 @@ def _format_project_statistics_data(
 ):
     """Add project staitsitcs to send to Heap."""
     project_statistics_properties = properties.copy()
+    # Initialize default value
+    project_statistics_properties["number_of_datasets"] = 0
+    project_statistics_properties["number_of_nodes"] = 0
+    project_statistics_properties["number_of_pipelines"] = 0
+
+    #
     project_statistics_properties["number_of_datasets"] = len(
         catalog.datasets.__dict__.keys()
     )
-    project_statistics_properties["number_of_nodes"] = len(default_pipeline.nodes)
+    if default_pipeline:
+        project_statistics_properties["number_of_nodes"] = len(default_pipeline.nodes)
     project_statistics_properties["number_of_pipelines"] = len(pipelines.keys())
     return project_statistics_properties
 
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -36,17 +36,21 @@ def _hash(string: str) -> str:
     return hashlib.sha512(bytes(string, encoding="utf8")).hexdigest()
 
 
+def _get_hashed_username():
+
+    try:
+        username = getpass.getuser()
+        return _hash(username)
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning(
+            "Something went wrong with getting the username. Exception: %s",
+            exc,
+        )
+        return ""
+
+
 # General Properties shared across hooks
 PROJECT_PROPERTIES = {}
-try:
-    username = getpass.getuser()
-    HASHED_USERNAME = _hash(username)
-except Exception as exc:  # pylint: disable=broad-except
-    logger.warning(
-        "Something went wrong with getting the username. Exception: %s",
-        exc,
-    )
-    HASHED_USERNAME = ""
 
 
 class KedroTelemetryCLIHooks:
@@ -80,8 +84,9 @@ class KedroTelemetryCLIHooks:
 
             logger.debug("You have opted into product usage analytics.")
 
+            hashed_username = _get_hashed_username()
             _prepare_project_properties(
-                PROJECT_PROPERTIES, HASHED_USERNAME, project_metadata
+                PROJECT_PROPERTIES, hashed_username, project_metadata
             )
             cli_properties = _format_user_cli_data(
                 PROJECT_PROPERTIES, masked_command_args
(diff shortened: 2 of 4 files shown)


noklam:
There is a missing piece from the discussion. How and `before_pipeline_run` called automatically without touching `settings.py`?

noklam:
Two key questions come up while I am working on this

1. Is it possible to have no users-side change at all? we need to just the Kedro run's hook i.e. `before_pipeline_run`, `before_pipeline_run` - Is it possible to not change `settings.py` at all? It seems not very feasible to ask the user to add a telemetry hook, and how would it work with `.telemetry`?
2. The original issue suggests that we get `__default__` everytime - this would have some performance drag potentially as we use lazy-loading pipeline, this would force a user to always load up the `__default__`, is it enough to use `before_pipeline_run` to only get the pipeline that user is running?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -16,20 +16,58 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def _hash(string: str) -> str:
+    return hashlib.sha512(bytes(string, encoding="utf8")).hexdigest()
+
+
+# General Properties shared across hooks
+PROJECT_PROPERTIES = {}
```
noklam, line 40: This is moved up because only the CLI has access to the project metadata

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -16,20 +16,58 @@
...
-
 HEAP_ENDPOINT = "https://heapanalytics.com/api/track"
 HEAP_HEADERS = {"Content-Type": "application/json"}
 TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 logger = logging.getLogger(__name__)
 
 
+def _hash(string: str) -> str:
+    return hashlib.sha512(bytes(string, encoding="utf8")).hexdigest()
```
noklam, line 36: minor refactoring since this is called a couple of times and they all look the same.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -16,20 +16,44 @@
...
 from kedro.framework.cli.hooks import cli_hook_impl
-from kedro.framework.startup import ProjectMetadata
+from kedro.framework.hooks import hook_impl
+from kedro.framework.project import pipelines
+from kedro.framework.startup import ProjectMetadata, bootstrap_project
+from kedro.io.data_catalog import DataCatalog
+from kedro.pipeline import Pipeline
 
-from kedro_telemetry import __version__ as telemetry_version
+from kedro_telemetry import __version__ as TELEMETRY_VERSION
```
AhdraMeraliQB, line 25: Nice catch

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
             )
 
 
-def _format_user_cli_data(
-    hashed_username: str, command_args: List[str], project_metadata: ProjectMetadata
+class KedroTelemetryHooks:  # pylint: disable=too-few-public-methods
+    """Hook to send proejct statistics data to Heap"""
```
AhdraMeraliQB, line 119:
```suggestion
    """Hook to send project statistics data to Heap"""
```

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
...
+
+    project_properties = PROJECT_PROPERTIES if PROJECT_PROPERTIES else {}
+
+    @hook_impl
+    def after_context_created(self, context):
+        """Hook implementation to send proejct statistics data to Heap"""
```
AhdraMeraliQB, line 125:
```suggestion
        """Hook implementation to send project statistics data to Heap"""
```

kedro-telemetry/setup.py
```diff
@@ -35,6 +35,7 @@
     include_package_data=True,
     zip_safe=False,
     entry_points={
-        "kedro.cli_hooks": ["kedro-telemetry = kedro_telemetry.plugin:cli_hooks"]
+        "kedro.cli_hooks": ["kedro-telemetry = kedro_telemetry.plugin:cli_hooks"],
+        "kedro.hooks": ["kedro-telemetry = kedro_telemetry.plugin:project_telemetry_hooks"]
```
AhdraMeraliQB, line 39: Would it make sense to call the latter `kedro.project_hooks`? It would follow the same format as the cli hooks and also help disambiguate.

AhdraMeraliQB (commented):
Nice work!

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
             )
 
 
-def _format_user_cli_data(
-    hashed_username: str, command_args: List[str], project_metadata: ProjectMetadata
+class KedroTelemetryHooks:  # pylint: disable=too-few-public-methods
```
MerelTheisenQB, line 118:
```suggestion
class KedroTelemetryProjectHooks:  # pylint: disable=too-few-public-methods
```

Minor: but this might make a clearer distinction between these hooks and the CLI hooks.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
...
+
+        _send_heap_event(
+            event_name="Kedro Project Statistics",
+            identity=hashed_username,
+            properties=project_statistics_properties,
+        )
+
+
+def _prepare_project_properties(
+    properties: dict, hashed_username: str, project_metadata: ProjectMetadata
```
MerelTheisenQB, line 149: I think it would be useful to add a docstring about what this method does, since it doesn't return anything.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -16,20 +16,44 @@
...
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning(
+            "Something went wrong with getting the username. Exception: %s",
+            exc,
+        )
+        return ""
+
+
+# General Properties shared across hooks
+PROJECT_PROPERTIES = {}
```
MerelTheisenQB, line 54: I'm not a big fan of having this global dict and modifying it in methods below, but I'm guessing it's necessary to share data between the several hooks..

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
             )
 
 
-def _format_user_cli_data(
-    hashed_username: str, command_args: List[str], project_metadata: ProjectMetadata
+class KedroTelemetryHooks:  # pylint: disable=too-few-public-methods
+    """Hook to send proejct statistics data to Heap"""
+
+    project_properties = PROJECT_PROPERTIES if PROJECT_PROPERTIES else {}
```
MerelTheisenQB, line 121: Do we need the `if else` statement here? If the `PROJECT_PROPERTIES` haven't been populated by a CLI call it's just an empty dict anyway right?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
...
+    return cli_properties
+
+
+def _format_project_statistics_data(
+    properties: dict,
+    catalog: DataCatalog,
+    default_pipeline: Pipeline,
+    project_pipelines: dict,
+):
+    """Add project staitsitcs to send to Heap."""
```
MerelTheisenQB, line 187:
```suggestion
    """Add project statistics to send to Heap."""
```

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -99,28 +115,90 @@ def before_command_run(
...
+    project_statistics_properties = properties.copy()
+    # Initialize default value
+    project_statistics_properties["number_of_datasets"] = 0
+    project_statistics_properties["number_of_nodes"] = 0
+    project_statistics_properties["number_of_pipelines"] = 0
```
MerelTheisenQB, line 192: Why do you need to initialise these values?

MerelTheisenQB (commented):
This is a great PR description and it also really helps you added the questions you had while implementing with the answers given during standup etc. 👍 

Thanks for refactoring the code into methods, it's much improved. I've left some questions about the new hook, but the general approach looks good to me.

kedro-telemetry/RELEASE.md
```diff
@@ -1,3 +1,7 @@
+# Release 0.2.3
+## Bug fixes and other changes
+* Expanding to track numbers of datasets, nodes, pipelines to better understand the size of kedro project.
```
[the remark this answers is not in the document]
noklam: Yes you are right😅

kedro-telemetry/kedro_telemetry/plugin.py
noklam: @MerelTheisenQB I have thought about this as well. The problem is we can't share data between CLI hooks and Project hooks as they are two separate defined spec . Currently, the CLI hook is separated from the Project hooks, and the metadata is only available in the CLI hook, so it's not possible to share this information. (Is it? could the class have a `hook_impl` and `cli_hook_impl` at the same time? I don't know. But even if they do, it still doesn't work because they would be registered as 2 entry points & we still need to take care of the situation when it is not triggered from CLI)

Originally I thought the metadata is only available via `bootstrap_project`, now I rethink about this and I can use `metadata = _get_project_metadata(project_path)`. It's still a redundant call but a less expensive one, I think prefer to kill the `PROJECT_PROPERTIES` with the cost of an extra `_get_project_metadata` call.

kedro-telemetry/kedro_telemetry/plugin.py
noklam: @MerelTheisenQB I think it's better to have a consistent schema as `default_pipelines` could be empty, in that case, the field will be missing and makes analysis on Heap more difficult.

 Maybe it's better to initialise it to `None` instead, wdyt?
`project_statistics_properties["number_of_nodes"] = len(default_pipeline.nodes) if default_pipeline else None`

MerelTheisenQB: Yes I think `None` is fine here.

kedro-telemetry/kedro_telemetry/plugin.py
MerelTheisenQB: Yeah I understand the problem. It'd definitely tricky! If the global `PROJECT_PROPERTIES` is mainly there because of the metadata then I think indeed it would be preferable to replace it with a call to `_get_project_metadata`

kedro-telemetry/kedro_telemetry/plugin.py
noklam: Good catch

kedro-telemetry/kedro_telemetry/plugin.py
noklam: Obsoleted as the `PROJECT_PROPERTIES` is killed now and this function now return a new dict instead.

AhdraMeraliQB (approved):
Thanks for explaining everything! Great work!

kedro-telemetry/tests/test_plugin.py
```diff
@@ -271,7 +317,7 @@ def test_check_for_telemetry_consent_file_invalid_yaml(self, mocker, fake_metada
     def test_confirm_consent_yaml_dump_error(self, mocker, fake_metadata, caplog):
         Path(fake_metadata.project_path, "conf").mkdir(parents=True)
         telemetry_file_path = fake_metadata.project_path / ".telemetry"
-        mocker.patch("yaml.dump", side_effect=Exception)
+        mocker.patch("yaml.dump", side_efyfect=Exception)
```
MerelTheisenQB, line 320:
```suggestion
        mocker.patch("yaml.dump", side_effect=Exception)
```

noklam:
Thanks @MerelTheisenQB  @AhdraMeraliQB.

The global variable is [now killed](https://github.com/kedro-org/kedro-plugins/pull/62/commits/d875d7e9b3f8df9322ba6950b79508498a4591eb) in favor of calling `_get_project_metadata`. The other refactoring work is in this [commit](
https://github.com/kedro-org/kedro-plugins/pull/62/commits/a932f13ef0b4f7302a082a2d6fd03da7efe0f6bd)

MerelTheisenQB (approved):
I'm glad we didn't have to use the global properties in the end. This is a nice solution! ⭐

PR #63, opened 2022-10-25
Migrate namespace from kedro_datasets.datasets to kedro_datasets

noklam:

## Description

Pre-requisite of https://github.com/kedro-org/kedro/issues/1711
Previously the namespace is incorrect. We need to update the namespace first.

## Checklist


noklam:
The 1 failing test on Python 3.10 can be ignored for now - it's not a new thing. It's a bit strange as we don't seem to have this problem in `kedro`. We should address this separately, maybe after we sync everything when we work on Release 1.0.0 and see if this problem persists.

PR #65, opened 2022-10-27
Refactor requirements.txt and test_requirements.txt to fix all dependencies being installed

### LINKED ISSUE #64: pip installing kedro-datasets[option] causes different dependencies to installing kedro[option]
## Description
Installing `kedro-datasets[option]` installs a different set of dependencies than `kedro[option]`. It appears that `kedro-datasets[option]` is installing the superset of requirements for all datasets.

## Context
This is currently blocking https://github.com/kedro-org/kedro/issues/1495

## Steps to Reproduce
1. `pip install "kedro[pandas.CSVDataSet]"; pip freeze > requirements-kedro.txt`
2. `pip install "kedro-datasets[pandas.CSVDataSet]"; pip freeze > requirements-kedro-datasets.txt`
3. Compare the requirements

jmholzer:

## Description
Closes #64

Currently, installing any of the datasets via options (e.g. `pip install "kedro-datasets[spark.SparkDataSet]"`) causes all of the requirements in `kedro-datasets/requirements.txt` and `kedro-datasets/test_requirements.txt` to be installed also. This causes a large amount of unnecessary requirements to be installed.

This PR fixes the bug by removing these requirements from setup.py.

## Development notes
I have tested this change by comparing the dependencies that are installed using `pip install kedro; pip install kedro[...]` and `pip install kedro-datasets[...]` for the options `spark.SparkDataSet` and `pandas.CSVDataSet`.

Before my changes, these options caused installation of the contents of `requirements.txt` and `test-requirements.txt`. After my changes, this no longer happens and there is no difference between using either method.

## Checklist


noklam (commented):
@jmholzer Now that I think about it more, it's weird if there is no install requirement at all.

My argument is that kedro itself should be a dependency, and some dataset would work already without any extra dependencies (i.e. Json DataSet)

I think the changes should be
-  move all the extras dependency from requirements.txt to test_requirement.txt
- requirements.txt may only have kedro (or a few more core dependency if any) 

Sorry for changing my mind as I wasn't thinking carefully enough today

merelcht:
> I think the changes should be
> 
> * move all the extras dependency from requirements.txt to test_requirement.txt
> * requirements.txt may only have kedro (or a few more core dependency if any)

I think what @noklam is saying here is the correct approach. In Kedro we have all the datasets requirements as `test_requirements.txt` and my guess is that's why it only installs the ones you need for a specific dataset.

noklam (approved):
Thanks for looking into this, just one minor comment to update the Issue's title to reflect the changes better. Will approve this now.

PR #68, opened 2022-11-01
Implement fix for failing test for loading CSVDataSet from S3 bucket

### LINKED ISSUE #67: Fix failing test for loading CSVDataSet from S3 bucket
## Description
The test for `kedro-datasets` in the file  `tests/pandas/test_csv_dataset.py::TestCSVDataSetS3::test_load_and_confirm` currently fails for Python 3.10. It only fails if called when all the tests in `tests/pandas/test_csv_dataset.py` are run. If it is called in isolation, the test will pass.

If the class `TestCSVDataSetS3` is moved to the top of the file, so that it is executed (in order) before the remaining tests in the file, the test case will pass. Since this test does not share resources with any other test in the file and also passes when called in isolation, the time taken between fixture definition and running the test may play a role in causing this error.

## Context
This bug currently blocks all development work on `kedro-datasets`.

## Steps to Reproduce
1. Checkout the `main` branch.
2. Navigate to the `kedro-datasets` directory.
3. Run `pytest --no-cov tests/pandas/test_csv_dataset.py`

All tests in this file pass apart from `TestCSVDataSetS3::test_load_and_confirm`.

The following error is produced:

```
E           kedro.io.core.DataSetError: Failed while loading data from data set CSVDataSet(filepath=test_bucket/test.csv, load_args={}, protocol=s3, save_args={'index': False}).
E           Forbidden
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.3

jmholzer:

## Description
Resolves #67

## Development notes
Discovered that the source of the bug is another bug in moto (tested with moto==4.0.8, moto==3.0.4). Changing order of tests, so that TestCSVDataSetS3::test_load_and_confirm runs before TestCSVDataSet::test_protocol_usage avoids encountering the bug.

## Checklist

### DIFF
--- a/kedro-datasets/pyproject.toml
+++ b/kedro-datasets/pyproject.toml
@@ -33,7 +33,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["tests/*", "kedro_datasets/datasets/holoviews/*"]
+omit = ["tests/*", "kedro_datasets/holoviews/*"]
 exclude_lines = ["pragma: no cover", "raise NotImplementedError"]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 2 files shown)


noklam (approved):
Thank you for looking into this.

PR #78, opened 2022-11-22
Draft PR for issue 1946 - snowflake integration

heber-urdaneta:
## Description

Moving PR from main kedro repo to kedro-plugins

## Development notes

Adding snowpark dataset and tests

## Checklist


kedro-datasets/test_requirements.txt
```diff
@@ -34,7 +34,7 @@ Pillow~=9.0
 plotly>=4.8.0, <6.0
 pre-commit>=2.9.2, <3.0  # The hook `mypy` requires pre-commit version 2.9.2.
 psutil==5.8.0
-pyarrow>=1.0, <7.0
+pyarrow>=1.0, <9.0
```
datajoely, line 37: This will need sign off from the wider team if it affects other pieces of the framework - hopefully not an issue.

datajoely: @noklam @merelcht  do you have a view here? What's the process for this sort of change.

kedro-datasets/kedro_datasets/snowflake/snowpark_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # TODO: Check if we can use default warehouse when user is not providing one explicitly
+        if not warehouse:
+            raise DataSetError("'warehouse' argument cannot be empty.")
+
+        if not database:
+            raise DataSetError("'database' argument cannot be empty.")
+
+        if not schema:
+            raise DataSetError("'schema' argument cannot be empty.")
```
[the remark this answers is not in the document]
sfc-gh-mgorkow: Not necessarily. When you create a user, you can define it's default namespace, but the default value is NULL:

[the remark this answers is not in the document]
datajoely: Thanks @sfc-gh-mgorkow then we'll be conservative here!

kedro-datasets/setup.py
```diff
@@ -73,6 +73,9 @@ def _collect_requirements(requires):
     "spark.SparkJDBCDataSet": [SPARK, HDFS, S3FS],
     "spark.DeltaTableDataSet": [SPARK, HDFS, S3FS, "delta-spark~=1.0"],
 }
+snowpark_require = {
+    "snowflake.SnowParkDataSet": ["snowflake-snowpark-python~=1.0.0", "pyarrow>=8.0, <9.0"]
```
noklam, line 77: Does it only work with `8.0`? This is equivalent to `pyarrow==8.0.0` since the next version after `8.0.0` is `9.0.0`

kedro-datasets/test_requirements.txt
noklam: Or we should include `10.0.0` at least?

kedro-datasets/kedro_datasets/snowflake/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Provides I/O modules for Snowflake."""
+
+__all__ = ["SnowParkDataSet"]
+
+from contextlib import suppress
+
+with suppress(ImportError):
```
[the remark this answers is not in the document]
heber-urdaneta: I do see this on each datasets __init__.py, for example, this is [spark's](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/spark/__init__.py)

kedro-datasets/setup.py
heber-urdaneta: I tested snowpark with pyarrow 10.0.0 and it is incompatible, the upper bound is actually more strict: "_please install a version that adheres to: 'pyarrow<8.1.0,>=8.0.0_"

noklam: How about 9.0.0?

You are right technically but this is the pyarrow release

heber-urdaneta: Version 9.0.0 doesn't work, so I think we can be more specific and adjust to `pyarrow==8.0.0`

noklam: Cool, that's fine, thanks for checking that🙏. Do you have an idea why is it not working? I think it would be nice to leave a comment there if we know certain API is not compatible.

heber-urdaneta: Sure! When creating the snowpark session it shows a warning if pyarrow version is different than 8.0.0 (**"please install a version that adheres to: 'pyarrow<8.1.0,>=8.0.0'**), and there are potential crashes when interacting with the dataframe API

kedro-datasets/kedro_datasets/snowflake/snowpark_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+
+    .. code-block:: yaml
+
+        >>> weather:
+        >>>   type: snowflake.SnowParkDataSet
+        >>>   table_name: weather_data
+        >>>   warehouse: warehouse_warehouse
+        >>>   database: meteorology
+        >>>   schema: observations
+        >>>   credentials: db_credentials
```
datajoely, line 31: I think you need `role` in the example

datajoely: And perhaps an example of how to use `externalbrowser` SSO

kedro-datasets/kedro_datasets/snowflake/snowpark_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+        >>>   load_args (WIP):
```
datajoely, line 32:
```suggestion
        >>>   load_args (WIP):
```

You don't need in the example, but It would be good to get examples of doing this form of credentials, but also the `externalbroser` SSO approach :)

Vladimir-Filimonov:
Hey everyone. We're wrapping up SnowParkData connector and I want to get feedback from the community before finalising the PR. 
We implemented connection to Snowflake via standard use of credentials of kedro. And connection to the Snowflake gets opened at the moment of catalog initialisation. Same pattern was used for [pandas SQL Dataset](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/pandas/sql_dataset.py).

But there are other ways of doing this, so below is thinking process and why we concluded that this pattern seems to be good fit.

## Using hook like pyspark dataset?
We can open connection to Snowflake using `after_context_created` hook like [pyspark starter does](https://github.com/kedro-org/kedro-starters/blob/main/pyspark/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/hooks.py). But the thing is - hooks are fired AFTER catalog & context initialisation so this means connection will be opened later than in current approach (no benefit of starting it earlier) at the cost of complicating project configuration (user have to use starter; for users using ex. AWS + Snowflake credentials getting split between credentials file and some separate `snowpark.yml` if we follow analogy of pyspark dataset).

So hook seems to be giving no value to a snowpark connector (as opposed to a pyspark, where hook makes sense) at the cost of complicating project structure.  

## Pipeline uses Snowflake only at very late stages
If we imagine kedro pipeline that does some file data preprocessing (hours?) and saving data into Snowflake only at later stages it means connection to the Snowflake will be opened all that time (hours?) before actually used. 

To address this issue we can initiate connection in a lazy fashion - `load`, `save` and `exists` methods will check if connection opened first and if not - very first use opens connection. This allows to avoid connection being opened w/o use BUT has downside that if connection credentials are wrong - we'll know about it only when pipeline will come to a stage of actually using Snowflake - which might be later in the pipeline and impact user experience. 
Based on this thinking I consider this as suboptimal design. 

### Snowflake costs consideration
Just opening connection to a Snowflake (from snowpark) does not awake Snowflake virtual warehouse you provided in connection string. So user is not charged for connection being open and from this perspective there are no drawbacks of opening connection in advance of actually using Snowflake data.

Appreciate feedback on thinking above and if you think we should change implementation of opening connection for Snowpark. @datajoely @noklam @marrrcin

heber-urdaneta:
Multiple of the circle CI checks have failed and we have some questions, here's a summary:

**1. lint and unit-tests 3.8:** it seems there was a timeout (10m) when installing requirements (particularly sqlalchemy) -> I am not sure this is caused by our commits. Have you seen this before? Can timeout maybe be extended beyond 10m?
**2. win-unit-test 3.8:** fails when installing GDAL related library -> this also doesn't look like it's caused by our changes, have you seen this occur before?
**3. tests done with python versions different than 3.8:**  for the time being, snowpark only works with Python 3.8 -> not sure how to proceed, could those tests be ignored under the acknowledgement that only version 3.8 should be used for snowpark?

@datajoely @noklam, let us know your thoughts, thanks!

datajoely:
Hi @Vladimir-Filimonov - 

Thank you for the hard work, I've checked with Kedro's Tech Lead @idanov that we don't need a hook here, so you're proposal is perfect.

@heber-urdaneta can if you run the check locally using the `makefile` do they pass?

datajoely:
Regarding this part 

>To address this issue we can initiate connection in a lazy fashion - load, save and exists methods will check if connection opened first and if not - very first use opens connection. This allows to avoid connection being opened w/o use BUT has downside that if connection credentials are wrong - we'll know about it only when pipeline will come to a stage of actually using Snowflake - which might be later in the pipeline and impact user experience.
Based on this thinking I consider this as suboptimal design.

Is there a middle-ground of where we do an eager credentials check and a lazy data load/save? Perhaps this where a `after_context_created` hook would be useful?

kedro-datasets/pyproject.toml
```diff
@@ -34,7 +34,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["tests/*", "kedro_datasets/holoviews/*"]
+omit = ["tests/*", "kedro_datasets/holoviews/*", "kedro_datasets/snowflake/*"]
```
datajoely, line 37: Why exclude?

heber-urdaneta: This was to exclude the execution of snowpark tests from the coverage report and avoid % coverage errors.

To execute snowpark tests we need a live connection to a snowflake account, we planned to keep it separated from the main tests and only execute when triggering tests with a snowflake marker (we added `make test-snowflake-only`). While the test execution is skipped if no snowflake marker is provided, it still affects the coverage report which expects 100% - which is why I added it to the omit from the report. 

Let us know your thoughts and if there's any other way around, and we can discuss further.

datajoely: No it's looking good - makes sense :)

Vladimir-Filimonov:
Ready for review PR was opened here https://github.com/kedro-org/kedro-plugins/pull/104. Due to need to fix DCO issues we had to have a clean start and sign all commits.

PR #84, opened 2022-11-28
Fix `kedro-telemetry` dependency breaking packaged projects without `pyproject.toml`

### LINKED ISSUE #83: Kedro Telemetry breaks packaged projects due to wrongly assuming `pyproject.toml` exists
## Description
Kedro Telemetry installed alongside a packaged and installed Kedro project breaks the project by assuming that the `pyproject.toml` file exists. The `pyproject.toml` is only a recipe for building the project and should not be assumed to be existing in the current folder in all cases.

The problem was introduced with https://github.com/kedro-org/kedro-plugins/pull/62

## Context
When deploying Kedro projects and if you have installed Kedro Telemetry, it breaks your project.

## Steps to Reproduce
1. Create a Kedro project
2. Add a dependency on kedro-telemetry
3. Package it through `kedro package`
4. Install it in a different environment
5. Run the project through `./<project>` in a folder where only the `conf/` is

## Expected Result
The project should run.

## Actual Result
An exception is thrown.

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.x
* Kedro plugin and kedro plugin version used (`pip show kedro-telemetry`): 0.2.2 
* Python version used (`python -V`): Not relevant
* Operating system and version: Not relevant

jmholzer:
## Description

Resolves #83.

## Development notes
Currently, when `kedro-telemetry` is installed in a packaged project that is deployed without a `pyproject.toml` file, a `RuntimeError` is thrown which prevents Kedro from being used at all in this way. It is currently necessary to read `pyproject.toml` to get three pieces of information about the project, which are then sent to heap:

- Package name
- Project name
- Project version

~~In the case that no `pyproject.toml` is present, I believe there is no other guaranteed 'source of truth' for these values. Therefore, my approach is to set these values to the string `undefined` in this case.~~

Instead of using `pyproject.toml` to find these values, this PR makes the following changes:

- Package name is found using the import `from kedro.framework.project import PACKAGE_NAME`.
- Project name is a leftover that is not important to store if we already know the package name and since there is currently no easy way of accessing this value, it is discarded.
- Project version is found using the import `from kedro import __version__`. This is the version of Kedro being used to run the project, which I feel is more important than the version of Kedro used to create the project. Feedback would be appreciated on this point.

`pyproject.toml` is therefore no longer read at all.

**Note:** in the case that `kedro.framework.project.PACKAGE_NAME` is `None`, `hashed_package_name` will be set to `undefined`. I am not sure when this would ever happen, but I think it would be better to have some value here that will at least stop the plugin from breaking the project if this case ever occurs.

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -9,16 +9,17 @@ import sys
 from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List
 
 import click
 import requests
 import yaml
+from kedro import __version__ as KEDRO_VERSION
 from kedro.framework.cli.cli import KedroCLI
 from kedro.framework.cli.hooks import cli_hook_impl
 from kedro.framework.hooks import hook_impl
-from kedro.framework.project import pipelines
-from kedro.framework.startup import ProjectMetadata, _get_project_metadata
+from kedro.framework.project import PACKAGE_NAME, pipelines
+from kedro.framework.startup import ProjectMetadata
 from kedro.io.data_catalog import DataCatalog
 from kedro.pipeline import Pipeline
 
@@ -82,9 +83,7 @@ class KedroTelemetryCLIHooks:
             logger.debug("You have opted into product usage analytics.")
 
             hashed_username = _get_hashed_username()
-            project_properties = _get_project_properties(
-                hashed_username, project_metadata
-            )
+            project_properties = _get_project_properties(hashed_username)
             cli_properties = _format_user_cli_data(
                 project_properties, masked_command_args
             )
@@ -122,12 +121,7 @@ class KedroTelemetryProjectHooks:  # pylint: disable=too-few-public-methods
         default_pipeline = pipelines.get("__default__")  # __default__
         hashed_username = _get_hashed_username()
 
-        try:
-            project_metadata = _get_project_metadata(context.project_path)
-        except RuntimeError:
-            # Case: project is packaged + deployed without `pyproject.toml` (GH #83)
-            project_metadata = None
-        project_properties = _get_project_properties(hashed_username, project_metadata)
+        project_properties = _get_project_properties(hashed_username)
 
         project_statistics_properties = _format_project_statistics_data(
             project_properties, catalog, default_pipeline, pipelines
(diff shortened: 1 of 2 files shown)


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -137,12 +141,15 @@ def after_context_created(self, context):
...
 
-    hashed_package_name = _hash(project_metadata.package_name)
-    hashed_project_name = _hash(project_metadata.project_name)
-    project_version = project_metadata.project_version
+    if project_metadata:
+        hashed_package_name = _hash(project_metadata.package_name)
+        hashed_project_name = _hash(project_metadata.project_name)
+        project_version = project_metadata.project_version
+    else:
+        hashed_package_name = hashed_project_name = project_version = "undefined"
```
noklam, line 152: As Ivan mentioned yesterday, it may be possible to get these metadata via `kedro.framework.project`. `package_name` and `project_name` are almost always the same.

Another out-of-scope question is, does the `KedroCLITelemetry` work with the packaged project then?

noklam:
> Project version is found using the import from kedro import __version__. This is the version of Kedro being used to run the project, which I feel is more important than the version of Kedro used to create the project. Feedback would be appreciated on this point.

To be more precise, this is supposed to be the version of the "starter" used. This is a good point though maybe the version of `Kedro` being used is more useful. In general, I would be ok with using either definition, but it should be consistent.

jmholzer:
> In general, I would be ok with using either definition, but it should be consistent.

You're right about this, @yetudada it would be great to hear your thoughts on whether we should upload either:

1. the version of the starter used to create a Kedro project or
2. the version of Kedro being used to run the Kedro project

to Heap. The context for this question is that currently option 1 is implemented, though option 2. is easier and safer and is currently what this PR will do.

merelcht:
The version used to create the project and the version of Kedro being used to run the project are the same. If you used an older version of Kedro to create the project e.g. `0.17.7` and then try running it with `0.18.3` you will get an error.

merelcht:
> Hmm, what about the case where the starter is e.g. 0.18.2 and the version of Kedro that is installed is 0.18.3? This is the case locally in one of my projects and I think it is valid.

Ah my bad! I thought the check was more strict and checking for a full version match. 

In that case I agree with you that we should just send the version of Kedro that's being used to run the project.

noklam (approved):
Thanks for looking into this. I think using the Kedro version is fine!

SajidAlamQB (approved):
This makes a lot of sense thanks for looking into this. 🌟

merelcht:
Don't forget to update the release notes with this fix! ✍️

PR #79, opened 2022-11-23
Optimize image layers with project code

szczeles:
## Description

Current templates in `kedro-docker` create 3 layers with user code for every build (due to separate copy->chown->chmod stages). However, during image build we can specify what should be the target owner of the files. Also, there is no need to run `chmod`, as file/dirs permission is preserved.

## Development notes

Tested locally by building the docker image and examining the filesystem. Layers optimization reduce both build time and image size.

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #92, opened 2022-12-07
[kedro-docker] Layers size optimization

szczeles:
## Description

Follow-up of #79 

This PR brings a few optimizaions for image layers:
* added `--no-cache` to the `pip install` so it clears downloaded packages after being installed
* optimized user code layers: current templates in kedro-docker create 3 layers with user code for every build (due to separate copy->chown->chmod stages), but it's not necessary, as we can specify owner of the files at `COPY` and preserve access mode
* split to 2-stage build: one with runtime-environment (a base image with no user code that can be cached and reused across projects) and second with project code only
* default base image changed to `slim` flavor and bumped python version to a modern 3.9
* redundant Spark and Hadoop libs removed from the `Dockerfile.spark` template + tests to cover this modification -> they are not necessary, because after installing `pyspark` via `pip` you have all the libraries downloaded already

## Development notes

Tested locally by building the docker image and examining the filesystem. Layers optimization reduce both build time and image size. Multi-stage build doesn't bring any extra benefits, but makes it easier to deploy Kedro in cloud environments like Azure.

## Checklist


kedro-docker/features/docker.feature
```diff
@@ -15,22 +14,6 @@ Feature: Docker commands in new projects
     And A .dive-ci file should exist
     And A .dockerignore file should exist
 
-  Scenario: Execute docker init --with-spark
```
szczeles, line 18: These two are moved to the separate file that uses different starter (`pyspark-iris`) to actually check spark context behaviour

szczeles:
@noklam This is a continuation of #79. I've measured optimizations as you asked, using the official starters.

Images built using current kedro-docker 0.3.0:

* pandas-iris (using `kedro docker build`): **1.58GB**
* pyspark-iris (using `kedro docker build --with-spark`): **3.83GB**

Same images built via kedro-docker code from this PR:

* pandas-iris: **1.44GB** - reduced by **8.8%**
* pyspark-iris: **1.91GB** - reduced by **50.1%**

marrrcin:
I think that those changes bring a real value and some best practices that are currently missing in the `Dockerfile` generated by `kedro-docker`.
Plus, @szczeles went an extra mile to fix the tests 👏🏻

kedro-docker/features/steps/cli_steps.py
```diff
@@ -161,6 +161,35 @@ def create_project_from_config_file(context):
...
+    formatter: simple
+    stream: ext://sys.stdout
+
+loggers:
+  kedro:
+    level: INFO
+
+root:
+  handlers: [console]
+"""
```
jmholzer, line 190: I think the indentation here broke, I feel this should be inline with the opening `"""`.

jmholzer (commented):
Thanks a lot for this @szczeles, this is excellent. I really appreciate the work you did on the tests.

It's great that the slim image passes the e2e tests, did you manage to do any manual testing with the slim base image?

A minor comment and one question from me.

@SajidAlamQB what do you think?

kedro-docker/test_requirements.txt
```diff
@@ -10,6 +10,7 @@ pylint>=2.4.4, <3.0
 pytest
 pytest-cov
 pytest-mock
+pytest-xdist[psutil]~=2.2.1
```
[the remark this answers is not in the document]
szczeles: Actually not, it was needed to make unit tests pass, as they work in parallel way. It's the same as [for](https://github.com/kedro-org/kedro-plugins/blob/kedro-telemtery-v0.2.2/kedro-datasets/test_requirements.txt#L43) [others](https://github.com/kedro-org/kedro-plugins/blob/kedro-telemtery-v0.2.2/kedro-telemetry/test_requirements.txt#L12)

kedro-docker/features/steps/cli_steps.py
szczeles: Good point, I remember trying this method but something dind't work. I will have another look

szczeles:
> It's great that the slim image passes the e2e tests, did you manage to do any manual testing with the slim base image?

Thanks! I did, on production we run slim images only. I've checked also that there is no issue running `pip install torch tensorflow xgboost` on the slim image (3.9) -  it looks that all major ML frameworks publish binaries already.

jmholzer:
> > It's great that the slim image passes the e2e tests, did you manage to do any manual testing with the slim base image?
> 
> Thanks! I did, on production we run slim images only. I've checked also that there is no issue running `pip install torch tensorflow xgboost` on the slim image (3.9) - it looks that all major ML frameworks publish binaries already.

Excellent, that's good to hear. It's something we have to keep in mind if we start getting reports of things not working as they should, but I think slim is the right way to go.

jmholzer (approved):
Excellent work, thanks so much! ⭐️

I've asked other core team members to take a look at this so we can push this one over the line before the winter break.

IMO we should do a release with these changes.

kedro-docker/test_requirements.txt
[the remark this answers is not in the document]
noklam: As we have one `Makefile` for all plugins, thus it runs pytests in parallel for all plugins. It only matters for the `kedro-datasets` repo but for simplicity we keep it in test requirement.

noklam (approved):
Thank you for making this PR! Lots of careful thoughts for optimizing the image sizes. Kudos for the effort for making the tests work too!

szczeles:
Thanks for your reviews @jmholzer and @noklam, do you think it would be possible to merge this to master branch and release new plugin version?

jmholzer:
> Thanks for your reviews @jmholzer and @noklam, do you think it would be possible to merge this to master branch and release new plugin version?

Yes 🙂 on it now.

noklam:
```
FROM runtime-environment
# copy the whole project except what is in .dockerignore
ARG KEDRO_UID=999
ARG KEDRO_GID=0
COPY --chown=${KEDRO_UID}:${KEDRO_GID} . .
```

I suspect this change is causing bug when UID and GID is specifcied.

PR #100, opened 2023-01-18
Bump relax pyarrow version to work the same way as Pandas

datajoely:
## Description
We only use PyArrow for `pandas.ParquetDataSet` as such I suggest we keep our versions pinned to the same range as [Pandas does](https://github.com/pandas-dev/pandas/blob/96fc51f5ec678394373e2c779ccff37ddb966e75/pyproject.toml#L100) for the same reason.

As such I suggest we remove the upper bound as we have users requesting later versions in [support channels](https://kedro-org.slack.com/archives/C03RKP2LW64/p1674040509133529)

## Development notes
Tweaked boundaries

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,6 +1,8 @@
 
 # Upcoming Release:
 
+* Relaxed PyArrow range in line with Pandas
+
 # Release 1.0.1:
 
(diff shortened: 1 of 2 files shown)


merelcht (approved):
Thanks @datajoely !

jmholzer (approved):
Good catch, thanks.

PR #127, opened 2023-03-14
feat: Add ManagedTableDataset for managed Delta Lake tables in Databricks

dannyrfar:
## Description
Creating first of few PRs to add functionality for Databricks in Kedro datasets. This PR includes the ManagedTableDataset which will allow users to interface with managed Delta tables in Databricks or locally in PySpark.

## Development notes
Changes include a net new dataset, databricks.ManagedTableDataSet, which allows users to interface with managed delta tables inside of Databricks.

## Checklist


dannyrfar:
hey @noklam just saw your comment in the other PR. I did see those two datasets, this will be more focused on Databricks Unity catalog tables. The SparkDataSet and DeltaTableDataSets are for interfacing with files directly. Both can be used on databricks but are intended for different purposes.

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,463 @@
...
+            Dict[str, str]: Dict with the details of the dataset
+        """
+        return {
+            "catalog": self._table.catalog,
+            "database": self._table.database,
+            "table": self._table.table,
+            "write_mode": self._table.write_mode,
+            "dataframe_type": self._table.dataframe_type,
+            "primary_key": self._table.primary_key,
+            "version": self._version,
```
jmholzer, line 432:
```suggestion
            "version": str(self._version),
```

For consistency with the generic `Dict[str, str]`.

jmholzer (commented):
A great start, thanks for this. I haven't yet manually tested this or tried to follow the logic too closely. It would be good if you could take me through this on a call.

The tests look fine, though I have left a few comments on style and correctness in `managed_table_dataset.py`.

In addition:
- There are quite a few comments I didn't make about things that I think running `make lint` on your project would fix, you should do this. This is currently why the `lint` check is failing.
- Tests are failing on Python 3.8 - 3.10 because the test `TestManagedTableDataSet::test_load_pandas` is failing. They are failing on 3.7 because the requirements call for a version of `mlflow` that doesn't support 3.7.

kedro-datasets/test_requirements.txt
```diff
@@ -24,6 +24,7 @@ lxml~=4.6
 matplotlib>=3.0.3, <3.4; python_version < '3.10' # 3.4.0 breaks holoviews
 matplotlib>=3.5, <3.6; python_version == '3.10'
 memory_profiler>=0.50.0, <1.0
+mlflow==2.2.1
```
jmholzer, line 27:
```suggestion
mlflow~=2.2.1; python_version>='3.8'
mlflow~=1.30.0; python_version=='3.7'
```

This dataset needs to support Python 3.7. [`mlflow 1.30.0` was the last version of `mlflow` to support 3.7](https://github.com/mlflow/mlflow/releases/tag/v2.0.1).

I would also use a compatible release operator here, not `==`.

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,463 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class ManagedTable:  # pylint: disable=R0902
+    """Stores the definition of a managed table"""
+
+    # regex for tables, catalogs and schemas
+    _NAMING_REGEX = r"\b[0-9a-zA-Z_-]{1,32}\b"
```
maria-olivia-lihn, line 32: We incorporated dash support in naming regex with the following: 

`    _NAMING_REGEX = r"\b[0-9a-zA-Z_\-]{1,32}\b"
`

maria-olivia-lihn (commented):
i would have a test where the unity catalog name contains a dash (as it used to fail)

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,463 @@
...
+    def _exists(self) -> bool:
+        """Checks to see if the table exists
+
+        Returns:
+            bool: boolean of whether the table defined
+            in the dataset instance exists in the Spark session
+        """
+        if self._table.catalog:
+            try:
+                self._get_spark().sql(f"USE CATALOG {self._table.catalog}")
```
maria-olivia-lihn, line 446: catalog should be between back tick marks as well: 
`
self._get_spark().sql(f"USE CATALOG `{self._table.catalog}`")`

maria-olivia-lihn (changes requested):

jmholzer:
I made a few changes:

- I added tests to reach 100% coverage in `managed_table_dataset`.
- I removed the functions for automatically getting and updating the version cache (and the version cache itself) as all of these were unused (and untested). They are not necessary for the dataset to function, and in other datasets we do not use this approach. They also introduced unnecessary dependencies. If we decide we want the functionality they intended (?) to offer, we can always add this in a future PR.
- I removed a walrus operator, as we need to support Python 3.7 :).

Issue #117, opened 2023-02-15
Implement `exists_function` resolution using `fsspec` in `SparkDataSet`


jmholzer:
## Description
The `exists_function` is used by `AbstractVersionedDataSet` to resolve the path to a stored, versioned dataset. In other dataset implementations, the `exists_function` is set to fsspec.filesystem(...)._fs.exists. In `SparkDataSet`, we have a complex series of logical tests to determine the `exists_function`, this is brittle (as every possible file system in use has to be anticipated a priori) and a source of error (as in https://github.com/kedro-org/kedro/issues/1801).

## Context
`SparkDataSet` sees high use, making this change would eliminate a source of bugs in a dataset that is important to our users.

noklam:
Development notes: (Continue to update)
* Tried `DatabricksFileSystem` - `from fsspec.implementations.dbfs import DatabricksFileSystem`
This could work in theory and you can pass the `token` as `credentials`. However it isn't that useful since you almost always run on **Databricks**, therefore you don't really want to authenticate via a token. This will be only useful when you try to access `dbfs` while you are running a program outside of **Databricks**.

noklam:
To summarise the current issue:

1. When `dbfs/` not include in path:
* Spark will always add `dbfs/` prefix
* `_fetch_latest_version` will use `os` library, by default it doesn't use the `dbfs` as root - as a result it can't find the file because Spark saved it in `dbfs/xxxx`

This is why when `versioned: True` + `dbfs/` not used in the filepath it cannot load the data successfully.

2. When `/dbfs` is included in path:
`kedro` strip the "dbfs" in the dataset - When using Spark it adds the `dbfs/` prefix
* DataSet get saved in `dbfs/filepath`
* `_fetch_latest_version` using os.glob("dbfs/filepath") - finding the correct version

Alternative:
Using `dbfs://` like `gcs`, `s3` instead of `dbfs/`
* This would be ideal if `fsspec` already handle it, but it doesn't
* It really only affect Spark, so conditioning in Spark seems inevitable to me.

We can potentially remove this block of condition and just let `fsspec` do the work

```python
        if fs_prefix in ("s3a://", "s3n://"):
            if fs_prefix == "s3n://":
                warn(
                    "'s3n' filesystem has now been deprecated by Spark, "
                    "please consider switching to 's3a'",
                    DeprecationWarning,
                )
            _s3 = S3FileSystem(**credentials)
            exists_function = _s3.exists
[...]

PR #128, opened 2023-03-14
Update SparkDataSet's doc and add support for `fsspec`

noklam:

## Description

Close https://github.com/kedro-org/kedro-plugins/issues/117

The final decision is to just update the doc and add support for `fsspec.filesystem` since the `DatabricksFileSystem` isn't useful for the Spark usecase.

> To summarise the current issue:
> 
> ## Summary
> 1. When `dbfs/` not include in path:❌
> 
> * Spark will always add `dbfs/` prefix - see https://docs.databricks.com/files/index.html#understand-default-locations-with-examples
> * `_fetch_latest_version` will use `os` library, by default it doesn't use the `dbfs` as root - as a result it can't find the file because Spark saved it in `dbfs/filepath` but `os` is looking in `filepath`
> 
> This is why when `versioned: True` + `dbfs/` not used in the filepath it cannot load the data successfully.
> 
> 2. When `/dbfs` is included in path: ✅
>    `kedro` strip the "dbfs" in the dataset - When using Spark it adds the `dbfs/` prefix
> 
> * SparkDataSet try to save in `filepath` - Databricks will automatically add `dbfs/` prefix which saving it in `dbfs/filepath`
> * `_fetch_latest_version` using os.glob("dbfs/filepath") - finding the correct version
> 
> ## Alternative - Use fsspec DatabrickFileSystem?:
> * Tried `DatabricksFileSystem` - `from fsspec.implementations.dbfs import DatabricksFileSystem`
>   This could work in theory and you can pass the `token` as `credentials`. However it isn't that useful since you almost always run on **Databricks**, therefore you don't really want to authenticate via a token. This will be only useful when you try to access `dbfs` while you are running a program outside of **Databricks**.
> * It really only affect Spark, so conditioning in Spark seems inevitable to me. Even if `fsspec` works, as long as we use Spark to handle save/write it will have discrepancy between `fsspec` and `spark`
> 
> ## Anything we can do?
> * We should advise user always use `dbfs` prefix ✅ - We did it already.
> * We can potentially remove this block of condition and just let `fsspec` do the work. The `dbfs` logic still need to be kept though, the condition is only special to `dbfs` or `hdfs`

## Development notes

* Add generic support using `fsspec.filesystem`
* Add documentation to explain clearly why the condition only apply to SparkDataSet
* Remove deprecation warning - since that warning has been there for many years already. I don't think it's kedro responsibility to flag Spark is not supporting `s3n` 

During development, I find it weird we actually didn't support filesystems other than `s3`. I was keen to remove the special condition for `s3` but it exists for a legacy bug (not sure if it is still a problem). IMO, I don't think that should be an issue since it should affect everything not just Spark.
However, I still keep it as I rather handle this conservatively since this is an important dataset and many legacies behind it. Once we open up this for things like `gcs`, it will give us more signal if that legacy bug still exists or not.

`glob_function = partial(_s3.glob, refresh=True)`  - This is the extra condition check for `s3` - otherwise it should be just fine to use `fsspec.filesystem` to handle.

## Additional notes (skip for review purpose)
Note:
I also found the `hdfs` tests are always wrong, I suspect no one actually uses it and I think it's a signal that we may need to re-work this dataset at some point. It's not a high priority now, but I think we stack up a lot of technical debts in this dataset already/

Some observations:
1.  Is the `hdfs` client useful at all? Some of the tests has not been running correctly for years
2. We don't have anything other than `s3` supported
3. The dataset invents it own `_split_filepath` logic which resembles `get_protocol_and_path` a lot, the only difference is the `hdfs` pathway, is this actually a bug that we should fix in `kedro`?
4. The use of `fs_prefix` and `protocol` is confusing, some of the code is read via `fsspec` some is read with Spark directly. I suspect it will break since there are inconsistencies. It works fine for `s3` and `dbfs` which is the majority of our user base I suspect.
5. Overall it just looks a lot different from other datasets, partly due to the I/O is mainly via `Spark` instead of `fsspec`

I want to refactor this dataset but I am not comfortable to do this because we don't have good coverage, maybe we will look at this again some day.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset.py
@@ -286,23 +286,17 @@ class SparkDataSet(AbstractVersionedDataSet[DataFrame, DataFrame]):
                 if ``filepath`` prefix is ``hdfs://``. Ignored otherwise.
         """
         credentials = deepcopy(credentials) or {}
-        fs_prefix, filepath = _split_filepath(filepath)
+        protocol, filepath = get_protocol_and_path(filepath, version)
         exists_function = None
         glob_function = None
 
-        if fs_prefix in ("s3a://", "s3n://"):
-            if fs_prefix == "s3n://":
-                warn(
-                    "'s3n' filesystem has now been deprecated by Spark, "
-                    "please consider switching to 's3a'",
-                    DeprecationWarning,
-                )
+        if protocol == "s3":
             _s3 = S3FileSystem(**credentials)
             exists_function = _s3.exists
             glob_function = partial(_s3.glob, refresh=True)
             path = PurePosixPath(filepath)
 
-        elif fs_prefix == "hdfs://" and version:
+        elif protocol == "hdfs" and version:
             warn(
                 f"HDFS filesystem support for versioned {self.__class__.__name__} is "
                 f"in beta and uses 'hdfs.client.InsecureClient', please use with "
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset.py
@@ -281,14 +281,14 @@ class SparkDataSet(AbstractVersionedDataSet[DataFrame, DataFrame]):
         exists_function = None
         glob_function = None
 
-        if fs_prefix == "s3":
+        if fs_prefix == "s3://":
             _s3 = S3FileSystem(**credentials)
             exists_function = _s3.exists
             # Ensure cache is not used so latest version is retrieved correctly.
             glob_function = partial(_s3.glob, refresh=True)
             path = PurePosixPath(filepath)
 
-        elif fs_prefix == "hdfs" and version:
+        elif fs_prefix == "hdfs://" and version:
             warn(
                 f"HDFS filesystem support for versioned {self.__class__.__name__} is "
                 f"in beta and uses 'hdfs.client.InsecureClient', please use with "
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset.py
@@ -123,15 +123,6 @@ def _deployed_on_databricks() -> bool:
     return "DATABRICKS_RUNTIME_VERSION" in os.environ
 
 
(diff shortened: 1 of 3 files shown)


kedro-datasets/kedro_datasets/spark/spark_dataset.py
```diff
@@ -290,15 +281,10 @@ def __init__(  # pylint: disable=too-many-arguments
         exists_function = None
         glob_function = None
 
-        if fs_prefix in ("s3a://", "s3n://"):
```
noklam, line 293: This is also a sign of legacy, since `fs_prefix` can be an empty string. `"" in ("s3a://")` will also regard as `True` and thus it always uses `S3FileSystem`. I wonder why no one ever reports this bug. Maybe all our users are using Spark on s3 or `dbfs`?

In general I found too many legacy in this dataset that deserve overhaul at some point, the trickiest part is how do we make sure it works on all platform and test it properly.

kedro-datasets/kedro_datasets/spark/spark_dataset.py
```diff
@@ -286,28 +277,36 @@ def __init__(  # pylint: disable=too-many-arguments
...
-            if fs_prefix == "s3n://":
-                warn(
-                    "'s3n' filesystem has now been deprecated by Spark, "
-                    "please consider switching to 's3a'",
-                    DeprecationWarning,
-                )
+        if not filepath.startswith("/dbfs") and _deployed_on_databricks():
+            logger.warning(
```
jmholzer, line 292: Outside of the scope of your PR, just a note to myself that this (and the accompanying test) should be changed to use `warn(...)`.

jmholzer (approved):
Excellent, thanks for this @noklam! ⭐️

jptissot:
@jmholzer I think this broke Azure DataLake Storage Gen2. I updated from kedro 0.18.8 to this version of kedro-datasets and am getting this error now: 

`Protocol not known: abfss.` 

I think the issue is the fact that the last else strips `://`

astrojuanlu:
Hi @jptissot , thanks for flagging this. For visibility, can you open an issue with the full traceback? Feel free to link back to this PR

jptissot:
@astrojuanlu Of course, I will try to get the traceback for you. I found another solution in the meantime to use `abfs` instead of `abfss`.

PR #140, opened 2023-03-22
build: Relax Kedro bound for `kedro-datasets`

merelcht:
## Description
Kedro is now pinned to `0.18.4` but actually should be a lot more flexible to work with any version of Kedro where the `AbstractVersionedDataSet` was in the same shape. See comment: https://github.com/kedro-org/kedro/issues/2409#issuecomment-1479496194

## Checklist


noklam (commented):
While it may be technically possible to use `kedro-datasets` with kedro<0.18.4, it's important to keep in mind that `kedro-datasets` was developed to be used in conjunction with kedro. If we follow this argument we can also say `kedro-datasets` will work with `kedro<0.18` as long as the user put the full namespace in `catalog.yml`, and it does not sound right to me.

I am worry that it creates some confusion. In addition, If they are using `kedro-datasets` independently, they shouldn't care about which `kedro` version it is anyway since `AbstractDataSet` was not changed for long time.

https://github.com/kedro-org/kedro/releases#:~:text=Make%20Kedro%20instantiate%20datasets%20from%20kedro_datasets%20with%20higher%20priority%20than%20kedro.extras.datasets.%20kedro_datasets%20is%20the%20namespace%20for%20the%20new%20kedro%2Ddatasets%20python%20package.

astrojuanlu (approved):

noklam (approved):
LGTM! thanks!

Issue #141, opened 2023-03-23
tensorflow dataset cannot be pip-installed on M1 Mac


astrojuanlu:
## Description
Best case scenario: the user knows the version they want, they get an immediate error:

```
> pip install "kedro-datasets[tensorflow.TensorflowModelDataset]==1.1.1"
Collecting kedro-datasets[tensorflow.TensorflowModelDataset]==1.1.1
  Using cached kedro_datasets-1.1.1-py3-none-any.whl (109 kB)
Collecting kedro~=0.18.4
  Using cached kedro-0.18.7-py3-none-any.whl (277 kB)
ERROR: Ignored the following versions that require a different python version: 0.14.0 Requires-Python >=3.5, <3.8; 0.14.1 Requires-Python >=3.5, <3.8; 0.14.2 Requires-Python >=3.5, <3.8; 0.14.3 Requires-Python >=3.5, <3.8; 0.15.0 Requires-Python >=3.5, <3.8; 0.15.1 Requires-Python >=3.5, <3.8; 0.15.2 Requires-Python >=3.5, <3.8; 0.15.3 Requires-Python >=3.5, <3.8; 0.15.4 Requires-Python >=3.5, <3.8; 0.15.5 Requires-Python >=3.5, <3.8; 0.15.6 Requires-Python >=3.5, <3.8; 0.15.7 Requires-Python >=3.5, <3.8; 0.15.8 Requires-Python >=3.5, <3.8; 0.15.9 Requires-Python >=3.5, <3.8
ERROR: Could not find a version that satisfies the requirement tensorflow~=2.0; extra == "tensorflow.tensorflowmodeldataset" (from kedro-datasets[tensorflow-tensorflowmodeldataset]) (from versions: none)
ERROR: No matching distribution found for tensorflow~=2.0; extra == "tensorflow.tensorflowmodeldataset"
```

Worst case scenario: the user does not specify a version, and pip starts backtracking aggressively...

```
> pip install "kedro-datasets[tensorflow.TensorflowModelDataset]"
Collecting kedro-datasets[tensorflow.TensorflowModelDataset]
  Using cached kedro_datasets-1.1.1-py3-none-any.whl (109 kB)
Collecting kedro~=0.18.4
  Using cached kedro-0.18.7-py3-none-any.whl (277 kB)
Collecting kedro-datasets[tensorflow.TensorflowModelDataset]
  Using cached kedro_datasets-1.1.0-py3-none-any.whl (109 kB)
[...]

PR #142, opened 2023-03-23
Fix tensorflow requirements for ARM Macs

astrojuanlu:
Fix gh-141.

## Checklist

### DIFF
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -86,7 +86,9 @@ tensorflow_required = {
     "tensorflow.TensorflowModelDataset": [
         # currently only TensorFlow V2 supported for saving and loading.
         # V1 requires HDF5 and serialises differently
-        "tensorflow~=2.0"
+        "tensorflow~=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
+        # https://developer.apple.com/metal/tensorflow-plugin/
+        "tensorflow-macos~=2.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
     ]
 }
 yaml_require = {"yaml.YAMLDataSet": [PANDAS, "PyYAML>=4.2, <7.0"]}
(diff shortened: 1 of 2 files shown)


noklam:
@ankatiyar Can we look into adding Mac or Arm? I know GitHub action support Mac for sure, we may not need to pass all checking for this hardware but it will be nice to know if we can support it.
* https://github.com/kedro-org/kedro/issues/2153

kedro-datasets/setup.py
```diff
@@ -86,7 +86,8 @@ def _collect_requirements(requires):
     "tensorflow.TensorflowModelDataset": [
         # currently only TensorFlow V2 supported for saving and loading.
         # V1 requires HDF5 and serialises differently
-        "tensorflow~=2.0"
+        "tensorflow-macos~=2.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
+        "tensorflow~=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
```
noklam, line 90: https://www.tensorflow.org/install/pip#macos_1

The official guide for MacOS is `pip install tensorflow`? Can you supplement the link where it suggest `tensorflow-macos`?

noklam (commented):
It reminds me the old day when `theano` is still the backend and installing tensorflow and CUDA properly is harder than the deep learning part🤣.

kedro-datasets/setup.py
astrojuanlu: Yep:

> For users of Apple M1 computers, to get native performance, you'll want to follow the instructions found [here](https://developer.apple.com/metal/tensorflow-plugin/).

image

So it should be this guide https://developer.apple.com/metal/tensorflow-plugin/ I'll add a comment

noklam (approved):
It looks fine for me but would be great to have someone have a M1 machine to try this out. Maybe @stichbury ?

astrojuanlu:
I tried this locally on my M1 Mac and it works, but I agree it would be nice to have someone else try it just in case 😄

deepyaman:
@astrojuanlu Wait, you mean commenting out the `tensorflow` requirement is not the best way to develop? :joy: TIL.

I'll try this out, thanks!

astrojuanlu:
> Wait, you mean commenting out the tensorflow requirement is not the best way to develop? 😂 

What are you talking about? I've never done such a thing... 😂

PR #95, opened 2023-01-07
[kedro-datasets ] Add `Polars.CSVDataSet`

wmoreiraa:
Signed-off-by: wmoreiraa <walber3@gmail.com>

## Description

Introduce python-polars to Kedro Datasets.
https://www.pola.rs/benchmarks.html

## Development notes

- Add polars.CSVDataSet
- Same tests (copy paste with some minor tweaks to pandas testcase)

## TODO on this PR:
- [x] CSVDataSet

All of those using only the eager I/O. 

## Checklist

- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the relevant `RELEASE.md` file
- [x] Added tests to cover my changes

PR #153, opened 2023-04-01
Polars Generic Dataset

wmoreiraa:
Signed-off-by: wmoreiraa <walber3@gmail.com>

## Description

Add `polars.GenericDataSet` as discussed in #95 
Ive probably closed the old PR doing wrong git things  #116 

## Development notes

Similar to `pandas.GenericDataSet`, changes made:
1. Add entry "write_mode" to be able to use formats that polars doesnt provide a write method (e.g.: delta, excel)
2. Write file permission changed to "wb" from "w". 
3. Added more tests cases for different file formats.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -4,6 +4,8 @@
 
 * Added the following new datasets:
 
+| Type                                 | Description                                                                | Location                      |
+| ------------------------------------ | -------------------------------------------------------------------------- | ----------------------------- |
 | `polars.GenericDataSet` | A `GenericDataSet` backed by [polars](https://www.pola.rs/), a lighting fast dataframe package built entirely using Rust. | `kedro_datasets.polars` |
 
 ## Bug fixes and other changes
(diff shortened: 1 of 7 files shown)


wmoreiraa:
Notes: the comments/review from @merelcht on #116  are on point and were not addressed (those not related to spark changes that are not on this PR anymore).


"Maybe I'm misunderstanding this, but if you're using a read-only data format, you simply shouldn't call save right? It kind of sounds that if you change the write mode to ignore you wouldn't get an error, but you would get raise DataSetError(f"Write mode '{self._write_mode}' is read-only.") as set in line 219." . I think this one  could go two different ways:
1. Removing the read_only formats and then removing the write_mode option : Easiest way, it would lose the functionality but I do think its very minor and it would allow more general usage and further improvement on this by community. 
2. Refactoring and fixing this implementation mistake that Ive made. 

@datajoely @astrojuanlu , I might not have any time soon to do this. Think you could finish this fix?

@merelcht all other review/comments are addressed if removing the write_mode, this is why I dont specifically copied here. Also, thank you for the review.

astrojuanlu:
Thanks a lot for getting this far @wmoreiraa. We'll take care of the final push 💪🏽

astrojuanlu:
Closing in favor of gh-170.

PR #155, opened 2023-04-03
ci: port lint, unit test, and e2e tests to Actions

ankatiyar:
## Description
Resolves https://github.com/kedro-org/kedro/issues/2153

## Development notes
The migration from CircleCI to Github Actions will be have to done in parts. This PR addresses the following parts -
* Unit tests on Linux for `kedro-docker`, `kedro-datasets`, `kedro-telemetry` & `kedro-airflow`
* Lint test on Linux for `kedro-docker`, `kedro-datasets`, `kedro-telemetry` & `kedro-airflow`
* [EDITED TO ADD] end to end test on Linux for `kedro-docker`, `kedro-telemetry` & `kedro-airflow`
* [EDITED TO ADD] windows tests for all plugins - I've been able to modify the `unit-test` job to work with both linux and windows

#### TO DOs 
The following will be added in separate PRs
- ~Unit tests on windows for `kedro-docker`, `kedro-datasets`, `kedro-telemetry` & `kedro-airflow`~
- ~End-to-end tests for `kedro-docker`, `kedro-datasets`, `kedro-telemetry` & `kedro-airflow`~
- sync and release workflows - To be done in a follow up task - https://github.com/kedro-org/kedro-plugins/issues/176

#### Notes 
- I've been experimenting with Github Actions [on a fork of this repository](https://github.com/ankatiyar/kedro-plugins). You can check the demo PRs on that repository that trigger these tests - https://github.com/ankatiyar/kedro-plugins/pulls
- The lint test is only run on Python 3.8 - This is what we had on CircleCI and also because the lint tests fails for any other version of python (for `kedro-datasets`) because of [this line in `kedro-datasets/test_requirements.txt`](https://github.com/kedro-org/kedro-plugins/blob/dd42e620efa4ca82ec31c8403cf617e4082dba4f/kedro-datasets/test_requirements.txt#L54)
- I've also added `kedro-datasets/tests/tensorflow/test_tensorflow_model_dataset.py` to `trufflehog-ignore.txt` because secret-scan was complaining [about this part](https://github.com/kedro-org/kedro-plugins/blob/8c04933d22f2eca2d466bc2f6c6170a55b48c105/kedro-datasets/tests/tensorflow/test_tensorflow_model_dataset.py#L81) 
- Also pushing some temporary changes to the `README.md` files of all plugins to trigger these tests and will revert them before merging.
- **[Question]** Adding macOS test to this would be fairly trivial - just adding `macos-latest` to `matrix.os` list in the `unit-tests` job. Should we do this? The Github action workflows are slow mostly for `kedro-datasets` (at least the way it's set up now) but they run in parallel.
- **[Question]** Should we also remove the corresponding CircleCI workflows as a part of the same PR?

## Checklist


merelcht:
FYI @SajidAlamQB and @AhdraMeraliQB

noklam (commented):
Few questions:
1. Are the unit tests triggered for all dataset? Can we trigger it conditionally with file changes like the current setup?
2. for the secret-scan, is there a way to disable by line level instead of file level? (similar to `pylint: disable=xxx`
3. Not entirely sure why lint fails only 3.8, is that because of missing imports? Can you share the related CI run.

> 
> [Question] Adding macOS test to this would be fairly trivial - just adding macos-latest to matrix.os list in the unit-tests job. Should we do this? The Github action workflows are slow mostly for kedro-datasets (at least the way it's set up now) but they run in parallel.
I'd love to see them but exclude it from our merge requirements.

> [Question] Should we also remove the corresponding CircleCI workflows as a part of the same PR?
No strong opinion about this, we can keep it run in parallel for a sprint if we want to take the safe side as long as it's not slowing us down. (I don't think it will anyway)

Haven't look into the config in details, so I just have high-level questions. I will have another look at this again.

ankatiyar:
Thanks @noklam for taking a look.

> 1. Are the unit tests triggered for all dataset? Can we trigger it conditionally with file changes like the current setup?

The current setup with circleCI also runs the tests for all datasets if the changes are made in `kedro-datasets`. I think it might be too complicated to only run specific tests for the datasets changed but we can look into it when we iterate over the GitHub actions setup once it's done. 

> 2. for the secret-scan, is there a way to disable by line level instead of file level? (similar to `pylint: disable=xxx`

I couldn't find a way to disable it for that specific line. 

> 3. Not entirely sure why lint fails only 3.8, is that because of missing imports? Can you share the related CI run.

Lint test currently on circleCi also is only set up to run with python 3.8. For any other version `snowflake-snowpark-python` is not installed and then the lint test complains about not being able to import `snowflake.snowpark`. [See this CI run](https://github.com/ankatiyar/kedro-plugins/actions/runs/4605262772/jobs/8137078950). Although error can be disabled. Either way it makes sense to run lint test only for one python version instead of all four.

> > [Question] Adding macOS test to this would be fairly trivial - just adding macos-latest to matrix.os list in the unit-tests job. Should we do this? The Github action workflows are slow mostly for kedro-datasets (at least the way it's set up now) but they run in parallel. 
>
> I'd love to see them but exclude it from our merge requirements.

I'll add it. As a side note, I think none of the tests currently are required to pass before merging, only two approvals are required. This is part of the repo settings not CI config.

merelcht (commented):
Is it necessary to have a separate CI file for each plugin? The content is pretty much the same apart from the specific plugin name, so I'm wondering if it's possible to generalise and just trigger the right plugin build depending on the changes made? 

If that's not straightforward, I'd be happy to get it setup like this first, and then we can try optimise later 🙂

ankatiyar:
> Is it necessary to have a separate CI file for each plugin? The content is pretty much the same apart from the specific plugin name, so I'm wondering if it's possible to generalise and just trigger the right plugin build depending on the changes made?
> 
> If that's not straightforward, I'd be happy to get it setup like this first, and then we can try optimise later 🙂

From what I understand, Github Actions is supposed to have one config file per workflow as opposed to CircleCI which has all workflows the config in one `config.yml` file. So my logic was to separate out the workflow/jobs which can be reused by all plugins in `unit-test.yml`. 
I'm sure there is a way to do path filtering in the same workflow but doing it with separate workflows and filtering paths under `on.push.paths` seemed like the way it is _intended_ to be done in Github Actions. I'm happy to look into it but I think it makes sense to also get the e2e tests, windows tests and the sync & release workflows in first and then optimising the whole thing.

noklam:
Path filtering is very much possible, I can help with that later.

ankatiyar:
Thanks to @SajidAlamQB for getting the end to end tests running! 🎉 I've also added them to this PR!

deepyaman:
image

Can we make sure the job names are unique? Can see https://futurestud.io/tutorials/github-actions-customize-the-job-name for an example of how to use a variable in the name (matrix already does that for you). This may also resolve itself if using a reusable workflow (as in https://github.com/kedro-org/kedro-plugins/pull/155#issuecomment-1499174327).

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python ${{matrix.python-version}}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{matrix.python-version}}
+          cache: 'pip'
```
deepyaman, line 22: image

All of the plugins are hitting the same cache key I think, which is why it's not really helping.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Install dependencies
+        run: |
+          cd ${{ inputs.plugin }}
+          pip install git+https://github.com/kedro-org/kedro@main
+          pip install -r test_requirements.txt
```
deepyaman, line 27:
```suggestion
          pip install git+https://github.com/kedro-org/kedro@main -r test_requirements.txt
```
Didn't check if this is also there is the existing CI, but we should not have multiple `pip install` commands, because they can override dependencies installed in a previous step.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.8
+          cache: 'pip'
+      - name: Install dependencies
+        run: |
+            cd ${{ inputs.plugin }}
+            pip install git+https://github.com/kedro-org/kedro@main
+            pip install -r test_requirements.txt
```
deepyaman, line 46: Same
```suggestion
            pip install git+https://github.com/kedro-org/kedro@main -r test_requirements.txt
```

.github/workflows/airflow-ci.yml
deepyaman, line 1: Nit: Out of curiosity, any reason to suffix them all with `-ci`? Will there be a `-cd` or something also? Else maybe it's extraneous.

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,15 @@
...
+
+on:
+  push:
+    paths:
+      - "kedro-docker/**"
+  pull_request:
+    paths:
+      - "kedro-docker/**"
+    types: [ synchronize ]
+jobs:
```
deepyaman, line 11:
```suggestion
    types: [ synchronize ]

jobs:
```
Nit :)

.github/workflows/airflow-ci.yml
```diff
@@ -0,0 +1,16 @@
+name: Run tests on kedro-airflow
```
deepyaman, line 1:
```suggestion
name: Run checks on kedro-airflow
```

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+          python-version: ${{matrix.python-version}}
+          cache: 'pip'
+      - name: Install dependencies
+        run: |
+          cd ${{ inputs.plugin }}
+          pip install git+https://github.com/kedro-org/kedro@main
+          pip install -r test_requirements.txt
+          pip freeze
+      - name: Run end to end tests
+        shell: 'script -q -e -c "bash {0}"'
```
deepyaman, line 78: What is this for?

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python ${{matrix.python-version}}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{matrix.python-version}}
+          cache: 'pip'
+      - name: Install dependencies
+        run: |
+          cd ${{ inputs.plugin }}
```
deepyaman, line 25: Nit:

```suggestion
          cd ${{ inputs.plugin }}
          grep ^this_should_fail_the_build /etc/passwd | cut -f 5 -d
```
At some point, can you try something like this, and make sure it fails?

I _think_ you need to explicitly set `shell: bash` to pipefail by default. (See https://github.com/actions/runner/issues/1955 and the updated documentation)

You can just set it in a defaults section up top, like

```yaml
defaults:
  shell:
    bash
```

I think. Or maybe it's not necessary, since there aren't really pipes so far, but just a thought to be on the safe side.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,79 @@
...
+          pip freeze
+      - name: Run unit tests
+        run: make plugin=${{ inputs.plugin }} test
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python 3.8
```
deepyaman, line 37: Nit: Any reason not to lint on the latest supported version instead of 3.8?

deepyaman:
Sorry for haphazardly dropping comments; I'm sure I've violated all sorts of PR review etiquette! Looks great overall. :) Left a number of comments, but I think a few that are particularly important to address (IMO):

* Make sure the cache keys are not all the same across plugins, jobs (they seem to only be differentiated by Python version right now)
* Make sure requirements for a plugin are installed in one go, rather than successive `pip install` commands

.github/workflows/unit-test.yml
SajidAlamQB: This is for the docker e2e tests, it's a custom shell. GitHub Actions has a known problem with TTY, https://github.com/actions/runner/issues/241, when running our docker e2e tests in the backend we run `docker run` with the `-it` flag which isn't supported in GitHub actions so we get an invalid TTY error. We can use a [custom shell](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#custom-shell) to run our e2e tests as a workaround.

deepyaman: Awesome, thanks! Wonder if this is worth a comment?

.github/workflows/unit-test.yml
ankatiyar: The current CI setup does it this way as well, kedro is installed first and then the `test_requirements.txt` for the plugin being tested. I tried making this change in my forked repo but the tests start failing at the "Installing dependencies" stage for `kedro-datasets` because of dependency resolution conflict. ([See this failed run](https://github.com/ankatiyar/kedro-plugins/actions/runs/4670428836/jobs/8270142915))

.github/workflows/airflow-ci.yml
ankatiyar: I changed the file names to the name of the respective plugins. (I didn't put much thought into naming the files initially but I saw some other repos had it like this)

.github/workflows/unit-test.yml
ankatiyar: Added cache step with different keys for each plugin

.github/workflows/unit-test.yml
ankatiyar: Current setup also runs lint on 3.8. The lint tests fail for other versions of python for `kedro-datasets` because of [this](https://github.com/kedro-org/kedro-plugins/blob/dd42e620efa4ca82ec31c8403cf617e4082dba4f/kedro-datasets/test_requirements.txt#L54) `snowflake-snowpark-python` is only installed for python version 3.8 so pylint throws import errors. We can change the python version to 3.10 and suppress the error

deepyaman: Thanks for answering, I'm fine with either route (or doing this in a separate task in the future); I was just wondering. :)

.github/workflows/unit-test.yml
deepyaman: I think this is a critical problem that needs resolving--whether it's in scope of this PR or not is a separate issue.

What this means is that, in reality, we don't have resolvable requirements, and we're only able to get to a resolvable state by overwriting some of the previously-installed requirements. Some of the stuff installed in the `pip install -r test_requirements.txt` are not actually going to be compatible with Kedro on `main`, it seems.

ankatiyar:
Update: I've been able make the windows tests work by modifying the `unit-test` job instead of creating a new job for the windows tests, so I've added it to this. 
I've created follow up tickets 
* #176 
* #177 

@merelcht There is no difference between the setup for `mac-OS` or `ubuntu` really. There is however a limit on the number of concurrent jobs on macOS which is 5.

merelcht:
> @merelcht There is no difference between the setup for `mac-OS` or `ubuntu` really. There is however a limit on the number of concurrent jobs on macOS which is 5.

If there's no difference I don't really see a point in adding them to be honest.

merelcht (approved):
Great work @ankatiyar and also @SajidAlamQB for making the Windows tests work 👍

SajidAlamQB (approved):
This is awesome work @ankatiyar! 🌟 I believe this is in a good state for now. We can optimise things like `pip install` overwrites in subsequent issues.

.github/workflows/check-plugin.yml
```diff
@@ -0,0 +1,133 @@
...
+          path: ~/.cache/pip
+          key: ${{inputs.plugin}}-${{matrix.os}}-python-${{matrix.python-version}}
+          restore-keys: ${{inputs.plugin}}
+      - name: Cache python packages for Windows
+        if: matrix.os == 'windows-latest'
+        uses: actions/cache@v3
+        with:
+          path: ~\AppData\Local\pip\Cache
+          key: ${{inputs.plugin}}-${{matrix.os}}-python-${{matrix.python-version}}
+          restore-keys: ${{inputs.plugin}}
```
deepyaman, line 39: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache looks cool, but not sure it would work

.github/workflows/check-plugin.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Install Kedro
+        run: pip install git+https://github.com/kedro-org/kedro@main
+      - name: Install dependencies
+        run: |
+          cd ${{ inputs.plugin }}
+          pip install -r test_requirements.txt
+      - name: Install pytables (only for kedro-datasets on windows)
+        if: matrix.os == 'windows-latest' && inputs.plugin == 'kedro-datasets'
+        run: pip install tables
```
deepyaman, line 48: Just thinking about it--if using pip, should this not be part of the `test_requirements.txt`, with an environment marker (like https://stackoverflow.com/a/54281345)

.github/workflows/check-plugin.yml
```diff
@@ -0,0 +1,133 @@
...
+          cd ${{ inputs.plugin }}
+          pytest tests
+      - name: Run unit tests for Windows / kedro-datasets / no spark sequential
+        if: matrix.os == 'windows-latest' && inputs.plugin == 'kedro-datasets' && matrix.python-version == '3.10'
+        run: |
+          make test-no-spark-sequential
+      - name: Run unit tests for Windows / kedro-datasets / no spark parallel
+        if: matrix.os == 'windows-latest' && inputs.plugin == 'kedro-datasets' && matrix.python-version != '3.10'
+        run: |
+          make test-no-spark
```
deepyaman, line 66: Maybe an item outside of the scope of the PR, but is this the same behavior on the current CircleCI build? Seems so convoluted. Maybe it makes more sense to abstract most (all?) of these differences in the Makefile, if using one anyway.

.github/workflows/check-plugin.yml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          path: ~/.cache/pip
+          key: ${{inputs.plugin}}-${{matrix.os}}-python-${{matrix.python-version}}
+          restore-keys: ${{inputs.plugin}}
+      - name: Install dependencies
+        run: |
+          cd ${{ inputs.plugin }}
+          pip install git+https://github.com/kedro-org/kedro@main
+          pip install -r test_requirements.txt
+          pip freeze
```
deepyaman, line 128: Nit: You have `pip freeze` as a separate step above, and as part of requirements installation in this workflow.

deepyaman (approved):
LGTM overall! Added some comments, but I think the main thing that would be quick/easy to try and add some value is adding `pytables` to the requirements file with an environment marker. (Also, if you could update the PR title to not be Linux specific now, or I can do it.)

.github/workflows/check-plugin.yml
ankatiyar: I quickly tried it. The `test_requirements.txt` for `kedro-datasets` has this -
```
tables~=3.6.0; platform_system == "Windows" and python_version < '3.9'
tables~=3.6; platform_system != "Windows"
```
installing`tables 3.6` does not work for windows with python 3.10. I can check in a separate PR if the version of pyTables can be safely bumped and this step can be removed after this is merged in.

PR #156, opened 2023-04-04
Fix telemetry bug - add checking for project statistics

noklam:
## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -21,7 +21,7 @@ Here is an example of what was collected.
 
 A fix must go beyond just releasing a bugfix for `kedro-telemetry`. Additionally, we will:
 
-- Delete user data collected from `kedro-telemetry >= 0.2.2 <0.2.4`.
+- Delete user data collected from `kedro-telemetry >= 0.2.2, <0.2.4`.
 - Host a team retrospective and will specify post-mortem actions
 
 > If you think of anything else we should prioritise then [let us know on Slack](https://slack.kedro.org/). And we also need to thank the contributor who flagged this: [@melvinkokxw](https://github.com/melvinkokxw) [helped identify this bug](https://github.com/kedro-org/kedro/issues/2492) and we're grateful to him because he helped give enough context for us to address this.
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,5 +1,17 @@
-# Upcoming Release 0.2.4
-* Change reference to `kedro.pipeline.Pipeline` object throughout test suite with `kedro.modular_pipeline.pipeline` factory.
+# Release 0.2.4
+* Added consent checking for collecting project statistics.
+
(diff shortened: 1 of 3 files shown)


merelcht (approved):
Thanks for the quick fix @noklam 👍 We'll need to add release notes for this, but let's wait for @yetudada to help with how we phrase this.

SajidAlamQB (approved):
Thank you!

yetudada (approved):
I added in a table showing what data is collected so that users can understand what exactly the data was. Thank you for fixing this @noklam!

kedro-telemetry/tests/test_plugin.py
```diff
@@ -433,3 +432,14 @@ def test_after_context_created_with_kedro_run(
 
         # CLI hook makes the first 2 calls, the 3rd one is the Project hook
         assert mocked_heap_call.call_args_list[2] == expected_call
+
+    def test_after_context_created_no_consent_given(self, mocker, fake_context):
```
AntonyMilneQB, line 436: This looks like a good test but just to be sure: does it fail before you applied your fix?

kedro-telemetry/RELEASE.md
```diff
@@ -1,5 +1,31 @@
-# Upcoming Release 0.2.4
-* Change reference to `kedro.pipeline.Pipeline` object throughout test suite with `kedro.modular_pipeline.pipeline` factory.
+# Release 0.2.4
+* Added consent checking for collecting project statistics.
+
+The `kedro-telemetry` plugin asks for consent to collect command run data. However, when we introduced functionality to understand the size of projects by counting the number of datasets, pipelines and nodes, the plugin did not verify consent, leading to the plugin collecting data without user consent. This bug affects users with `kedro-telemetry >= 0.2.2` installed, 22.5% of our user base.
+
+<img width="730" alt="Screenshot 2023-04-04 at 16 17 43" src="https://user-images.githubusercontent.com/43755008/229840293-fea91403-b02d-4221-9167-71d3a6196cc5.png">
```
AntonyMilneQB, line 6: Should this picture be on the repo somewhere? And is there any reason to use the HTML `` rather than the markdown `` syntax like we normally do for images?

AntonyMilneQB: (I'd also be happy to just not have the graph there at all because we already say it's 22.5% of the user base above and I doubt any user really cares about seeing the full breakdown. But it is a nice and interesting graph)

kedro-telemetry/RELEASE.md
```diff
@@ -1,5 +1,31 @@
...
+|Python version| 3.8.10 (default, Jun  2 2021, 10:49:15) |
+|Operating system used| darwin |
+| Number of datasets | 7 |
+| Number of pipelines | 2 |
+| Number of nodes | 12 |
+
+A fix must go beyond just releasing a bugfix for `kedro-telemetry`. Additionally, we will:
+
+- Delete user data collected from `kedro-telemetry >= 0.2.2, <0.2.4`.
+- Host a team retrospective and will specify post-mortem actions
```
AntonyMilneQB, line 25:
```suggestion
- Host a team retrospective and specify post-mortem actions
```

AntonyMilneQB (approved):
Looks good, thank you for the quick fix!

Ideally we should make this consent checking more robust and not just copy and pasted  between each hook e.g. by putting the check in `_send_heap_event` itself or decorating the hook functions with some `@check_consent` logic. But this fix is fine for now - we can figure out better patterns as part of the post-mortem if needed.

PR #178, opened 2023-04-18
Closes #166 Dataset api add save method

### LINKED ISSUE #166: Add a `save` method for the `APIDataSet`
## Description
We sometimes want to save some data with a REST Api (say with Django Rest Framework), most of the time issuing a POST request. So far, the `APIDataSet` is read only, with the `_save` method raising a `DataSetError`, we'd want to extend it to make a request.

## Context
I have successfully subclassed the `APIDataSet` to overwrite this method, but it could be useful to have this natively here.

## Possible Implementation
One option would be to use the same args as the `APIDataSet`, and to only define it as a saved dataset. One argument that I needed was a `chunk_size` which could be part of the `save_args` to limit the size of the payload when the data we try to save is too big and causes issues.
The data that we save is a list of dict, most of the time the result of a `pandas.DataFrame.to_dict(orient="records")`.

## Possible Alternatives
The accepted data could be a dataframe directly, but I feel like adding a node so that the users have to be explicit about their serialization would lead to less pain in the end ;).

We could hardcode the `POST` method, but I can image cases where a `PUT` method for a partial update could be useful as well.

cc @McDonnellJoseph

McDonnellJoseph:
## Description
We sometimes want to save some data with a REST Api (say with Django Rest Framework), most of the time issuing a POST request. So far, the APIDataSet is read only, with the _save method raising a DataSetError, we'd want to extend it to make a request.

## Development notes
If applied, this commit will:
1. Enable save method for APIDataSet, by sending packets of size chunk_size, default is 100 to server. Default HTTP method is POST and returns a request.Response object. 
2. Add, save_args dictionary to start conforming with other datasets.
3. Raise error when trying to save non-json data
4. Raise error for save in case of HTTP error

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/api/api_dataset.py
+++ b/kedro-datasets/kedro_datasets/api/api_dataset.py
@@ -1,6 +1,7 @@
 """``APIDataSet`` loads the data from HTTP(S) APIs.
 It uses the python requests library: https://requests.readthedocs.io/en/latest/
 """
+import json as json_  # make pylint happy
 from copy import deepcopy
 from typing import Any, Dict, Iterable, List, Union
 
@@ -51,7 +52,6 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
         >>> )
         >>> data = data_set.load()
 
-    Example of saving data with a REST API.
     """
 
     DEFAULT_SAVE_ARGS = {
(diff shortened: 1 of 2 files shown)


kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -63,19 +76,21 @@ def __init__(
...
         credentials: Union[Iterable[str], AuthBase] = None,
+        save_args: Dict[str, Any] = None,
     ) -> None:
         """Creates a new instance of ``APIDataSet`` to fetch data from an API endpoint.
 
         Args:
-            url: The API URL endpoint.
-            method: The Method of the request, GET, POST, PUT, DELETE, HEAD, etc...
-            data: The request payload, used for POST, PUT, etc requests
+            url: The API URL endpoint. method: The Method of the request, GET, POST, PUT,
```
datajoely, line 84:
```suggestion
            url: The API URL endpoint. 
            method: The Method of the request, GET, POST, PUT,
```

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -85,12 +100,20 @@ def __init__(
                 https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
             credentials: same as ``auth``. Allows specifying ``auth`` secrets in
                 credentials.yml.
-
+            save_args: Options for saving data on server. Includes all parameters used
+            during load method.
+                        Adds an optional parameter, ``chunk_size`` which determines the
```
datajoely, line 105: formatting doesn't look right here either

datajoely (commented):
Looks really really good!

PR #184, opened 2023-04-20
sync APIDataSet  from kedro's `develop`

noklam:
## Description

Fix #183 

As per early discussion, I found `APIDataSet` and ~~`ParquetDataSet`~~ have more recent changes that we forgot to sync when we moved to `kedro-datasets`.

The sync process is quite manual, I `cp -r ` the entire directory and review one by one, tests have been updated too.
## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #189, opened 2023-04-25
feat: Add metadata attribute to datasets

AhdraMeraliQB:
## Description

Resolves https://github.com/kedro-org/kedro/issues/2440

This is connected to https://github.com/kedro-org/kedro/pull/2537 which adds the metadata attribute to the datasets within `kedro.io` (MemoryDataSet, LambdaDataSet, PartitionedDataSet)

Also addresses some changes made in https://github.com/kedro-org/kedro-plugins/pull/184

## Development notes

The metadata attribute is defined solely within `self.metadata`. Some datasets make use of a `_describe` method to return a dictionary of the dataset's attributes. I have not included the metadata in these methods as in some instances it would necessitate defining it twice, and I find the use redundant - however _I would like to hear the reviewer's opinions on this matter_.

Metadata is accessible through `dataset_name.metadata`.Depending on the implementation this is at times inconsistent within the dataset, but it remains consistent across all datasets.

These changes have been tested manually using both the Python api and through `catalog.yml` using hooks. The hook I used to access the dataset's metadata is as follows:

```python
from kedro.framework.hooks import hook_impl
from typing import Any, Dict
from kedro.io import DataCatalog, MemoryDataSet

class MetadataHooks:
    @hook_impl
    def after_catalog_created(self,
        catalog: DataCatalog,
        conf_catalog: Dict[str, Any],
        conf_creds: Dict[str, Any],
        feed_dict: Dict[str, Any],
        save_version: str,
        load_versions: Dict[str, str],
    ):
        for k,v in catalog.datasets.__dict__.items():
            print(k + "metadata: \n" + str(v.metadata))
```

## Checklist


noklam:
Just thinking top of my head, do we need to change this in all the dataset? How was `layer` being used before without being an argument in `DataSet`? I think it should be yes.

It would be also great to think about how `metadata` could be access via hooks, i.e. how will viz using this new field. `after_caterlog_create` hook I guess?

AhdraMeraliQB:
> How was `layer` being used before without being an argument in `DataSet`? I think it should be yes.

`layer` is currently consumed by the DataCatalog and the general consensus is that this is not ideal. Previously it was defined on every dataset as in this PR

> It would be also great to think about how metadata could be access via hooks, i.e. how will viz using this new field. after_caterlog_create hook I guess?

`metadata` is accessible in the dummy hook implementation that I used to test the datasets, for the specifics on how viz (and other plugins) would consume and use this however, I am not sure

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -76,6 +74,8 @@ def __init__(
             credentials: Allows specifying secrets in credentials.yml.
                 Expected format is ``('login', 'password')`` if given as a tuple or list.
                 An ``AuthBase`` instance can be provided for more complex cases.
+            metadata: Any arbitrary user metadata.
```
merelcht, line 77: I think it would be good to mention here that Kedro doesn't do anything with this metadata, but that it can be consumed by plugins or directly by the user.

kedro-datasets/kedro_datasets/snowflake/snowpark_dataset.py
```diff
@@ -10,7 +10,9 @@
 logger = logging.getLogger(__name__)
 
 
-class SnowparkTableDataSet(AbstractDataSet):
+class SnowparkTableDataSet(
+    AbstractDataSet
+):  # pylint:disable=too-many-instance-attributes
```
merelcht, line 15: Maybe we should just ignore this on the repo level, it's not really a relevant check for the datasets.

noklam:
@AhdraMeraliQB @merelcht Not sure if we have a ticket already, but we should have docs explaining how this metadata should be consume i.e. via the hook

The implemention in the description is hacky as we are accessing the internal dict in weird way, I think this is related to the discussion we have about the datasets and frozen datasets.

noklam:
Or maybe just use conf_catalog or feed_dict?

kedro-datasets/kedro_datasets/svmlight/svmlight_dataset.py
```diff
@@ -98,6 +98,7 @@ def __init__(
         version: Optional[Version] = None,
         credentials: Dict[str, Any] = None,
         fs_args: Dict[str, Any] = None,
+        metadata: Dict[str, Any] = None,
     ) -> None:
         _fs_args = deepcopy(fs_args) or {}
         _fs_open_args_load = _fs_args.pop("open_args_load", {})
```
merelcht, line 104: I just noticed this dataset is missing some docstring to explain the arguments. Might be good to add that now it's being updated.

kedro-datasets/kedro_datasets/video/video_dataset.py
```diff
@@ -19,9 +19,10 @@
 class SlicedVideo:
     """A representation of slices of other video types"""
 
-    def __init__(self, video, slice_indexes):
+    def __init__(self, video, slice_indexes, metadata: Dict[str, Any] = None):
```
merelcht, line 22: The `SlicedVideo` class isn't a dataset, so it doesn't make sense to add the metadata here.

kedro-datasets/kedro_datasets/video/video_dataset.py
```diff
@@ -67,10 +68,11 @@ def __getitem__(self, index: Union[int, slice]):
 class FileVideo(AbstractVideo):
     """A video object read from a file"""
 
-    def __init__(self, filepath: str) -> None:
+    def __init__(self, filepath: str, metadata: Dict[str, Any] = None) -> None:
```
merelcht, line 71: Same here. The only dataset in this file is the `VideoDataSet`, so that's the only class where the `metadata` should be added.

kedro-datasets/kedro_datasets/api/api_dataset.py
rashidakanchwala: > > How was `layer` being used before without being an argument in `DataSet`? I think it should be yes.
> 
> `layer` is currently consumed by the DataCatalog and the general consensus is that this is not ideal. Previously it was [defined on every dataset](https://github.com/quantumblacklabs/private-kedro/pull/548/files) as in this PR
> 
> > It would be also great to think about how metadata could be access via hooks, i.e. how will viz using this new field. after_caterlog_create hook I guess?
> 
> `metadata` is accessible in the dummy hook implementation that I used to test the datasets (in PR description) , for the specifics on how viz (and other plugins) would consume and use this however, I am not sure

Hey @AhdraMeraliQB  -- i think this is very important because currently we get layer information when we context.catalog -- would the metadata information also be in that way?

merelcht (approved):
Thanks for addressing all comments. This looks good to me now 👍

kedro-datasets/kedro_datasets/api/api_dataset.py
antonymilne: We would need to change how it's consumed on kedro-viz so that instead of looking at `catalog.layers` we look at `get_dataset(dataset_name).metadata["kedro-viz"]["layer"]` (where `get_dataset` is defined in `CatalogRepository`).

antonymilne:
@noklam @AhdraMeraliQB agree that we should have a ticket to add this to the documentation. I think there's three ways this could be used by hooks:

1. `after_catalog_created`: loop through `catalog` and do something with `metadata` for each dataset. This is Ahdra's example and you're absolutely right there isn't a nice way to do this at the moment. See https://github.com/kedro-org/kedro/issues/1778#issuecomment-1237332711. But instead of using `__dict__` I think the best way for now is `for dataset_name in catalog.list(): catalog._get_dataset(dataset_name).metadata`.
2. `node_run` hooks: you have access to `catalog` and `inputs` and `outputs` which contain `dataset_name`, and so you can again do `catalog._get_dataset(dataset_name).metadata`
3. `dataset_loaded/saved` hooks: you have access to `dataset_name` but not the dataset instance. So to use `metadata` here you'd need to do the `self.catalog = catalog` trick in an earlier hook and then `self.catalog._get_dataset(dataset_name).metadata` here

Probably 1 and 2 would best examples since they're the least hacky. They're still not great due to use of `_get_dataaset`, but I'm reasonably happy to advertise this publicly because it's used in kedro-viz so won't be disappearing any time soon... It would be great to resolve #1778 some time soon so we have a nicer way of doing this.

noklam:
@antonymilne 1. is a better solution, but would it still work once we added the dataset factory https://github.com/kedro-org/kedro/pull/2560, IIRC we talked about `catalog.list()` should not be used to iterate all the dataset anymore? It only covers the registered datasets in `catalog.yml`

Maybe we don't have to worry about it now but worth to mention. I agree doing what viz is doing probably will be the best example we have, as the redesign isn't likely to happen anytime soon.

antonymilne:
@noklam you're right that going through `catalog.list()` would only show those that are explicitly defined in catalog.yml and would not cover patterned ones. If you want all the datasets actually used in a pipeline then you should use `pipelines.data_sets` instead. Either way, `catalog._get_dataset(dataset_name)` will work for both explicitly defined and pattern-matched dataset.

antonymilne:
As for the major change here: I wonder whether we should actually use `metadata` rather than `_metadata`. It looks inconsistent with other properties, but since metadata will by definition not get used anywhere in kedro we don't really need to mark it as a protected attribute. Since it's a dataset implementation attribute and not part of `AbstractDataSet` we don't currently have any sort of interface associated with exposing `_metadata`. Hence there's currently no way for consumers (e.g. plugins) to use `metadata` through a public interface, which is not very encouraging for people who want to actually use it.

In theory we could have a `@property` for `metadata` to show the underlying `_metadata` attribute but I don't see any particular need for that when we could just make `metadata` public in the first place.

Curious what @idanov thinks here...

Aside from that, this and the other PR LGTM.

> The metadata attribute is defined solely within `self.metadata`. Some datasets make use of a `_describe` method to return a dictionary of the dataset's attributes. I have not included the metadata in these methods as in some instances it would necessitate defining it twice, and I find the use redundant - however _I would like to hear the reviewer's opinions on this matter_.

Fine by me to not put it in `_describe`, but I don't understand what you mean by defining it twice here - can you give an example?

> Metadata is accessible through `dataset_name.metadata`.Depending on the implementation this is at times inconsistent within the dataset, but it remains consistent across all datasets.

Also curious what you mean by this. Where is the inconsistency?

AhdraMeraliQB:
@antonymilne 
> As for the major change here: I think I may disagree @merelcht here and wonder whether we should actually use metadata rather than _metadata. It looks inconsistent with other attribute, but since metadata will by definition not get used anywhere in kedro we don't really need to mark it as protected.

I did a little digging and it looks like the `_preview` function [added by Viz](https://github.com/kedro-org/kedro-plugins/pull/129) is the same, in that it's not used by Kedro itself. For this reason, I'd argue to keep `_metadata` and maintain consistency.

However I also agree that that marking it as protected implies that accessing it is somewhat hacky. In this case I'd actually vouch for introducing a metadata property that accesses `self._metadata`.

> Fine by me to not put it in _describe, but I don't understand what you mean by defining it twice here - can you give an example?

This really only came up in APIDataSet - the dataset has an instance variable `_request_args` which stores most of the arguments provided to the class. It's this instance variable that is then accessed by `_describe()`, so to include the metadata in `_describe()` it would need to be stored in two separate places - `self._metadata` and `self._request_args`.

> Also curious what you mean by this. Where is the inconsistency?

Another example born from APIDataSet: it mostly doesn't make use of instance variables, instead having everything stored within the one variable `self._request_args`. This implementation ignores that and defines always metadata as an instance variable for every dataset, regardless of any variance in the dataset implementations.

antonymilne:
> I did a little digging and it looks like the `_preview` function [added by Viz](https://github.com/kedro-org/kedro-plugins/pull/129) is the same, in that it's not used by Kedro itself. For this reason, I'd argue to keep `_metadata` and maintain consistency.
> 
> However I also agree that that marking it as protected implies that accessing it is somewhat hacky. In this case I'd actually vouch for introducing a metadata property that accesses `self._metadata`.

I don't think the comparison with `_preview` is quite accurate here for various reasons, but I do agree that it looks a little odd to have one public attribute when everything else is protected. 

Having a `metadata` property is ok, but I think it should probably live in `AbstractDataSet`. It's then maybe a bit weird then that the `_metadata` attribute is defined in implementations but not the abstract dataset, but I guess it's consistent with how e.g. public `AbstractDataSet.exists` function wraps the implementation's `_exists` function (this type of template method pattern I'm also not a fan of, but that's for another time).

So yeah, if you and others think the property is a good idea then fine by me 👍 It does at least make `metadata` read-only after dataset instantiation, which is probably a good thing, and also provide a public interface for users.

Thanks for explaining about `APIDataSet`. I think what you've done here is good.

kedro-datasets/RELEASE.md
```diff
@@ -3,7 +3,9 @@
 ## Major features and improvements:
 * Added pandas 2.0 support.
 * Added SQLAlchemy 2.0 support (and dropped support for versions below 1.4).
+* Added `metadata` attribute to all existing datasets.
```
antonymilne, line 6:
```suggestion
* Added `metadata` attribute to all existing datasets. This is ignored by Kedro, but may be consumed by users or external plugins.
```

antonymilne: A short explanation of what `metadata` is for would be good I think.

AhdraMeraliQB:
I agree that the metadata `property` should live in the AbstractDataSet - I'll make the changes in #2537.

@merelcht do you have any thoughts on this?

merelcht:
> I agree that the metadata property should live in the AbstractDataSet - I'll make those changes in #2537. I suppose I'll have to open a separate PR to add pass them through the datasets that can then be merged in after the changes on Kedro are released.
> 
> @merelcht do you have any thoughts on this?

I think the comments and observations @antonymilne made are very valid. I hadn't properly thought it through. So I'd prefer to just make `metadata` public then and not add a `property`. If we add the property, it would be a breaking change to remove it again in future and we don't know yet how this `metadata` feature is going to be used so I'd prefer to not tie ourselves into this additional property.

AhdraMeraliQB:
>  If we add the property, it would be a breaking change to remove it again in future and we don't know yet how this metadata feature is going to be used so I'd prefer to not tie ourselves into this additional property.

Wouldn't it be breaking either way?

noklam (approved):
The discussion has been mainly about how we should implement this new attributes. There are 3 options discussed in this PR and we go for 2 in this PR
1. self._metadata
2. self.metadata as a instance attribute
3. self.metadata (as a property)

If we want to change 2/3 in the future, it will be a breaking change since they are both public method. I do think this should goes into the Abstract Class, in future if a new dataset PR comes in and it doesn't have the metadata field, we will reject it. For this reason we should just define it explicit in the contract (abstract class).

3. seem to address the point about contract better, but it actually didn't enforce what we want. What we really want is enforcing `metadata` is part of the signature in the Class Constructor. I think the right way to do it is to enforce it in the abstract class constructor, i.e. `super().__init__(xxx, xxx)`



Follow up actions:
- We should add example hooks for how to consume this new `metadata`. Viz will likely be the first consumer.
- Remove of the `layer` attribute for `0.19.0` in favor of the `metadata` attribute

antonymilne:
@noklam agree with everything you say here, but I believe that adding `metadata` to the `AbstractDataSet` constructor was difficult for some reason. I don't know the details (I think @merelcht does) but maybe comes down to the things discussed here and related issue about where `AbstractDataSet` lives: https://github.com/kedro-org/kedro/issues/1076#issuecomment-1473645622.

> If we want to change 2/3 in the future, it will be a breaking change since they are both public method.

Agree, although a breaking change to a dataset is less awkward to deal with than to framework.

antonymilne (approved):
🌟

PR #190, opened 2023-04-25
style(datasets): use PEP 526 syntax for variable type annotations

merelcht:
## Description
Same as https://github.com/kedro-org/kedro/pull/2538 but on datasets

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #191, opened 2023-04-26
style(datasets): use PEP 585

merelcht:
## Description
Follow up on https://github.com/kedro-org/kedro-plugins/pull/190

## Development notes
https://peps.python.org/pep-0585/

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/api/api_dataset.py
+++ b/kedro-datasets/kedro_datasets/api/api_dataset.py
@@ -1,7 +1,9 @@
 """``APIDataSet`` loads the data from HTTP(S) APIs.
 It uses the python requests library: https://requests.readthedocs.io/en/latest/
 """
-from typing import Any, Dict, List, NoReturn, Tuple, Union
+from __future__ import annotations
+
+from typing import Any, NoReturn, Union
 
 import requests
 from kedro.io.core import AbstractDataSet, DataSetError
@@ -63,8 +65,8 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
         self,
         url: str,
         method: str = "GET",
-        load_args: Dict[str, Any] = None,
-        credentials: Union[Tuple[str, str], List[str], AuthBase] = None,
+        load_args: dict[str, Any] = None,
+        credentials: Union[tuple[str, str], list[str], AuthBase] = None,
     ) -> None:
         """Creates a new instance of ``APIDataSet`` to fetch data from an API endpoint.
 
@@ -95,7 +97,7 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
         if "timeout" in self._load_args:
             self._load_args["timeout"] = self._convert_type(self._load_args["timeout"])
 
-        self._request_args: Dict[str, Any] = {
+        self._request_args: dict[str, Any] = {
             "url": url,
             "method": method,
             "auth": self._convert_type(self._auth),
@@ -109,11 +111,11 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
         However, for some parameters in the Python requests library,
         only Tuples are allowed.
         """
-        if isinstance(value, List):
+        if isinstance(value, list):
             return tuple(value)
         return value
 
-    def _describe(self) -> Dict[str, Any]:
+    def _describe(self) -> dict[str, Any]:
         # prevent auth from logging
         request_args_cp = self._request_args.copy()
         request_args_cp.pop("auth", None)
--- a/kedro-datasets/kedro_datasets/biosequence/biosequence_dataset.py
+++ b/kedro-datasets/kedro_datasets/biosequence/biosequence_dataset.py
@@ -1,16 +1,18 @@
 """BioSequenceDataSet loads and saves data to/from bio-sequence objects to
 file.
 """
+from __future__ import annotations
+
 from copy import deepcopy
 from pathlib import PurePosixPath
-from typing import Any, Dict, List
+from typing import Any
 
 import fsspec
 from Bio import SeqIO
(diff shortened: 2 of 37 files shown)


merelcht:
What am I doing wrong @deepyaman @astrojuanlu

I've changed
```python
Union[plt.figure, List[plt.figure], Dict[str, plt.figure]]
```
to
```python
Union[plt.figure, list[plt.figure], dict[str, plt.figure]]
```

(and added the import `from __future__ import annotations` at the top of the file)

But now the Python 3.7 and 3.8 tests are failing with: 
```python
tests/matplotlib/test_matplotlib_writer.py:12: in <module>
    from kedro_datasets.matplotlib import MatplotlibWriter
kedro_datasets/matplotlib/__init__.py:8: in <module>
    from .matplotlib_writer import MatplotlibWriter
kedro_datasets/matplotlib/matplotlib_writer.py:24: in <module>
    Union[plt.figure, list[plt.figure], dict[str, plt.figure]], NoReturn
E   TypeError: 'type' object is not subscriptable
```

astrojuanlu:
Uh, sorry for sending you through this rabbit hole @merelcht. It seems like there's a common misconception that these annotations are usable in Python 3.7 and 3.8...

- https://stackoverflow.com/a/66007596/554319
- https://stackoverflow.com/q/62871524/554319 (long and amusing prose by @leycec as always)

I think we have to park this until we drop support for Python 3.8.

merelcht:
> Uh, sorry for sending you through this rabbit hole @merelcht. It seems like there's a common misconception that these annotations are usable in Python 3.7 and 3.8...

I managed to get it working okay on the core Kedro repo. Just the read the docs builds complaining there.

astrojuanlu:
😳 Well, I'm puzzled now.

merelcht:
Same, I thought this repo would be much more straightforward..

leycec:
*...heh.* @leycec has been summoned! Exhausting block of monolithic text now follows. Oh – and thanks so much for pinging me on, @astrojuanlu. Kedro looks *amazing,* everybody. I have two concrete suggestions with respect to PEP 585 under Python < 3.9. Something is guaranteed to work if you hit it hard enough:

* **(Recommended)** Add `typing_extensions` as a mandatory runtime dependency. Then import type hint factories from `typing_extensions` (e.g., `from typing_extensions import Dict, List`) rather than attempting to subscript raw builtin types (e.g., `list[str]`):

```python
# Import the requisite machinery. If lucky, then worky.
from typing_extensions import Dict, List, Union
#from beartype.typing import Dict, List, Union  # <-- alternate timeline approach

# @leycec guarantees success. Would he lie?
Union[plt.figure, List[plt.figure], Dict[str, plt.figure]]
```

* Alternately, add @beartype as a mandatory runtime dependency. You, of course, want to do this anyway. Then import type hint factories from `beartype.typing` (e.g., `from beartype.typing import Dict, List`) rather than attempting to subscript raw builtin types (e.g., `list[str]`). See above.

> **Type hints:** the place where nothing is straightforward and every road leads back to a critically explosive blocker in your codebase.

astrojuanlu:
Thanks a lot for chiming in! It's still not clear to me though why things are failing here while working normally over https://github.com/kedro-org/kedro/pull/2540, and also how your advice helps us leverage PEP 585 proper, namely using native collections. We'll have to do a bit more digging I suppose.

leycec:
`from __future__ import annotations` is almost certainly the issue, whatever the issue is. `from __future__ import annotations` is almost always the issue, because enabling that enables PEP 563. PEP 563 is well-known to be harmful to things that introspect type hints at runtime. This includes things like Sphinx's builtin `autodoc` extension, which would explain this:

> Just the read the docs builds complaining there.

Likewise...

> ...helps us leverage PEP 585 proper, namely using native collections...

That is a noble goal. Sadly, you can't really achieve that noble goal under Python < 3.9 without breaking runtime things. Thankfully, while noble, that isn't really a *useful* goal. Right? PEP 585 is basically just syntactic sugar; it doesn't actually gain you anything tangible. `typing_extensions` and `beartype.typing` are both dramatically preferable. They work everywhere.

In fact, [mypy itself now strongly encourages use of `typing_extensions`](https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=typing_extensions#using-new-additions-to-the-typing-module) over `from __future__ import annotations`:

> Starting with Python 3.9 ([PEP 585](https://peps.python.org/pep-0585/)), the type objects of many collections in the standard library support subscription at runtime. This means that you no longer have to import the equivalents from [typing](https://docs.python.org/3/library/typing.html#module-typing); you can simply use the built-in collections or those from [collections.abc](https://docs.python.org/3/library/collections.abc.html#module-collections.abc):

```python
from collections.abc import Sequence
x: list[str]
y: dict[int, str]
z: Sequence[str] = x
```

> There is limited support for using this syntax in Python 3.7 and later as well: if you use from `__future__ import annotations`, mypy will understand this syntax in annotations. However, since **this will not be supported by the Python interpreter at runtime**, make sure you’re aware of the caveats mentioned in the notes at [future annotations import](https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=typing_extensions#future-annotations). 

Note the critical statement: **"this will not be supported by the Python interpreter at runtime".** In other words, what you are currently trying to do is likely to raise exceptions under Python < 3.9 from things that introspect type hints at runtime (e.g., `sphinx.ext.autodoc`).

merelcht:
I'm going to close this for now as it doesn't seem easy to enforce PEP 585 everywhere and it doesn't add a lot other than syntactic sugar.

astrojuanlu:
Well, in fact I think #442 already does this 😄

PR #192, opened 2023-04-26
ci: don't run checks on both `push`/`pull_request`

deepyaman:
## Description

If you see a PR like #189, you'll see jobs are being run on `push` and `pull_request` (instead of either-or).

image

Furthermore, the path filtering is a bit too aggressive, in that it won't consider changes to files like `.pre-commit-config.yaml`, etc. It may be a bit too conservative now, in that a change to `kedro-airflow.yaml` will trigger all the checks, but:

1. changing CI workflows is comparatively rare
2. ignoring folders is easier than including the plugin folder + an exhaustive list of files

Maybe we can just add the CI workflow definitions to `paths-ignore` as well. 

## Checklist

### DIFF
--- a/.github/workflows/kedro-airflow.yml
+++ b/.github/workflows/kedro-airflow.yml
@@ -1,13 +1,20 @@
-name: Run checks on kedro-airflow
+name: Run checks on Kedro-Airflow
 
 on:
   push:
-    paths:
-      - "kedro-airflow/**"
+    branches:
+      - main
+    paths-ignore:
+      - "kedro-docker/**"
+      - "kedro-datasets/**"
+      - "kedro-telemetry/**"
   pull_request:
-    paths:
-      - "kedro-airflow/**"
-    types: [ synchronize ]
+    branches:
+      - main
+    paths-ignore:
+      - "kedro-docker/**"
+      - "kedro-datasets/**"
+      - "kedro-telemetry/**"
 
 jobs:
   airflow-test:
--- a/.github/workflows/kedro-datasets.yml
+++ b/.github/workflows/kedro-datasets.yml
@@ -1,13 +1,20 @@
-name: Run checks on kedro-datasets
+name: Run checks on Kedro-Datasets
 
 on:
   push:
-    paths:
-      - "kedro-datasets/**"
+    branches:
+      - main
+    paths-ignore:
+      - "kedro-airflow/**"
+      - "kedro-docker/**"
+      - "kedro-telemetry/**"
   pull_request:
-    paths:
-      - "kedro-datasets/**"
-    types: [ synchronize ]
+    branches:
+      - main
+    paths-ignore:
+      - "kedro-airflow/**"
+      - "kedro-docker/**"
+      - "kedro-telemetry/**"
 
 jobs:
   datasets-test:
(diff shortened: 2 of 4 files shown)


merelcht (commented):
I noticed the duplicate CI runs! Thanks for changing it. 

I don't really understand why this change made it not run for both `push` and `pull_request` though 😅  Is it because you added the branch filter? 
I'm also wondering what did `types: [ synchronize ]` do before?

deepyaman:
> I noticed the duplicate CI runs! Thanks for changing it.
> 
> I don't really understand why this change made it not run for both `push` and `pull_request` though 😅 Is it because you added the branch filter?

You want to run on both merge to `main` and pull requests against `main`.

> I'm also wondering what did `types: [ synchronize ]` do before?

There's a better description of `pull_request` types on https://frontside.com/blog/2020-05-26-github-actions-pull_request/ than on the actual docs. :) In short, it adds a trigger when commits are pushed to the branch. However, by default, it triggers on `pull_request` on `opened`, `repopened`, or `synchronize`, so this is more limiting (you'll need to explicitly trigger if not running on `push` and not opening a PR. In short, the changes make it so that it will only trigger CI on open PRs.

merelcht (approved):
Thanks @deepyaman ⭐

SajidAlamQB:
> > I noticed the duplicate CI runs! Thanks for changing it.
> > I don't really understand why this change made it not run for both `push` and `pull_request` though 😅 Is it because you added the branch filter?
> 
> You want to run on both merge to `main` and pull requests against `main`.
> 
> > I'm also wondering what did `types: [ synchronize ]` do before?
> 
> There's a better description of `pull_request` types on https://frontside.com/blog/2020-05-26-github-actions-pull_request/ than on the actual docs. :) In short, it adds a trigger when commits are pushed to the branch. However, by default, it triggers on `pull_request` on `opened`, `repopened`, or `synchronize`, so this is more limiting (you'll need to explicitly trigger if not running on `push` and not opening a PR. In short, the changes make it so that it will only trigger CI on open PRs.

Without `synchronize` would that not mean it won't run CI when you update the `branch` PR with changes on `main`?

deepyaman:
> > > I noticed the duplicate CI runs! Thanks for changing it.
> > > I don't really understand why this change made it not run for both `push` and `pull_request` though 😅 Is it because you added the branch filter?
> > 
> > 
> > You want to run on both merge to `main` and pull requests against `main`.
> > > I'm also wondering what did `types: [ synchronize ]` do before?
> > 
> > 
> > There's a better description of `pull_request` types on https://frontside.com/blog/2020-05-26-github-actions-pull_request/ than on the actual docs. :) In short, it adds a trigger when commits are pushed to the branch. However, by default, it triggers on `pull_request` on `opened`, `repopened`, or `synchronize`, so this is more limiting (you'll need to explicitly trigger if not running on `push` and not opening a PR. In short, the changes make it so that it will only trigger CI on open PRs.
> 
> Without `synchronize` would that not mean it won't run CI when you update the `branch` PR with changes on `main`?

As mentioned above in the reply you quoted, `synchronize` scope is included by default, together with `opened` and `reopened`. So it broadens the trigger types by removing `[ synchronize ]`.

ankatiyar:
I was also trying to figure out a way to only run tests once - this seems like a good way! +1 on ignoring paths instead. The only thing I guess is that the tests won't run on your branch before you open a PR? I find it useful to have the tests passing/failing before/without opening a PR (prototyping something for eg) but I don't think that's a very big problem.

SajidAlamQB (approved):

Issue #176, opened 2023-04-18
Migrate the release workflow from CircleCI to GitHub Actions


ankatiyar:
## Description
Following https://github.com/kedro-org/kedro/issues/2153, which would conclude with moving all the checks (unit tests, e2e tests, lint and windows unit tests) for the plugins to Github Actions, we also need to migrate the release workflow.

PR #203, opened 2023-05-11
ci: Migrate the release workflow from CircleCI to GitHub Actions

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-plugins/issues/176

Following https://github.com/kedro-org/kedro/issues/2153, we also need to migrate the release workflow to GA.

## Development notes

Tested on forked repo of kedro-plugins, you can find the example release workflow running [here](https://github.com/SajidAlamQB/kedro-plugins/actions/runs/4946962369/jobs/8845705942).

To fully work this, I will need the kedro-orgs pypi API key and GitHub access token and add them to the `kedro-plugins` repo secrets.

[17-05-23]
Here is the latest workflow running on my [fork](https://github.com/SajidAlamQB/kedro-plugins/actions/runs/5002681364/jobs/8963219574)

## Checklist


ankatiyar:
This looks good! We should add a step for running the unit tests, linters and e2e tests before publishing a release. To that end, it might be worth dividing the workflow into separate jobs like -
```yaml
jobs:
  check-version: 
  (Checks if the version was updated and the new version exists on PyPi, set package_name)
  test: 
    needs: check-version
  (Reuse the `check-plugin.yml` workflow to run tests)
  build: 
    needs: check-version
  (Build only for one version of python?)
  release:
    needs: [test, build]
    steps:
      - name: GitHub release
      - name: PyPi release
```
`check-plugin.yml` currently runs all the tests on all versions of python but we can modify that in a separate PR to accept which python versions to run tests on just like it accepts plugin name from the "caller" workflow so we don't have to run all the tests 4 times.

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,80 @@
...
+            new_release=true
+            break
+          else
+            echo "Skipped: $package_name $package_version already exists on PyPI"
+          fi
+        done
+        echo "new_release=$new_release" >> "$GITHUB_ENV"
+        echo "dist_path=$dist_path" >> "$GITHUB_ENV"
+        echo "package_name=$package_name" >> "$GITHUB_ENV"
+        echo "package_version=$package_version" >> "$GITHUB_ENV"
```
noklam, line 53: You can use Python for scripting without writing Bash.

https://github.com/noklam/github_action/blob/main/.github/workflows/ci.yaml

noklam: Ideally, I don't expect the release logic only exist in the CI file, it should be a simple call. We should be able to run it locally similarly with `make xxxx`.

SajidAlamQB: > One more comment: I think we need to run the test before we publish.
> 
> If all plugins are going to be released. Will they all be published as 1 workflow or 4 separate one? I think it's better to have separate jobs because each of them will have their own testing requirements.
> 
> 1. Check version - see if need to release
> 2. Each plugin gets a separate job for the release (some could fail, some could pass)

This is something I was wondering about, currently, if we were to release several plugins in one PR the script won't work properly as it's designed to only accept 1 plugin release. But would there be a scenario where we would have multiple releases in one PR?

SajidAlamQB: > Ideally, I don't expect the release logic only exist in the CI file, it should be a simple call. We should be able to run it locally similarly with `make xxxx`.

I am open to this idea, just to clarify maybe we keep the `Check versions and Build-Publish` logic in a python file similar to `tools/circleci/circleci_release.py`?

noklam: I think it's not likely we release everything in a PR. But it's good to avoid coupling tightly, we can have a release job that accept plugin as an argument. Then the job will be simply looping through the plugin and decide if it needs to trigger a release job. The same plugin argument  will also be passed to the check plugin job to trigger tests.

ankatiyar: One way I think this can be done is to have a workflow with 4 jobs each for different plugin that checks if the release should be done eg - `main-updated.yml`
```yaml
name: Main updated

on:
  push:
    branches:
      - main

jobs:
  maybe-trigger-release-docker: 
    steps:
       - name: check if docker needs to be released (__init__.py file updated & new version doesn't exist on PyPi)
       - name: if previous step true, call reusable release workflow
         uses: .github/workflows/release.yml
         with: 
           plugin: docker

  maybe-trigger-release-datasets: 
    steps:
       - name: check if datasets needs to be released (__init__.py file updated & new version doesn't exist on PyPi)
       - name: if previous step true, call reusable release workflow
         uses: .github/workflows/release.yml
         with: 
           plugin: datasets

(same for airflow and telemetry)
``` 
And then have a reusable workflow `release.yml` which accepts plugin name and does the following jobs -
```yaml
on:
  workflow_call:
    inputs:
      plugin:
        type: string

jobs:
  build: 
  test: (call check-plugin.yml with plugin)
  release: 
    needs: [build, test]
```

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,81 @@
+name: Check Versions and Build-Publish
```
ankatiyar, line 1: Nitpicking but for consistency the name could be sentence case since all the other workflows' names are.

tools/github_actions/github_actions_release.py
```diff
@@ -0,0 +1,50 @@
...
+if __name__ == "__main__":
+    """Check if a package needs to be released"""
+    base_path = Path()
+    for package_path in PACKAGE_PATHS:
+        package_name, _ = package_path.split("/")
+        package_version = get_package_version(base_path, package_path)
+        pypi_endpoint = f"https://pypi.org/pypi/{package_name}/{package_version}/json/"
+        env_file = os.getenv('GITHUB_ENV')
+
+        print(package_name, package_version)
```
ankatiyar, line 43: This isn't needed I think.

noklam (approved):
It looks ready to me! One last minor comment is usually we set ENV variable with CAPITAL letter,i.e. `GITHUB_ENV`.

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+jobs:
+  check-version:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: 3.8
```
merelcht, line 16: Can we also use a newer python version? I'd go for the most recent Kedro supports (so 3.10) if possible.

tools/github_actions/github_actions_release.py
```diff
@@ -0,0 +1,49 @@
...
+    init_file_path = Path(base_path) / package_path / "__init__.py"
+    match_obj = re.search(VERSION_MATCHSTR, Path(init_file_path).read_text())
+    return match_obj.group(1)
+
+
+def check_no_version_pypi(pypi_endpoint, package_name, package_version):
+    print(f"Check if {package_name} {package_version} is on pypi")
+    response = requests.get(pypi_endpoint, timeout=10)
+    if response.status_code == 404:
+        # Not exist on Pypi - do release
```
merelcht, line 26:
```suggestion
        # Version doesn't exist on Pypi - do release
```

merelcht (commented):
Two small comments from my side, but overall this looks good 👍

PR #206, opened 2023-05-12
Feat/add managed delta table dataset

jmholzer:
## Checklist


kedro-datasets/test_requirements.txt
```diff
@@ -24,13 +24,15 @@ lxml~=4.6
 matplotlib>=3.0.3, <3.4; python_version < '3.10' # 3.4.0 breaks holoviews
 matplotlib>=3.5, <3.6; python_version == '3.10'
 memory_profiler>=0.50.0, <1.0
+mlflow~=1.30.0; python_version=='3.7'
+mlflow~=2.2.1; python_version>='3.8'
```
noklam, line 28: Where are these dependencies coming from? I don't see `mlflow` is imported

kedro-datasets/test_requirements.txt
```diff
@@ -24,13 +24,15 @@ lxml~=4.6
...
 moto==1.3.7; python_version < '3.10'
 moto==3.0.4; python_version == '3.10'
 networkx~=2.4
 opencv-python~=4.5.5.64
 openpyxl>=3.0.3, <4.0
 pandas-gbq>=0.12.0, <0.18.0
-pandas>=1.3  # 1.3 for read_xml/to_xml
+pandas>=1.3, <2  # 1.3 for read_xml/to_xml, <2 for compatibility with Spark < 3.4
```
noklam, line 35: Why do pandas dependency matter here? I expect Pyspark has it own pandas dependency pin already (if needed)

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+)
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql.types import StructType
+from pyspark.sql.utils import AnalysisException, ParseException
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class ManagedTable:  # pylint: disable=R0902
```
noklam, line 24:
```suggestion
class ManagedTable:  # pylint: disable=too-many-instance-attributes
```

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+    json_schema: StructType
+
+    def __post_init__(self):
+        """Run validation methods if declared.
+        The validation method can be a simple check
+        that raises DataSetError.
+        The validation is performed by calling a function named:
+            `validate_<field_name>(self, value) -> raises DataSetError`
+        """
+        for name, _ in self.__dataclass_fields__.items():  # pylint: disable=E1101
```
noklam, line 48:
```suggestion
        for name in self.__dataclass_fields__.keys():  # pylint: disable=no-member
```

noklam (commented):
Thanks for the PR. I cannot manually test it so I mainly focus on reviewing the code itself.

One additional question I have is that this seem to inherit from `AbstractVersionedDataSet`, how does the versioning work since it's a Table if I understand correctly, does it works with the Kedro versioning?

kedro-datasets/test_requirements.txt
jmholzer: True, good spot. These are for a set of MLflow Datasets that are highly related to this one and will get their own PR soon. I'll remove these for now.

kedro-datasets/test_requirements.txt
jmholzer: Pandas >= 2 removed a method that that Spark < 3.4.0 relies on. This was fixed in Spark 3.4.0. With Spark < 3.4.0 and Pandas >= 2, new tests fail.

There are two options for getting tests to pass here:

1. require PySpark >= 3.4.0, or
2. require Pandas < 2.0

I went for option 2, as pinning PySpark ~= 3.4.0 (and therefore Spark 3.4.0) causes a dependency issue with `delta-spark` and causes all new tests to fail (plus Spark tests to fail at the import statements), whereas requiring Pandas < 2.0 caused no other problems with the test suite. Since there is some pressure to get this released and getting our tests to work with Spark 3.4.0 will take some work, I opted to limit Pandas instead.

I would keep it like this, then open an issue to upgrade our Spark to 3.4.0 asap. I can tackle this issue after the Databricks work is delivered.

jmholzer:
> One additional question I have is that this seem to inherit from `AbstractVersionedDataSet`, how does the versioning work since it's a Table if I understand correctly, does it works with the Kedro versioning?

Good question. When delta lake files are modified in any way, a new version of the table is created. Versions of a delta lake table can be accessed either a version number (ascending enum).

`ManagedTableDataset` currently handles versioning by allowing the user to pass in an object of `kedro.io.core.Version` specifying this version number. When the table is modified (by any of the `_save` methods), a new table version is automatically saved in the delta lake file itself. Since versioning is handled by the delta lake format, we do not have to worry about it ourselves.

To my understanding, this is compatible with the`AbstractVersionedDataset` interface, what do you think?

noklam (approved):
thanks for the work!

I think the versioning is fine, is it specific to databricks or any delta table support versioning? (sorry I may get confused with all the delta terminology?) If this is something supported with delta table then we may even extend it to some datasets like SQLtabledataset.

The versioning feature relies on a file path and doesn't work for tables. It relies on the timestamp so basically this dataset implement it's own logic to fetch latest version of dataset.

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+        """validates database name
+
+        Raises:
+            DataSetError: If the table name does not conform to naming constraints.
+        """
+        if self.database:
+            if not re.fullmatch(self._NAMING_REGEX, self.database):
+                raise DataSetError("database does not conform to naming")
+
+    def validate_catalog(self):
```
SajidAlamQB, line 72: These look to be doing the same thing, could we not just take in an argument to determine if it's a table, database or catalog and have 1 method?

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
```diff
@@ -0,0 +1,433 @@
...
+            raise DataSetError(f"`dataframe_type` must be one of {valid_types}")
+
+    def validate_primary_key(self):
+        """validates the primary key of the table
+
+        Raises:
+            DataSetError: If no `primary_key` is specified.
+        """
+        if self.primary_key is None or len(self.primary_key) == 0:
+            if self.write_mode == "upsert":
```
SajidAlamQB, line 112: Why only in upsert mode?

SajidAlamQB (approved):
This looks great, just some minor comments.

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
jmholzer: The upsert operation works by joining on the table using its primary key. The other write modes (`append` and `overwrite`) don't need a primary key to work, as they don't perform this join.

kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
jmholzer: This is a great push. Managed tables are stored in the hierarchy `Catalog` -> `Database` -> `Table`. A given table needs at least a database name and a table name specified, as a catalog name can (I think) be inferred.

I thought about it a bit in response to your comment, and I added a step that also raises a `DatasetError` if no database name or no table name is specified. With these changes it is only `validate_catalog` that has a different structure to the other two namespace validation methods. This being the case, my view is that is clearer to keep these two methods separate rather than DRY them.

PR #234, opened 2023-06-13
fix(datasets): Fix missing pickle setup in `kedro-datasets`

noklam:
## Description

The setup.py missed the `pickle` extras installation.

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #236, opened 2023-06-13
fix(datasets): fix `SparkStreamingDataSet` docstring

noklam:
## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
@@ -49,7 +49,7 @@ class SparkStreamingDataSet(AbstractDataSet):
 
         Args:
             filepath: Filepath in POSIX format to a Spark dataframe. When using Databricks
-                specify ``filepath``s starting with ``/dbfs/``. For messag ebrokers such as
+                specify ``filepath``s starting with ``/dbfs/``. For message brokers such as
                 Kafka and all filepath is not required.
             file_format: File format used during load and save operations. 
                 These are formats supported by the running SparkContext including parquet,
--- a/kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
@@ -35,6 +35,7 @@ class SparkStreamingDataSet(AbstractDataSet):
             schema:
                 filepath: data/01_raw/schema/inventory_schema.json
     """
+
     DEFAULT_LOAD_ARGS = {}  # type: Dict[str, Any]
     DEFAULT_SAVE_ARGS = {}  # type: Dict[str, Any]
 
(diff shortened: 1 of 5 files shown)


astrojuanlu:
This contains commits from gh-234, can be rebased

kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
```diff
@@ -20,15 +20,16 @@ class SparkStreamingDataSet(AbstractDataSet):
     `YAML API <https://kedro.readthedocs.io/en/stable/data/\
     data_catalog.html#use-the-data-catalog-with-the-yaml-api>`_:
     .. code-block:: yaml
+
```
astrojuanlu, line 23: Looks good, did you test it locally? Maybe open a PR on the framework side cherry picking this commit https://github.com/kedro-org/kedro/pull/2485/commits/9c08611a781518f25fc78b8e9fe2ff6dc5c923b2

kedro-datasets/kedro_datasets/spark/spark_streaming_dataset.py
```diff
@@ -20,15 +20,16 @@ class SparkStreamingDataSet(AbstractDataSet):
...
-            checkpoint: data/04_checkpoint/raw_new_inventory
-            header: True
-        load_args:
+          type: spark.SparkStreamingDataSet
+          filepath: data/01_raw/stream/inventory/
+          file_format: json
+          save_args:
+                output_mode: append
+                checkpoint: data/04_checkpoint/raw_new_inventory
+                header: True
```
deepyaman, line 31:
```suggestion
            output_mode: append
            checkpoint: data/04_checkpoint/raw_new_inventory
            header: True
```

noklam: thanks, I don't know why I keep messing up the docstring.

deepyaman (approved):
Nice! I was looking at this yesterday; this seems to cover the two comments Ivan had left on the original PR that weren't addressed. :)

I actually had one more question on this, and maybe you can roll it into the same PR--does it make sense for the "Spark Streaming" docs to be in `spark/README.md`? It only applies to a single dataset implementation, so it's a bit confusing.

But approving in case not sure how to best handle that/want to do it later.

noklam:
@deepyaman Frankly I don't know what's the README.md for, it's not a standard and only exists for very few datasets. I don't believe it will be render in the Kedro's doc either so I would say it's not very important at all.

noklam:
This is still failing - I try to fix the build but it starts complaining the `__init__` method.

astrojuanlu:
> does it make sense for the "Spark Streaming" docs to be in spark/README.md?

Until we significantly rework how our docs are built, we don't have a way to place narrative/prose documentation from this repo on docs.kedro.org (see https://github.com/kedro-org/kedro/issues/2072#issuecomment-1528834875)

noklam:
https://github.com/kedro-org/kedro/pull/2679

I created this PR in `kedro` , so whenever you made changes in this PR you can re-trigger the CI to see if the build is success.

AhdraMeraliQB (approved):

Issue #226, opened 2023-06-06
Support Delta Table with a non-spark implementation


everdark:
## Description

In `kedro-datasets` 1.3.0 a [new dataset dedicated for Delta Table](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/spark/deltatable_dataset.py) has been added, which is based on `pyspark`. Is there any plan to also add support for a non-spark implementation for Delta Lake? If not, I myself with some other folks are happy to contribute. :)

## Context

Not all use cases depend on Spark but they may still need to interact with Delta Lake. It won't be suitable to add Spark as a dependency simply because we want to deal with Delta Tables.

## Possible Implementation

Create a custom dataset that is based on the official Delta Lake Python-binding: https://github.com/delta-io/delta-rs/tree/main/python

noklam:
This is a reasonable request and PR is very welcomed! We have a ticket here for reading Delta via Pandas but the priority isn't high for now. We would prefer it is contributed by the community.

https://github.com/kedro-org/kedro-plugins/issues/159

everdark:
Nice! Will work on this.

merelcht:
Completed in https://github.com/kedro-org/kedro-plugins/issues/226

PR #243, opened 2023-06-19
feat(datasets): Added `DeltaTableDataSet`

afaqueahmad7117:
## Description
This PR adds `DeltaTableDataSet` to the Kedro datasets plugin. Issue [226.](https://github.com/kedro-org/kedro-plugins/issues/226)

## Development notes
1. `DeltaTableDataSet` here `kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py`
2. Unit Tests `kedro-datasets/tests/pandas/test_deltatable_dataset.py`

## Checklist


everdark:
Just a side note, there is a bug for the delta-rs python binding for using catalog: https://github.com/delta-io/delta-rs/issues/1466; so right now I'm not able to smoke test the `from_catalog` method to see if it actually works.

noklam:
@everdark Is this ready to be review now? I saw you have closed the old PR but this one is still in draft.

everdark:
> @everdark Is this ready to be review now? I saw you have closed the old PR but this one is still in draft.

Depending on whether @afaqueahmad7117 is doing some more works. I don't have anything to add from my side as of now. We close the old PR cause there are some mess-up in the commits.

afaqueahmad7117:
@noklam @everdark I'm done with my changes. Marked PR ready for review.

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
```diff
@@ -0,0 +1,258 @@
...
+from typing import Any, Dict, List, Optional
+
+import pandas as pd
+from deltalake import DataCatalog, DeltaTable, Metadata
+from deltalake.exceptions import TableNotFoundError
+from deltalake.writer import write_deltalake
+from kedro.io.core import AbstractDataSet, DataSetError
+
+
+class DeltaTableDataSet(AbstractDataSet):  # pylint:disable=too-many-instance-attributes
```
noklam, line 15: Any chance we can make this a versioned dataset? Is the `DeltaTable` version compatible with Kedro's own timestamp versioning or it has to be a incremental version i.e. 1,2,3 ?

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
```diff
@@ -0,0 +1,258 @@
...
+                ``Azure``:
+                    `az://<container>/<path>`
+                    `adl://<container>/<path>`
+                    `abfs://<container>/<path>`
+                ``GCS``:
+                    `gs://<bucket>/<path>`
+                If any of the prefix above is not provided, `file` protocol (local filesystem)
+                will be used.
+            catalog_type (DataCatalog, optional): `AWS` or `UNITY` if filepath is not provided.
+                Defaults to None.
```
noklam, line 109: What does this catalog_type do?

kedro-datasets/tests/pandas/test_deltatable_dataset.py
```diff
@@ -0,0 +1,147 @@
...
+def deltatable_data_set_from_path(filepath, load_args, save_args, fs_args):
+    return DeltaTableDataSet(
+        filepath=filepath,
+        load_args=load_args,
+        save_args=save_args,
+        fs_args=fs_args,
+    )
+
+
+class TestDeltaTableDataSet:
```
noklam, line 32: Can you create some test for the object storage? You can refer to `CSVDataSet` using `mock_s3` we can have a local test.

noklam:
Noted that the CI is failing for some other reason, https://github.com/kedro-org/kedro/pull/2673 would fix this, so ignore the irrelevant error or try to install a slightly older kedro version to test it locally.

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
afaqueahmad7117: `DeltaTable` can accept both - incremental version numbers i.e. 1, 2, 3 or a `datetime` string as stated here: https://delta-io.github.io/delta-rs/python/usage.html#time-travel

If we make this a versioned dataset, we would be adding `version` to accepted list of attributes in `__init__`. Which `version` attribute would users refer to if they want to load a specific version of `DeltaTable` from Kedro `yml` catalog?  Would that confuse users with `DeltaTable` version?

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
afaqueahmad7117: `catalog_type` loads tables from either `glue` (aws) or `unity` (databricks) if specified, reference here: https://delta-io.github.io/delta-rs/python/usage.html#loading-a-delta-table

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
everdark: IMHO, If we allow it to be also a versioned dataset, we are allowing the user to do a nested versioning structure where the Delta Table is on top of the Kedro version. This may not be useful practically. I think if user want to use Delta Table, they want to leverage only the versioning in Delta table. Adding one more versioning layer to it seems redundant and can create confusion.

kedro-datasets/tests/pandas/test_deltatable_dataset.py
```diff
@@ -0,0 +1,147 @@
...
+        """Test the describe method."""
+        deltatable_data_set_from_path = DeltaTableDataSet(filepath)
+        desc = deltatable_data_set_from_path._describe()
+        assert desc["filepath"] == filepath
+        assert desc["version"] is None
+
+    @patch("kedro_datasets.pandas.deltatable_dataset.DeltaTable")
+    def test_from_aws_glue_catalog(self, mock_delta_table):
+        """Test dataset creation from AWS Glue catalog."""
+        _ = DeltaTableDataSet(catalog_type="AWS", database="db", table="tbl")
```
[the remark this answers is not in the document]
everdark: Changed.

kedro-datasets/setup.py
```diff
@@ -29,6 +29,7 @@ def _collect_requirements(requires):
 networkx_require = {"networkx.NetworkXDataSet": ["networkx~=2.4"]}
 pandas_require = {
     "pandas.CSVDataSet": [PANDAS],
+    "pandas.DeltaTableDataSet": [PANDAS, "deltalake>=0.10.0"],
```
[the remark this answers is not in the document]
afaqueahmad7117: `exceptions` have been introduced from `0.10.0` onwards. We're using `deltalake.exceptions` to `import TableNotFoundError`. So, the older versions below `0.10.0` do not work.

kedro-datasets/tests/pandas/test_deltatable_dataset.py
everdark: It seems that we have difficulty mocking AWS SDK calls from the python side since `deltalake` actually use Rust to Handle the API calls behind the scene. I tried the `moto` way of mocking S3 but it won't work.

everdark: I believe I can mock this `_internal.RawDeltaTable` for sure, but in that case it won't have a proper local boto3 behavior like the one introduced by `moto` to properly simulate the behavior, unless we also code them all. wdyt?

noklam: I see, in that case it's fine to leave the unit tests for now. Did you got a chance to test it manually instead?

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
noklam: I agree, let's keep this as is. Delta's versioning is more powerful than Kedro's one. I will leave the discussion in the dedicated issue. It does creates some inconsistent behavior, for example you can't use the `--load-version` CLI argument if you use Delta, but I guess this is not too important.

- https://github.com/kedro-org/kedro/issues/2703

kedro-datasets/tests/pandas/test_deltatable_dataset.py
everdark: Yes, It works when I test it with my actual AWS account/session. :)

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
```diff
@@ -0,0 +1,258 @@
...
+    def history(self) -> List[Dict[str, Any]]:
+        """Returns the history of actions on DeltaTableDataSet as a list of dictionaries."""
+        return self._delta_table.history()
+
+    def get_loaded_version(self) -> int:
+        """Returns the version of the DeltaTableDataSet that is currently loaded."""
+        return self._delta_table.version()
+
+    def _load(self) -> pd.DataFrame:
+        return self._delta_table.to_pandas()
```
datajoely, line 223: Is there any merit in making this configurable so that in the future we can use one of the other public methods e.g. `to_pyarrow()` or some future addition?

astrojuanlu: Similar issue: https://github.com/ibis-project/ibis/issues/6351

I guess, if this is under the `kedro_datasets.pandas` though, it makes sense to do `.to_pandas()` on load. But I'd love to see more future-proof of interoperability. I think this requires some broader discussion.

noklam (approved):

astrojuanlu:
The new dataset does not yet have 100 % coverage, some lines are missing:

```
kedro_datasets/pandas/deltatable_dataset.py                66      3    95%   153, 211, 216
```

everdark:
> The new dataset does not yet have 100 % coverage, some lines are missing:
> 
> ```
> kedro_datasets/pandas/deltatable_dataset.py                66      3    95%   153, 211, 216
> ```

Coverage increased to 100% in [2b1c5e8](https://github.com/kedro-org/kedro-plugins/pull/243/commits/2b1c5e87f01884072d95c66f45db4685b19c31d4).

astrojuanlu:
Thanks a lot folks, this is looking really good!

Just to help me contextualize and sorry for the basic question: but could you summarize what is the difference between this `DeltaTableDataSet` and the existing `ManagedTableDataSet`? I see the latter can be used with Delta tables too https://kedro.org/blog/managed-delta-tables-kedro-dataset

everdark:
> Thanks a lot folks, this is looking really good!
> 
> Just to help me contextualize and sorry for the basic question: but could you summarize what is the difference between this `DeltaTableDataSet` and the existing `ManagedTableDataSet`? I see the latter can be used with Delta tables too https://kedro.org/blog/managed-delta-tables-kedro-dataset

The `ManagedTableDataSet` is for delta table managed by Databricks Unit Catalog. It is a very specific implementation on top of the open source delta table. Our dataset is a generic non-spark solution to handle the open source version of the delta table.

By adding this into `kedro-datasets`, there will be 3 possible ways of handling delta table:

1. Apacje Spark
2. `delta-rs`, a non-Spark approach (our approach)
3. Databricks Unity Catalog

The user should decide which one best suits their need.

noklam:
> By adding this into kedro-datasets, there will be 3 possible ways of handling delta table:
> Apache Spark
> delta-rs, a non-Spark approach (this PR)
> Databricks Unity Catalog

Agree, although `DeltaTable` is often associated with Spark, but it's actually just a file format and you can read it via Pandas or maybe other libraries later. 

I think the current space is still dominated by Parquet for data processing, Delta files are usually larger due to the compression and version history. IMO the versioning features are quite important and it deserves wider adoption outside of the Spark ecosystem.

I have no idea how the compacting and re-partitioning works with non-spark implementation? This feels like the responsibility of some kind of DB or data processing engine, it's probably too much for the Dataset abstraction. WDYT?

everdark:
> > By adding this into kedro-datasets, there will be 3 possible ways of handling delta table:
> > Apache Spark
> > delta-rs, a non-Spark approach (this PR)
> > Databricks Unity Catalog
> 
> Agree, although `DeltaTable` is often associated with Spark, but it's actually just a file format and you can read it via Pandas or maybe other libraries later.
> 
> I think the current space is still dominated by Parquet for data processing, Delta files are usually larger due to the compression and version history. IMO the versioning features are quite important and it deserves wider adoption outside of the Spark ecosystem.
> 
> I have no idea how the compacting and re-partitioning works with non-spark implementation? This feels like the responsibility of some kind of DB or data processing engine, it's probably too much for the Dataset abstraction. WDYT?

It will be mostly based on the Apache Arrow. It will make sure we don't need to load all files into memory.
Advanced use case can be one leveraging DuckDB as the in-memory query engine, and with Arrow as the backend.

Some examples (in terms of analytical query) can be found here:
https://delta-io.github.io/delta-rs/python/usage.html#querying-delta-tables

SajidAlamQB (approved):
Thanks for your contribution @afaqueahmad7117! Your `DeltaTableDataSet` class looks great.

afaqueahmad7117:
> Thanks for your contribution @afaqueahmad7117! Your DeltaTableDataSet class looks great.

@everdark too!

noklam:
Thank you both! @afaqueahmad7117 @everdark, I will take over from now to fix the conflict and merge.

noklam:
Moving requirements around as we moved from setup.pu -> pyproject.toml

astrojuanlu:
Let's get this merged 🚀 thanks @afaqueahmad7117 @everdark and team!

PR #170, opened 2023-04-16
feat: Polars generic dataset (take 3)

astrojuanlu:
## Description
Close gh-110.

## Development notes
See past discussion in gh-153 and gh-116.

Not 100 % sure I did the rebase right, will look into it later.

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #268, opened 2023-07-13
feat: Upgrade Polars

astrojuanlu:
## Description

While we keep working on https://github.com/kedro-org/kedro-plugins/pull/170, let's upgrade the Polars version to the newest one.

## Checklist


deepyaman (commented):
Why do we need to keep bumping the lower bound? Is there something that prevents us from just using a less-conservative upper bound (and leaving a minimal lower bound)?

astrojuanlu:
At the moment polars is breaking backwards compatibility quite often. But maybe you're right, and we should change this to `~=0.17` 🤔

astrojuanlu:
Actually I'm closing this and ask for a review of https://github.com/kedro-org/kedro-plugins/pull/170 instead, I think the dataset is ready to use and it only needs approvals.

Issue #265, opened 2023-07-11
Support Python 3.11 in all plugins


astrojuanlu:
- [ ] `kedro-airflow`
- [ ] `kedro-docker`
- [ ] `kedro-telemetry`
- [x] `kedro-datasets` #154

astrojuanlu:
`kedro-telemetry` done in #292

AhdraMeraliQB:
Completed by #329 #330 and #332

Issue #154, opened 2023-04-03
Add Python3.11 support for `kedro-datasets`


noklam:
## Description
Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

Pre-requisite of https://github.com/kedro-org/kedro/issues/2270

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

noklam:
I suspect this could be a 5 or more.

https://github.com/kedro-org/kedro/pull/2417#issuecomment-1611645475

I have run into issues. There are some funny interaction between `delta-spark`, `pyspark>3.4`. We may have to update `setup.py`, `test_requirements.txt`, and potentially implementation of some datasets.

One example is SparkDataSet, AnalysisException.desc does not exist anymore >=3.4. I don't think we can drop pyspark <3.4 directly, so we may need to find a compatible solution.

PR #276, opened 2023-07-20
build: loosen the Python version requirement

sbrugman:
## Description
Start for https://github.com/kedro-org/kedro-plugins/issues/265

This won't have any direct impact as all plugins require `kedro` which is not available for 3.11. (That's why I didn't add the `py3.11` test runner)
It does however loosen the circular dependency which we have now.

Frankly, there is little reason not to upgrade asap. 
(Some datasets that are too slow to update can just be only available for 3.10)

Alternatively, remove the upper bound completely

## Checklist

### DIFF
--- a/kedro-airflow/pyproject.toml
+++ b/kedro-airflow/pyproject.toml
@@ -8,7 +8,7 @@ authors = [
     {name = "Kedro"}
 ]
 description = "Kedro-Airflow makes it easy to deploy Kedro projects to Airflow"
-requires-python = ">=3.7, <3.11"
+requires-python = ">=3.7, <3.12"
 license = {text = "Apache Software License (Apache 2.0)"}
 dependencies = [
     "kedro>=0.17.5",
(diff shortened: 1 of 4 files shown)


astrojuanlu (changes requested):
I don't even like having this upper bound in there, some past discussion:

https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special

(and also https://discuss.python.org/t/requires-python-upper-limits/12663?u=astrojuanlu)

So I'd rather evaluate whether we could lift the upper version uncapped for all the plugins, and whether we need something like https://github.com/kedro-org/kedro/pull/2742 for, say, `kedro-datasets` (which is a bit hairy).

sbrugman:
I agree. There isn't anything preventing us from doing so, is there?

astrojuanlu:
`kedro-telemetry` was already lifted in #292, and with `kedro-datasets` we want to be a bit more careful and the team has it on the radar #154 

I'd say let's wait for Kedro 0.18.13 (which will have Python 3.11 support https://github.com/kedro-org/kedro/pull/2851) and then let the team evaluate each plugin case by case. `telemetry` will probably be trivial, `datasets` not so much (although most of the heavy lifting has already been done), and no idea about `docker` and `airflow`.

astrojuanlu:
Better to close this PR and open one per remaining plugin, thanks @sbrugman!

Issue #278, opened 2023-07-23
Bump dask version to the latest


jw-cpnet:
The dask version is pinned to 2021.12.0 and pandas is currently on version 2. 

This dask version doesn't work with pandas 2. 

With any pandas dataframe, if you run following script:

```
ddf = dd.from_pandas(df, npartitions=1)
```

you will get an error:

```
AttributeError: module 'pandas.core.strings' has no attribute 'StringMethods'
```

According to dask's changelog, the latest dask supports pandas 2:

https://docs.dask.org/en/stable/changelog.html

Is there any plan to bump dask version to the latest?

noklam:
I am open to bump dask, would you like to create a PR for that?

jw-cpnet:
sure. https://github.com/kedro-org/kedro-plugins/pull/282

PR #282, opened 2023-07-27
feat(datasets): bump dask version to the latest

jw-cpnet:
## Description
The latest pandas 2 is not compatible with old version of dask so we need to bump it to the latest to support pandas 2.

## Checklist


kedro-datasets/pyproject.toml
```diff
@@ -21,7 +21,7 @@ api-apidataset = ["requests~=2.20"]
 biosequence = ["kedro-datasets[biosequence.BioSequenceDataSet]"]
 biosequence-biosequencedataset = ["biopython~=1.73"]
 dask = ["kedro-datasets[dask.ParquetDataSet]"]
-dask-parquetdataset = ["dask[complete]~=2021.10", "triad>=0.6.7, <1.0"]
+dask-parquetdataset = ["dask[complete]~=2023.7", "triad>=0.6.7, <1.0"]
```
astrojuanlu, line 24: What about a more flexible pinning, like

```suggestion
dask-parquetdataset = ["dask[complete]>=2021.10", "triad>=0.6.7, <1.0"]
```

? Dask doesn't follow SemVer https://github.com/dask/community/issues/100 so this pinning is not very useful - breaking changes might be introduced at any time. We can shift the onus to the user on maintaining an appropriate pinning if something breaks. Which I don't expect to, because we use pretty fundamental Dask APIs anyway.

astrojuanlu (commented):
Thanks for your pull request @jw-cpnet ! Left a comment

kedro-datasets/pyproject.toml
jw-cpnet: Thank you for your suggestion. I agree that a more flexible pinning approach would be beneficial given Dask's versioning policy. Shifting the responsibility to the user for maintaining an appropriate pinning seems reasonable.

astrojuanlu:
Fixes #278

noklam (approved):

noklam:
I think we are good here, re-requesting @astrojuanlu for approval.

astrojuanlu (approved):
🚀

PR #180, opened 2023-04-18
feat: Dataset API add `save` method

McDonnellJoseph:
## Description
We sometimes want to save some data with a REST Api (say with Django Rest Framework), most of the time issuing a POST request. So far, the APIDataSet is read only, with the _save method raising a DataSetError, we'd want to extend it to make a request.

Closes kedro-org/kedro-plugins#166 

## Development notes
If applied, this commit will:

1. Enable save method for APIDataSet, by sending packets of size chunk_size, default is 100 to server. Default HTTP method is POST and returns a request.Response object.
2. Add, save_args dictionary to start conforming with other datasets.
3. Raise error when trying to save non-json data.
4. Raise error for save in case of HTTP error.

## Checklist

- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the relevant `RELEASE.md` file
- [x] Added tests to cover my changes

PR #301, opened 2023-08-10
fix(datasets): do not double encode the data as json when saving an A…

FlorianGD:
APIDataSet

## Description
When saving an `APIDataSet`, the data was doubly encoded as json (making it a string instead of a dict for example), once by us and once by the `requests` library. This PR fixes that.

## Development notes
The `json` parameter of `requests.request` is now a python object instead of a string. The tests have been updated to catch the problem.

## Checklist


noklam:
Thanks @FlorianGD Is this a bug or it has been always like this? I suspect this is introduced in https://github.com/kedro-org/kedro-plugins/pull/180. Is it only affecting `load` method? since `save` was only added recently I assume this doesn't affect `load`

noklam:
This looks like a good fix to me. Thanks @FlorianGD

noklam (approved):

PR #302, opened 2023-08-11
fix(datasets): Correct pyproject.toml syntax for optional dependencies

DimedS:
## Description
#263 modified optional-dependencies in pyproject.toml, leading to issues with dependency loading for `spark` when running `kedro-datasets[spark.SparkDataSet, pandas.ParquetDataSet]` (as seen in the iris-databricks starter). This PR corrects the pyproject.toml syntax to resolve the issue.

## Development notes
Tested manually using the iris-databricks starter only.

## Checklist

### DIFF
--- a/kedro-datasets/pyproject.toml
+++ b/kedro-datasets/pyproject.toml
@@ -25,10 +25,7 @@ plotly-base = ["plotly>=4.8.0, <6.0"]
 polars-base = ["polars~=0.17.0"]
 s3fs-base = ["s3fs>=0.3.0, <0.5"]
 spark-base = ["pyspark>=2.2, <4.0"]
-"spark.DeltaTableDataSet" = [
-  "kedro-datasets[spark-base,hdfs-base,s3fs-base]",
-  "delta-spark>=1.0, <3.0"
-]
+
 
 # Datasets dependencies
 api = ["kedro-datasets[api.APIDataSet]"]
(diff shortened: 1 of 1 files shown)


merelcht (approved):
LGTM! 👍

noklam:
Looks good to me in general. I want to do a manual test and re-order some of the dependencies to reflect the structures. I will commit it directly since git doesn't allow to comment on unchanged code.

noklam (approved):
Tested with `pip install .[spark.SparkDataSet]` and it works as expected. 

I made a few changes:
1. reorder some dependencies to reflect the alphabetical order
2. deleted `polars-base` and `delta-base` since this is inconsistent, they are only used by 1 dataset. If we need base for each library we should have `pillow-base` etc, I don't think we should include this unless necessary.
3. I move all base to the top of the `optional-dependencies ` to reflect the structure.

The use of `xxxx-base` is not very consistent in current settings. i.e. `polars-base` is only used in one dataset, I don't think

noklam:
@deepyaman IMO this is not a breaking change, but rather a bug fix since we never intend to release the `spark-dataset` syntax. We didn't avoid this problem completely because `kedro-datasets[s3-base]` still exist, but I don't see better workaround.

astrojuanlu:
I also see this as a bug fix, these names were working before

Issue #306, opened 2023-08-15
kedro-datasets 1.5.2 breaks `pip install kedro-datasets[all]`


noklam:
## Description
Short description of the problem here.
doing `pip install kedro-dataset[all]==1.5.2` will get these warnings.

> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘api-apidataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘biosequence-biosequencedataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘dask-parquetdataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘databricks-managedtabledataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘geopandas-geojsondataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘holoviews-holoviewswriter’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘matplotlib-matplotlibwriter’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘networkx-networkxdataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘pickle-pickledataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘pillow-imagedataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘plotly-jsondataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘plotly-plotlydataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘polars-csvdataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘redis-pickledataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘snowflake-snowparktabledataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘spark-deltatabledataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘spark-sparkdataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘spark-sparkhivedataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘spark-sparkjdbcdataset’
> WARNING: kedro-datasets 1.5.2 does not provide the extra ‘svmlight-svmlightdataset’
[...]

PR #307, opened 2023-08-15
fix(datasets): Add optional dependencies duplication in pyproject.toml

DimedS:
## Description
This PR serves as an interim solution to the issue of nesting with "." in the pyproject.toml file. All lines have been duplicated following this pattern:
```
"pandas.JSONDataSet" = ["kedro-datasets[pandas-base]"]
"pandas-jsondataset" = ["kedro-datasets[pandas-base]"]
```
It appears the long-term solution might involve aligning names with the packaging standards, as mentioned by @astrojuanlu in [this comment](https://github.com/kedro-org/kedro-plugins/issues/306#issuecomment-1680094495), or transitioning from setuptools to poetry.

## Development notes
Manually tested: works well with .[all] and .[spark.SparkDataSet].
It seems we should introduce tests to automate the verification of dependency installations.

## Checklist

### DIFF
--- a/kedro-datasets/pyproject.toml
+++ b/kedro-datasets/pyproject.toml
@@ -28,27 +28,35 @@ spark-base = ["pyspark>=2.2, <4.0"]
 # Datasets dependencies
 api = ["kedro-datasets[api.APIDataSet]"]
 "api.APIDataSet" = ["requests~=2.20"]
+"api-apidataset" = ["requests~=2.20"]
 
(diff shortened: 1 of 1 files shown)


astrojuanlu (commented):
I'm OK with this approach as long as we are fine with keeping the new extras we're introducing in the long run. In other words: at some point in the future, we settle with the new syntax and tell our users to do `pip install kedro-datasets[pandas-csvdataset]`. See more context in my comment here https://github.com/kedro-org/kedro-plugins/issues/306#issuecomment-1680094495

PR #297, opened 2023-08-10
feat(datasets): Add Support Python 3.11 for kedro-datasets

SajidAlamQB:
## Description

Add Python 3.11 support for Kedro-datasets

## Development notes

- Updated Github actions with new job that checks if `plugin == kedro-datasets` and adds `3.11` to the setup matrix if it does. This helps unit tests to run on python `3.11` only for `kedro-datasets`.

## Checklist

- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #308, opened 2023-08-16
chore: Make kedro-datasets open-bound

SajidAlamQB:
## Description

Add open-bound to kedro-datasets to unblock work on adding python 3.11 support to kedro-datasets, https://github.com/kedro-org/kedro-plugins/pull/297.

## Checklist


noklam:
Why do we need a separate PR to remove the bound? Can we do it in the same PR.

SajidAlamQB:
> Why do we need a separate PR to remove the bound? Can we do it in the same PR.

It initially looked like the dependencies were failing because of the open-bound not being committed this has since been fixed in the PR so will close this.

PR #310, opened 2023-08-17
fix(datasets): Revert the optional dependencies back to setup.py

DimedS:
## Description
This PR transfers optional dependencies from `pyproject.toml` back to `setup.py`.

The naming conventions for `kedro-datasets` extras do not adhere to Python packaging standards. This misalignment leads to issues with nesting in `pyproject.toml` optional dependencies. Our solution is two-steps:

- In the first step, we'll revert the optional dependencies back to `setup.py`.
- In the second step, we'll modify the naming syntax in the breaking release 2.0.0, coinciding with the release of kedro 0.19.

## Development notes
Manually tested: works well with .[all] and .[spark.SparkDataSet].

## Checklist

### DIFF
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -1,0 +1,210 @@
+from itertools import chain
+
+from setuptools import setup
+
+# at least 1.3 to be able to use XMLDataSet and pandas integration with fsspec
+PANDAS = "pandas>=1.3, <3.0"
+SPARK = "pyspark>=2.2, <4.0"
+HDFS = "hdfs>=2.5.8, <3.0"
+S3FS = "s3fs>=0.3.0, <0.5"
+POLARS = "polars~=0.17.0"
+DELTA = "delta-spark~=1.2.1"
+
+
+def _collect_requirements(requires):
+    return sorted(set(chain.from_iterable(requires.values())))
+
+
+api_require = {"api.APIDataSet": ["requests~=2.20"]}
+biosequence_require = {"biosequence.BioSequenceDataSet": ["biopython~=1.73"]}
+dask_require = {
+    "dask.ParquetDataSet": ["dask[complete]>=2021.10", "triad>=0.6.7, <1.0"]
+}
+databricks_require = {"databricks.ManagedTableDataSet": [SPARK, PANDAS, DELTA]}
+geopandas_require = {
+    "geopandas.GeoJSONDataSet": ["geopandas>=0.6.0, <1.0", "pyproj~=3.0"]
+}
+holoviews_require = {"holoviews.HoloviewsWriter": ["holoviews~=1.13.0"]}
+matplotlib_require = {"matplotlib.MatplotlibWriter": ["matplotlib>=3.0.3, <4.0"]}
+networkx_require = {"networkx.NetworkXDataSet": ["networkx~=2.4"]}
+pandas_require = {
+    "pandas.CSVDataSet": [PANDAS],
+    "pandas.ExcelDataSet": [PANDAS, "openpyxl>=3.0.6, <4.0"],
+    "pandas.DeltaTableDataSet": [PANDAS, "deltalake>=0.10.0"],
+    "pandas.FeatherDataSet": [PANDAS],
+    "pandas.GBQTableDataSet": [PANDAS, "pandas-gbq>=0.12.0, <0.18.0"],
+    "pandas.GBQQueryDataSet": [PANDAS, "pandas-gbq>=0.12.0, <0.18.0"],
+    "pandas.HDFDataSet": [
+        PANDAS,
+        "tables~=3.6.0; platform_system == 'Windows'",
+        "tables~=3.6; platform_system != 'Windows'",
+    ],
+    "pandas.JSONDataSet": [PANDAS],
+    "pandas.ParquetDataSet": [PANDAS, "pyarrow>=6.0"],
+    "pandas.SQLTableDataSet": [PANDAS, "SQLAlchemy>=1.4, <3.0"],
+    "pandas.SQLQueryDataSet": [PANDAS, "SQLAlchemy>=1.4, <3.0", "pyodbc~=4.0"],
+    "pandas.XMLDataSet": [PANDAS, "lxml~=4.6"],
+    "pandas.GenericDataSet": [PANDAS],
+}
+pickle_require = {"pickle.PickleDataSet": ["compress-pickle[lz4]~=2.1.0"]}
+pillow_require = {"pillow.ImageDataSet": ["Pillow~=9.0"]}
+plotly_require = {
+    "plotly.PlotlyDataSet": [PANDAS, "plotly>=4.8.0, <6.0"],
+    "plotly.JSONDataSet": ["plotly>=4.8.0, <6.0"],
+}
+polars_require = {"polars.CSVDataSet": [POLARS]}
+redis_require = {"redis.PickleDataSet": ["redis~=4.1"]}
+snowflake_require = {
+    "snowflake.SnowparkTableDataSet": [
+        "snowflake-snowpark-python~=1.0.0",
+        "pyarrow~=8.0",
+    ]
+}
+spark_require = {
+    "spark.SparkDataSet": [SPARK, HDFS, S3FS],
+    "spark.SparkHiveDataSet": [SPARK, HDFS, S3FS],
+    "spark.SparkJDBCDataSet": [SPARK, HDFS, S3FS],
+    "spark.DeltaTableDataSet": [SPARK, HDFS, S3FS, "delta-spark>=1.0, <3.0"],
+}
+svmlight_require = {"svmlight.SVMLightDataSet": ["scikit-learn~=1.0.2", "scipy~=1.7.3"]}
+tensorflow_require = {
+    "tensorflow.TensorFlowModelDataSet": [
+        # currently only TensorFlow V2 supported for saving and loading.
+        # V1 requires HDF5 and serialises differently
+        "tensorflow~=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
+        # https://developer.apple.com/metal/tensorflow-plugin/
+        "tensorflow-macos~=2.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    ]
+}
+video_require = {"video.VideoDataSet": ["opencv-python~=4.5.5.64"]}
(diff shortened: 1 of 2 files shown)


.github/workflows/check-plugin.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Install dependencies
         run: |
           cd ${{ inputs.plugin }}
-          pip install ".[test]"
+          pip install . -r test_requirements.txt  # TODO(deepyaman): Define `test` extra and `pip install .[test]`
```
astrojuanlu, line 45: I'm wondering if we could keep this one? Hence retain the `test` dependencies in the project metadata file (will be `setup.py` in this PR), rather than going back to a separate `test_requirements.txt`.

kedro-datasets/setup.py
```diff
@@ -0,0 +1,144 @@
+from itertools import chain
```
astrojuanlu, line 1: Looks good on a quick first inspection but would like to give a closer one on the second round.

astrojuanlu (commented):
Thanks for the PR @DimedS 🙏🏽 I think we should have been more explicit though in that we wanted to revert #263 only for `kedro-datasets` (which is the one that has the problematic recursive dependencies). All the other plugins are fine as they were, I believe.

So, I think this PR should mostly (1) bring back `kedro-datasets/setup.py` (you already did), (2) remove the `optional-dependencies` table from `kedro-datasets/pyproject.toml` (you already did), (3) and not much else. If we could not bring back `kedro-datasets/test_requirements.txt` that would be optimal, I'd say.

.github/workflows/check-plugin.yml
DimedS: Yes, it seems we need to modify our approach. We should not only revert but transfer the `test_requirements.txt` to `setup.py` instead of `pyproject.toml`. I can't simply revert changes for only kedro-datasets, as it would lead to inconsistencies with test_requirements. Currently, I've caused a disruption in the end-to-end test with RTD, since it was previously assumed that we could use ".[test]".

DimedS:
I've decided it's simpler not to revert, but instead to transfer the optional dependencies from `pyproject.toml` to `setup.py.`

noklam (approved):
Looks fine to me, I haven't tested but I see this is already manually tested.

astrojuanlu (commented):
I confirm that `pip install "kedro-datasets[all] @ git+https://github.com/kedro-org/kedro-plugins@306-kedro-datasets-152-breaks-v2#subdirectory=kedro-datasets"` works 👍🏽 as well as `[spark]` (as reported in https://linen-slack.kedro.org/t/14154135/hi-i-am-a-bit-confused-with-the-inclusion-of-optional-depend#7da8695b-cba2-4386-8e62-afe59eb7efec).

Giving this a visual inspection before the final approval.

astrojuanlu:
Complete diff between pre-migration and now in `setup.py` and `pyproject.toml`. Everything looks good 👍🏽 

```diff
❯ git diff 884c1fe..HEAD -- kedro-datasets/setup.py kedro-datasets/pyproject.toml                                                            (kedro310-dev) 
diff --git a/kedro-datasets/pyproject.toml b/kedro-datasets/pyproject.toml
index cfa32ba..31c66c2 100644
--- a/kedro-datasets/pyproject.toml
+++ b/kedro-datasets/pyproject.toml
@@ -12,6 +12,7 @@ requires-python = ">=3.7, <3.11"
 license = {text = "Apache Software License (Apache 2.0)"}
 dependencies = [
     "kedro>=0.16",
+    "lazy_loader",
 ]
 dynamic = ["readme", "version", "optional-dependencies"]
 
@@ -39,23 +40,31 @@ load-plugins = [
 ]
 extension-pkg-whitelist = "cv2"
 unsafe-load-any-extension = false
+
 [tool.pylint.messages_control]
 disable = [
     "ungrouped-imports",
     "duplicate-code",
-    "too-many-instance-attributes"
+    "too-many-instance-attributes",
+    "too-few-public-methods",  # https://github.com/pylint-dev/pylint/issues/8865
 ]
 enable = ["useless-suppression"]
+
 [tool.pylint.refactoring]
 max-nested-blocks = 5
+
 [tool.pylint.format]
-indent-after-paren=4
-indent-string="    "
+# Regexp for a line that is allowed to be longer than the limit.
+ignore-long-lines='^\s*(# )?<?https?://\S+>?$'
+indent-after-paren = 4
+indent-string = "    "
+
 [tool.pylint.miscellaneous]
 notes = [
     "FIXME",
     "XXX"
 ]
+
 [tool.pylint.design]
 min-public-methods = 1
 
@@ -66,7 +75,7 @@ omit = ["tests/*", "kedro_datasets/holoviews/*", "kedro_datasets/snowflake/*"]
 exclude_lines = ["pragma: no cover", "raise NotImplementedError"]
 
 [tool.pytest.ini_options]
-addopts="""
+addopts = """
 --cov-report xml:coverage.xml \
 --cov-report term-missing \
 --cov kedro_datasets \
diff --git a/kedro-datasets/setup.py b/kedro-datasets/setup.py
index e55e421..63cd364 100644
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -18,7 +18,7 @@ def _collect_requirements(requires):
 api_require = {"api.APIDataSet": ["requests~=2.20"]}
 biosequence_require = {"biosequence.BioSequenceDataSet": ["biopython~=1.73"]}
 dask_require = {
 api_require = {"api.APIDataSet": ["requests~=2.20"]}
 biosequence_require = {"biosequence.BioSequenceDataSet": ["biopython~=1.73"]}
 dask_require = {
-    "dask.ParquetDataSet": ["dask[complete]~=2021.10", "triad>=0.6.7, <1.0"]
+    "dask.ParquetDataSet": ["dask[complete]>=2021.10", "triad>=0.6.7, <1.0"]
 }
 databricks_require = {"databricks.ManagedTableDataSet": [SPARK, PANDAS, DELTA]}
 geopandas_require = {
@@ -30,6 +30,7 @@ networkx_require = {"networkx.NetworkXDataSet": ["networkx~=2.4"]}
 pandas_require = {
     "pandas.CSVDataSet": [PANDAS],
     "pandas.ExcelDataSet": [PANDAS, "openpyxl>=3.0.6, <4.0"],
+    "pandas.DeltaTableDataSet": [PANDAS, "deltalake>=0.10.0"],
     "pandas.FeatherDataSet": [PANDAS],
     "pandas.GBQTableDataSet": [PANDAS, "pandas-gbq>=0.12.0, <0.18.0"],
     "pandas.GBQQueryDataSet": [PANDAS, "pandas-gbq>=0.12.0, <0.18.0"],
@@ -137,6 +138,72 @@ extras_require["docs"] = [
     "myst-parser~=1.0.0",
     "Jinja2<3.1.0",
 ]
+extras_require["test"] = [
+    "adlfs>=2021.7.1, <=2022.2",
+    "bandit>=1.6.2, <2.0",
+    "behave==1.2.6",
+    "biopython~=1.73",
+    "blacken-docs==1.9.2",
+    "black~=22.0",
+    "compress-pickle[lz4]~=1.2.0",
+    "coverage[toml]",
+    "dask[complete]",
+    "delta-spark~=1.2.1",
+    # 1.2.0 has a bug that breaks some of our tests: https://github.com/delta-io/delta/issues/1070
+    "deltalake>=0.10.0",
+    "dill~=0.3.1",
+    "filelock>=3.4.0, <4.0",
+    "gcsfs>=2021.4, <=2022.1",
+    "geopandas>=0.6.0, <1.0",
+    "hdfs>=2.5.8, <3.0",
+    "holoviews~=1.13.0",
+    "import-linter[toml]==1.2.6",
+    "ipython>=7.31.1, <8.0",
+    "Jinja2<3.1.0",
+    "joblib>=0.14",
+    "jupyterlab~=3.0",
+    "jupyter~=1.0",
+    "lxml~=4.6",
+    "matplotlib>=3.0.3, <3.4; python_version < '3.10'",  # 3.4.0 breaks holoviews
+    "matplotlib>=3.5, <3.6; python_version == '3.10'",
+    "memory_profiler>=0.50.0, <1.0",
+    "moto==1.3.7; python_version < '3.10'",
+    "moto==3.0.4; python_version == '3.10'",
+    "networkx~=2.4",
+    "opencv-python~=4.5.5.64",
+    "openpyxl>=3.0.3, <4.0",
+    "pandas-gbq>=0.12.0, <0.18.0",
+    "pandas>=1.3, <2",  # 1.3 for read_xml/to_xml, <2 for compatibility with Spark < 3.4
+    "Pillow~=9.0",
+    "plotly>=4.8.0, <6.0",
+    "polars~=0.15.13",
+    "pre-commit>=2.9.2, <3.0",  # The hook `mypy` requires pre-commit version 2.9.2.
+    "psutil==5.8.0",
+    "pyarrow~=8.0",
+    "pylint>=2.5.2, <3.0",
+    "pyodbc~=4.0.35",
+    "pyproj~=3.0",
+    "pyspark>=2.2, <4.0",
+    "pytest-cov~=3.0",
+    "pytest-mock>=1.7.1, <2.0",
+    "pytest-xdist[psutil]~=2.2.1",
+    "pytest~=7.2",
+    "redis~=4.1",
+    "requests-mock~=1.6",
+    "requests~=2.20",
+    "s3fs>=0.3.0, <0.5",  # Needs to be at least 0.3.0 to make use of `cachable` attribute on S3FileSystem.
+    "scikit-learn~=1.0.2",
+    "scipy~=1.7.3",
+    "snowflake-snowpark-python~=1.0.0; python_version == '3.8'",
+    "SQLAlchemy>=1.4, <3.0",
+    # The `Inspector.has_table()` method replaces the `Engine.has_table()` method in version 1.4.
+    "tables~=3.7",
+    "tensorflow-macos~=2.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
+    "tensorflow~=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
+    "triad>=0.6.7, <1.0",
+    "trufflehog~=2.1",
+    "xlsxwriter~=1.0",
+]
 
 setup(
     extras_require=extras_require,
```

astrojuanlu (approved):
The "kedro-datasets does not provide extra" messages are gone from https://readthedocs.org/projects/kedro/builds/21641718/, so I'm approving.

astrojuanlu:
And I confirm docs are fixed https://kedro--2944.org.readthedocs.build/en/2944/ release of kedro-datasets can proceed after this PR lands in `main` 👍🏽

PR #259, opened 2023-07-05
chore: Move pre-commit config to the top level of kedro-plugins

SajidAlamQB:
## Description

related to: https://github.com/kedro-org/kedro/issues/2337

Pre-commit configuration is maintained per-plugin so we should move pre-commit config to the top level and use path filters if/when necessary.

Not ideal because part of the idea of a monorepo is that configuration for consistent styling, etc. can be shared.

## Development notes

I've been running into several issues with pre-commit, most stem from the fact that we can't dynamically run the hooks depending on the plugin environment we are in on the CI. Pre-commit runs every hook resulting in errors appearing in test environments not set up for a certain plugin.

Update:

Updated `make lint` it will now run each individual pre-commit hook with `&&` the drawback from this is that it will stop running if one hook fails. The alternative is to use `;` but CI won't detect if a hook fails so I've gone with `&&`. This also means we can go back to using `make plugin=${{ inputs.plugin }} lint` in `check-plugins.yml`.

```
lint:
	pre-commit run trailing-whitespace --all-files && pre-commit run end-of-file-fixer --all-files && pre-commit run check-yaml --all-files && pre-commit run check-added-large-files --all-files && pre-commit run check-case-conflict --all-files && pre-commit run check-merge-conflict --all-files && pre-commit run debug-statements --all-files && pre-commit run requirements-txt-fixer --all-files && pre-commit run flake8 --all-files && pre-commit run isort-$(plugin) --all-files --hook-stage manual && pre-commit run black-$(plugin) --all-files --hook-stage manual && pre-commit run secret_scan --all-files --hook-stage manual && pre-commit run bandit --all-files --hook-stage manual && pre-commit run pylint-$(plugin) --all-files --hook-stage manual && pre-commit run pylint-$(plugin)-features --all-files --hook-stage manual && pre-commit run pylint-$(plugin)-tests --all-files --hook-stage manual
```

## Checklist

- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #311, opened 2023-08-17
add ruff config in pyproject.toml

noklam:
## Checklist

### DIFF
--- a/kedro-airflow/pyproject.toml
+++ b/kedro-airflow/pyproject.toml
@@ -71,3 +71,18 @@ fail_under = 100
 show_missing = true
 omit = ["tests/*"]
 exclude_lines = ["pragma: no cover", "raise NotImplementedError"]
+
+[tool.ruff]
+line-length = 88
+show-fixes = true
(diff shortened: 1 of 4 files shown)


noklam:
cc @SajidAlamQB noted kedro has enabled ruff already. I started this but feel free to takeover.

SajidAlamQB:
> cc @SajidAlamQB noted kedro has enabled ruff already. How should this adapt to the monorepo? Why do we hardcode the pre-commit run in `Makefile` now? I started this but feel free to takeover.

We used to run the pre-commit like this in the Makefile:

`cd $(plugin) && pre-commit run -a --hook-stage manual`

But since https://github.com/kedro-org/kedro-plugins/pull/259 moved all pre-commit files to top level,  pre-commit would run every hook resulting in errors appearing in test environments not set up for a certain plugin. It doesn't have a way to dynamically run the hooks depending on the plugin environment we are on. Hence the hardcoded pre-commit hooks in the `Makefile`.

merelcht:
Completed in https://github.com/kedro-org/kedro-plugins/pull/354

Issue #274, opened 2023-07-19
`kedro-airflow` convert all pipelines


sbrugman:
## Description
In the current setup, one needs to write a script to convert all pipelines at once, e.g. 
```bash
kedro registry list  | grep '^-' | sed "s/^- //g" >> ./pipelines.txt
cat ./pipelines.txt | while read line;
do 
    kedro airflow create --pipeline "$line" --jinja-file ./resources/airflow/airflow_dag_template.py.j2 --target-dir dags/; 
done
rm ./pipelines.txt
```

It would be useful if `kedro-airflow` took care of this.

Are you open to a PR for this?

## Context
In practice we would like to have all pipelines available in Airflow in order to run them with different configurations, or on individual schedules. For instance, some pipelines are heavier than others and can be ran less often.

## Possible Implementation
We could add another parameter, e.g. `--all` that is exclusive with the `--pipeline` argument

## Possible Alternatives
Currently, the default argument is `__default__`.
We could choose to make the default argument `None`, and if left unspecified to convert all pipelines.

Alternatively, add a command that wraps the `kedro airflow create` command.
`kedro airflow create-all`

astrojuanlu:
> We could add another parameter, e.g. `--all` that is exclusive with the `--pipeline` argument

Sounds good to me! @amandakys what do you think from a DX perspective?

sbrugman:
I'll submit a PR

PR #335, opened 2023-09-11
feat: Kedro-Airflow convert all pipelines option

sbrugman:
## Description
Closes https://github.com/kedro-org/kedro-plugins/issues/274

## Development notes
The text output of the command was outdated and verbose: 7 lines per DAG. Some of the text is moved to the Readme, other lines are discarded as they seem not relevant (about paths in the template). If needed, we can always choose to add them to the documentation instead of the CLI output.

## Checklist


sbrugman:
Let me know if there is anything to do on my side to move this forward.

ankatiyar:
Hi @sbrugman, I wanted to ask what the difference is between just running `kedro airflow create` with no `--pipeline` filter vs doing it with `--all`?

kedro-airflow/kedro_airflow/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Kedro plugin for running a project with Airflow."""
 
-__version__ = "0.6.0"
+__version__ = "0.6.1"
```
ankatiyar, line 3: Could you revert this change? It'll create a new release and we have a separate release process.

sbrugman:
> Hi @sbrugman, I wanted to ask what the difference is between just running `kedro airflow create` with no `--pipeline` filter vs doing it with `--all`?

The default argument for `--pipeline` is the pipeline `__default__`. With `--all`, the conversion takes place for all registered pipelines, which includes `__default__`, and others such as `data_science` and `data_engineering` in some of the starter templates. The `--all` flags generate a DAG for each. This can be used for instance to run them on a different frequency.

I've experimented with changing the default of `--pipeline` to generate all, since users can still use `--pipeline __default__`. Apart from a breaking change, this also seemed unintuitive, so I decided on this approach.

kedro-airflow/kedro_airflow/__init__.py
sbrugman: Will do!

ankatiyar:
> The default argument for `--pipeline` is the pipeline `__default__`. With `--all`, the conversion takes place for all registered pipelines, which includes `__default__`, and others such as `data_science` and `data_engineering` in some of the starter templates. The `--all` flags generate a DAG for each. This can be used for instance to run them on a different frequency.
> 
> I've experimented with changing the default of `--pipeline` to generate all, since users can still use `--pipeline __default__`. Apart from a breaking change, this also seemed unintuitive, so I decided on this approach.

Could this then not be achieved by allowing a `--pipeline` to accept lists : `kedro airflow create --pipeline=<list all pipelines>` or updating `--pipeline` filter to accept `--pipeline=all` without changing the defaults. What do you think @merelcht @noklam?

sbrugman:
Interesting points.

Accepting a list would be additionally interesting, however would not solve the problem described in the issue.
I could add it anyway, but as extra feature, not as replacement.

`--pipeline=all` would conflict if the user has a pipeline named `all`

astrojuanlu:
I agree with `--pipeline=all` being problematic. What about `--pipeline=*`? Although this does not (yet) exist on Kedro core, for example `kedro run --pipeline=*`.

sbrugman:
Arguments against `--all` are that it introduces a new option and in itself does not refer to `--pipeline`.
Is `*` a valid pipeline name? In that case it could conflict in theory, even though at a low probability.
Reasons for still going with `--all` is that it does not add additional reserved words, nor does it require users to get familiar with the docs: it will just show up in `kedro airflow -h`.
Now that `--pipeline` will support multiple arguments, it will also require less logic to filter out cases such as `--pipeline=de --pipeline=:all:`.

No strong preference on the selected option from my side, as long as we can move forward: this feature will simplify the conversion logic quite a bit.

astrojuanlu:
IIUC, pipeline names must be valid Python identifiers, because they're derived from the subpackage name. Hence both `*` and `:all:` are illegal pipeline names.

I'm also okay with either `--pipeline=:all:` (I suggested this because there was the pip precedent, but I always found it a bit obscure) or adding a separate CLI option `--all`, no strong preference. @ankatiyar @merelcht  what do you think?

amandakys:
Just weighing in, the options that I can see presented here are: 

1. `kedro airflow create --all`
2. `kedro airflow create --pipeline=:all:`
3. `kedro airflow create --pipeline=*`

These commands would create multiple DAG files, one for each registered pipeline. 

My first instinct is that `--all` is the most intuitive solution, if we see the `create` command as describing the functionality of creating DAG files and the `--pipeline` command as taking <pipeline-name> inputs to specify what to target. I don't feel that the lack of `pipeline` in the command would be confusing as the create commands description would be "create DAG files for given pipelines" 

In this vein, I think allowing `--pipeline` to take list input would complement `--all` well. As per our work on plural commands (https://github.com/kedro-org/kedro/issues/2244), I would recommend renaming the command `--pipelines` to indicate the possibility of multiple pipeline inputs. 

On the other hand, we currently have in progress the command `kedro new --add-ons=all` which is similar in concept, but since we control the legal inputs, `all` without modification is okay. With this in mind, Option 3 seems the most consistent, and could become a convention that we extend across our CLI whenever we allow multiple/list inputs to a command and want to allow a shortcut to inputting all values. 

IMO if we aren't interested in rolling out this * syntax and/or see a use case for this (i couldn't find any on a quick skim(, then for this specific issue I would favour the `--all` solution, as it seems the tidiest. 

Some questions I had: 
re: @sbrugman's comment about `--all` being better for discoverability in the CLI as it will come up in `kedro airflow -h` in the case of option 2/3, would it be possible to indicate these special inputs to `--pipeline` in the help interface?

sbrugman:
Extended the CLI option to take multiple `--pipeline` arguments.

For completeness, the help output now is:
```
Usage: Kedro-Airflow airflow create [OPTIONS]

  Create an Airflow DAG for a project

Options:
  -p, --pipeline, --pipelines TEXT
                                  Name of the registered pipeline to convert. If
                                  not set, the '__default__' pipeline is used.
                                  This argument supports passing multiple values
                                  using `--pipeline [p1] --pipeline [p2]`. Use
                                  the `--all` flag to convert all registered
                                  pipelines at once.
  --all                           Convert all registered pipelines at once.
  -e, --env TEXT                  Kedro configuration environment name. Defaults
                                  to `local`.
  -t, --target-dir DIRECTORY      The directory path to store the generated
                                  Airflow dags
  -j, --jinja-file FILE           The template file for the generated Airflow
                                  dags
  --params TEXT                   Specify extra parameters that you want to pass
                                  to the context initialiser. Items must be
                                  separated by comma, keys - by colon or equals
                                  sign, example: param1=value1,param2=value2.
                                  Each parameter is split by the first comma, so
                                  parameter values are allowed to contain
                                  colons, parameter keys are not. To pass a
                                  nested dictionary as parameter, separate keys
                                  by '.', example: param_group.param1:value1.
  --help                          Show this message and exit.
```

sbrugman:
Resolved the merge conflict (pylint)

astrojuanlu (approved):
Love the new docstring, and happy with the `--all` flag 💯

kedro-airflow/RELEASE.md
```diff
@@ -1,6 +1,15 @@
 # Upcoming Release
 * Added support for Python 3.11
 
+# Release 0.6.1
```
ankatiyar, line 4: This heading can be removed and the new additions should be under the # Upcoming Release heading. We change these headings from "Upcoming Release" to "Release 0.6.x" when we do a release PR. Could you update this please? 😄

sbrugman: Done!

ankatiyar (approved):
Thank you @sbrugman for this feature! 💯

PR #338, opened 2023-09-14
ci: Update condition for creating a notification for nightly build failure

ankatiyar:
## Description

I noticed that the datasets tests were failing for a day or two but no issue was created. 
The if conditions on the success/failure of github actions jobs works weirdly. If one of the sub tasks fails the jobs for other python versions and os are cancelled so sometimes it skips the notification creation step. 

The status for the failed nightly build is shown as cancelled instead -> https://github.com/kedro-org/kedro-plugins/actions/workflows/nightly-build.yml

## Development notes

Tested on fork -

With `if: failure()` condition - https://github.com/ankatiyar/kedro-plugins/actions/runs/6184855180
With `if: ${{ !success() }}` condition - https://github.com/ankatiyar/kedro-plugins/actions/runs/6184875778
## Checklist

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -24,7 +24,7 @@ jobs:
       issues: write
     name: Notify failed build for airflow
     needs: airflow-test
-    if: failure()
+    if: ${{ !success() }}
     runs-on: ubuntu-latest
     steps:
       - uses: jayqi/failed-build-issue-action@v1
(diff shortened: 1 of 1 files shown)


merelcht (approved):
Thanks for keeping an eye on this and fixing it! 👍

AhdraMeraliQB (approved):
Just a curiosity so don't want to block the PR - what about `${{ failure() }}` ?

ankatiyar:
> Just a curiosity so don't want to block the PR - what about `${{ failure() }}` ?

So when you use the `if:` statement, there's no need for the `{{ }}` because the statement is already parsed as an expression except when a statement starts with `!`(not) since `!` is reserved notation in YAML format. It was already set to trigger the notification workflow incase of failure but when one of the tests fails, the entire suite of tests is cancelled and the status is set to cancelled and not failure and it hasn't been triggering the workflows.

AhdraMeraliQB:
@ankatiyar That makes sense, thanks!

Issue #319, opened 2023-08-29
Incorrect(?) warning when using SparkDataSet on Databricks with S3


alamastor:
## Description
Using a `SparkDataSet` on Databricks with a `filepath` that starts with `s3://` produces a warning; `Using SparkDataSet on Databricks without the `/dbfs/` prefix in the filepath is a known source of error.` The error comes from [here](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/spark/spark_dataset.py#L287). I don't think this is actually a problem.

## Context
The warning is not clear about why this is a problem, using `s3` doesn't seem to be a cause of error for my use case. It creates log messages which need to be filtered.

## Steps to Reproduce
1. Use a `SparkDataSet` on Databricks with a `filepath` that starts with `s3://`.

## Expected Result
The warning should be more clear about why this is a problem if it is actually a problem or should not be produced.

## Actual Result
Confusing warning produced.

## Your Environment
Occurs on current master when running on Databricks due to https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/spark/spark_dataset.py#L286C7-L286C7.

astrojuanlu:
Thanks for the heads up @alamastor. @noklam I see this comes from https://github.com/kedro-org/kedro-plugins/issues/117#issuecomment-1470891953, do you want to add some more context?

noklam:
I see what's the problem. The assumption here is we assume you are using `dbfs` on Databricks. Is there a reason why you are not using `dbfs` when you are using Databricks already?

I think in this case there is no problem and we shouldn't raise the warning. In databricks if you are not using `dbfs`, the data will be saved to the driver node, which creates weird bug( You may see the data on the notebook you created, but you get FileNotFound on other nodes). Under the hood, Databricks also have some [magic path conversion](https://docs.databricks.com/en/files/index.html#understand-default-locations-with-examples) (Spark only).

What we can do here is add an extra condition for the warning.
1. startswith `dbfs` (no warning as is)
2. startswith local path `/` or relative path (show warning as is)
3. startswith `s3` or equivalent (use `CLOUD_PROTOCOL`) (new change)

astrojuanlu:
Thanks for the explanation @noklam. @alamastor would you like to send a pull request for this one?

alamastor:
Thanks for the quick response!
> Is there a reason why you are not using dbfs when you are using Databricks already?
[...]

PR #341, opened 2023-09-18
fix(datasets): Don't warn for SparkDataset on Databricks when using s3

alamastor:
## Description
Fix for #319 

## Development notes
I've implemented the fix for `s3`, `s3a`, `s3n`. If there's protocols I should add for Azure, GCP or others, let me know - I don't have experience with these and couldn't find the protocols to use, if they exist.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -10,9 +10,7 @@
 ## Community contributions
 Many thanks to the following Kedroids for contributing PRs to this release:
 * [PtrBld](https://github.com/PtrBld)
-
-## Community contributions
-Many thanks to the following Kedroids for contributing PRs to this release:
+* [Alistair McKelvie](https://github.com/alamastor)
 * [Felix Wittmann](https://github.com/hfwittmann)
 
 # Release 1.7.1
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,6 +1,8 @@
 # Upcoming Release
 ## Major features and improvements
 ## Bug fixes and other changes
+* Fix erroneous warning when using an cloud protocol file path with SparkDataSet on Databricks.
+
 ## Upcoming deprecations for Kedro-Datasets 2.0.0
 
 # Release 1.7.1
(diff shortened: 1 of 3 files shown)


noklam:
@alamastor Thanks for the PR! I think we can reuse the existing [`CLOUD_PROTOCOLS`](https://github.com/kedro-org/kedro/blob/2f76e7fdf47b0b8a6840b9c9ad19d61897daf0da/kedro/io/core.py#L30).

The point here is that we should only raise error when it's a local path. Maybe there are more direct ways, otherwise we can keep your logic that if it's not DBFS and not in the known protocols, then we raise the error.

alamastor:
Thanks @noklam, that makes sense. If we want to check it's local path, I could just check that `://` is not in the path. Or I could reuse `CLOUD_PROTOCOLS`, in that case I think it would be nice to update `_split_filepath` to return the protocol name without `://`, so it works easily with `CLOUD_PROTOCOLS`, and update other uses of `fs_prefix`. What do you think?

noklam:
@alamastor I think it may break quite a few things, `_split_filepath` is rather special due to Spark & databricks. In `kedro`, there is a function that used extensively in most versioned datasets.

https://github.com/kedro-org/kedro/blob/3d41425c39ff061809f2febb87994d787f03bf4d/kedro/io/core.py#L668-L679

There are also few reasons we shouldn't do this:
- The protocol definition is from `urllib` (more or less standard), I don't want to introduce confusion here.
- The protocol need to be the same as `fsspec` (abstract of filesystem, the library that we used in Kedro to load file from remote storage). Internally we fork some function from fsspec to fix some bug historically. 

Do you think you can make it work without changing the `CLOUD_PROTOCOL` definition?

alamastor:
@noklam, I went with a simpler approach. Does that look ok?

noklam:
I am on PTO now, will have a look as soon as I back to work 🙏

noklam (approved):
Thank you for the PR, looks good :)

merelcht (approved):
Thanks for the contribution @alamastor 👍 ⭐

Issue #337, opened 2023-09-14
s3fs dependency is more than 3 years old. 


MatthiasRoels:
## Description

Not really a bug, but when I was running a kedro pipeline for a demo project, I noticed that I got a warning from `fsspec`: 

```
UserWarning: Your installed version of    warnings.py:109
                             s3fs is very old and known to cause                                                                                      
                             severe performance issues, see also https://github.com/dask/dask/issues/10276                                            
                                                                                                                                                      
                             To fix, you should specify a lower version bound on s3fs, or                                                             
                             update the current installation.                                   
``` 

## Steps to Reproduce

Run kedro-iris starter project with kedro-datasets installed 


## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): v0.18.13
* Kedro plugin and kedro plugin version used (`pip show kedro-airflow`): kedro-datasets v1.7.0
* Python version used (`python -V`): 3.10
* Operating system and version: Debian 
* fsspec: 2023.9.0
* s3fs: 0.4.2 (March 2020), current version: 2023.9.0

merelcht:
This was completed in https://github.com/kedro-org/kedro-plugins/pull/463

PR #348, opened 2023-09-26
build(datasets): upgrade s3fs to newer calver

MatthiasRoels:
## Description

s3fs version is already 3+ years old. The aim of this PR is to upgrade it to their new calver versioning approach. 

Fixes https://github.com/kedro-org/kedro-plugins/issues/337

## Development notes

When s3fs switched to calver versioning, they switched from using `botocore` to `aiobotocore`, which is an async client for AWS services. As a result it became much more difficult to use `moto` for unit testing as many people faced issues like the one described here https://github.com/aio-libs/aiobotocore/issues/755. To fix it, we had to switch to using moto in server mode.

## Checklist


MatthiasRoels:
@noklam: No I didn't encounter any weird authn errors. If you don't use moto in server mode, you always run into these kind of errors: 

```
AttributeError: 'AWSResponse' object has no attribute 'raw_headers'
``` 

For some reason, it occurs only in some test files. The tests for `PandasCSVDataset` run fine for example (although it's just one test involving S3). When I google the error, 2 fixes consistently appear; either you patch some boto function calls or you use moto in sever mode. The later looked like the most elegant one to fix the failing tests.

As a side note: I noticed there is a lot of inconsistency in the types of tests between different dataset classes. Is that intentional?

noklam:
@MatthiasRoels that's fine, I have research this in the past I think server mode is the right way.

kedro-datasets/tests/video/test_video_dataset.py
```diff
@@ -139,21 +122,26 @@ def test_load_missing_file(self, empty_dataset_mp4):
...
+        body_stream = mocked_s3_bucket.get_object(
+            Bucket=S3_BUCKET_NAME, Key=f"{S3_KEY_PATH}/{video_name}"
+        )["Body"].read()
         tmp_file = tmp_path / video_name
-        mocked_s3_bucket.download_file(
-            Bucket=S3_BUCKET_NAME,
-            Key=S3_KEY_PATH + "/" + video_name,
-            Filename=str(tmp_file),
-        )
+        with open(tmp_file, "wb") as file:
```
sbrugman, line 142: nit: we could do `tmp_file.open("wb")` since it's already a pathlib object

kedro-datasets/setup.py
```diff
@@ -6,7 +6,7 @@
 PANDAS = "pandas>=1.3, <3.0"
 SPARK = "pyspark>=2.2, <4.0"
 HDFS = "hdfs>=2.5.8, <3.0"
-S3FS = "s3fs>=0.3.0, <0.5"
+S3FS = "s3fs>=2021.04, <2024.1"  # Upper bound set arbitrarily, to be reassessed in early 2024
```
noklam, line 9: Can we keep the lower bound?

kedro-datasets/tests/conftest.py
```diff
@@ -32,3 +41,87 @@ def save_args(request):
...
+
+    return fs_args
+
+
+@fixture
+def credentials():
+    return {
+        "key": "fake_access_key",
+        "secret": "fake_secret_key",
+    }
```
noklam, line 64: Why there are 2 sets of credentials?

Do we need `key` and `secret` or the `AWS_XXXXX`?

kedro-datasets/setup.py
MatthiasRoels: Why do you want to keep the lower bound? That's already more than 4years old and incompatible with fsspec's calVer versioning, that we already set in kedro (e.g. for kedro `v0.18.0` https://github.com/kedro-org/kedro/blob/0.18.0/requirements.txt#L6)

kedro-datasets/tests/conftest.py
MatthiasRoels: One is for credentials file, the other one are env vars. The env vars are required for bucket setup etc while the credentials are used in the Dataset classes.

kedro-datasets/tests/video/test_video_dataset.py
MatthiasRoels: I didn't knew that was possible :-). Do you want me to change it?

kedro-datasets/setup.py
astrojuanlu: I agree it should be OK to bump the lower bound, the only thing here is that this PR would then have to wait for the upcoming kedro-datasets 2.0.0 :+1: cc @merelcht

MatthiasRoels:
Seems like the unit tests failed on (windows-latest, 3.8). Unfortunately, I have no access to a Windows machine for debugging and testing. So it would be nice if someone can help here. Thanks!

kedro-datasets/setup.py
noklam: @MatthiasRoels You are right, I thought the Kedro pin is at 0.3.0 but it is actually just for tests.
```suggestion
S3FS = "s3fs>=2021.4, <2024.1"  # Upper bound set arbitrarily, to be reassessed in early 2024
```

Not sure if this matter but the version name should be `2021.4.0`

merelcht:
@MatthiasRoels I've tried running this on Gitpod (ubuntu) and the tests also fail there, so it's not just a windows problem. You can also see that in the builds: https://github.com/kedro-org/kedro-plugins/actions/runs/6557080300/job/17809502926?pr=348 I guess it was just the windows 3.8 that failed first.

noklam:
Update branch to see if the problem persist

MatthiasRoels:
Thanks @merelcht for taking over, I wanted to work on this for a while, but I’m in the middle of a cloud migration so I’m way to busy 🙁

merelcht:
@MatthiasRoels I opened https://github.com/kedro-org/kedro-plugins/pull/463 where I've got most tests passing now and marked the two failing ones with `@pytest.mark.xfail`

Issue #224, opened 2023-06-02
Support Polars lazy evaluation


noklam:
## Description
Polars is efficient and support lazy evaluation mode which could be useful for memory hungry pipeline.

https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.scan_csv.html
Maybe supporting a flag with `lazy` to use `pl.scan_csv` ?

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

noklam:
I think it's different. Polars already support this out of the box, so there isn't much implementation overhead.  Something like this:
```python
if lazy:
  pl.scan_csv(xxx)
else:
  pl.read_csv(xxx)
```

astrojuanlu:
The third one looks the most promising to me! Many things in Polars are experimental and the project evolves fast, but I think it's worth trying.

noklam:
Why using fsspec contradict with lazy loading?

MatthiasRoels:
> Why using fsspec contradict with lazy loading?

It does not contradict it per se. It’s just that when `fsspec`, you use Python for I/O which means that you loose all the `polars` specific I/O goodies (read Rust implementation). Hence, this will negatively impact performance. 

What happens under the hood is that Python code (from `fsspec`) is responsible for reading data in memory at which moment in time Rust (`polars`) takes over for the actual processing. This means that you can’t benefit from things like predicate pushdown etc.

astrojuanlu:
IIUC, `pyarrow.dataset` https://arrow.apache.org/docs/python/dataset.html provides lazy/deferred reading capabilities in general, am I right?

Don't think Arrow would work for all datasets (think of weird stuff like Video, HDF5 and so on) but definitely worth giving it some thought for tabular datasets.

In any case, for a first iteration maybe we don't need to strive for consistency, and just focus on Polars lazy evaluation.

PR #350, opened 2023-09-27
feat(datasets): support Polars lazy evaluation 

MatthiasRoels:
## Description

For optimal data processing, it is recommended to use Polar's Lazy API (instead of the Eager one) so it is only natural that kedro-datasets supports it too.

Closes https://github.com/kedro-org/kedro-plugins/pull/224

## Checklist


noklam:
You are killing it with this rate of PR 🔥

MatthiasRoels:
And just as I created this PR, Polars releases a new version with native support to read parquet files from AWS/GCP/Azure... From what I can see, it should also work for the lazy API. 

So, this brings us to the next question: how do we proceed with this PR? 

1. Should we just finish it as-is (allowing versions < 0.19.4) and use arrow to read data AWS/GCP/Azure? I have been using this since beginning of summer and never ran into issue, so I know it works properly!
2. Bump polars version to 0.19.4 to allow native reading of parquet files from Cloud? Pro: easier to implement. Con: it only works for parquet for now...

astrojuanlu:
Maybe wait a couple of releases to see if it goes from experimental to mature? If we are going to throw away this code in a couple of months, I wouldn't say it makes a lot of sense to continue working on it.

They also said that the first release candidate of polars 1.0 is coming before the end of the year https://github.com/pola-rs/polars/issues/6616#issuecomment-1736901318 so maybe things will estabilize soon :)

MatthiasRoels:
I gave it a little thought and if anyone is willing to review, I would like to finish it, despite Polars’ native support for reading/writing to object store. The reason is simple; if this functionality stabilises on Polars’ side, we only have change (read, simplify) our load/save methods! 

The way I see it moving forward is that we need 2 datasets: one for eager loading and one for lazy loading. So I would keep the `GenericDataset` (potentially rename it to e.g. `EagerDataset`) and rename mine to `LazyDataset`. The reason I want to keep the two is that there are some file extensions (e.g. Excel) that don’t have a Lazy loading option. 

Curious to hear other opinions!

astrojuanlu:
There's two options: `LazyPolarsGenericDataset` (ugh, a bit long) or 

```
if lazy:
  pl.scan_csv(xxx)
else:
  pl.read_csv(xxx)
```

as @noklam suggested in https://github.com/kedro-org/kedro-plugins/issues/224#issuecomment-1573727946

Since the return type is going to be completely different, I'd rather have a different dataset indeed.

MatthiasRoels:
Is it ok if I:

1. Rename my dataset to `LazyDataset` (other suggestions welcome)
2. Rename `GenericDataset` to `EagerDataset` (other suggestions welcome)

To make it clear that one uses the eager API while the other one uses the Lazy API

Shouldn't we then also consider removing (or mark as deprecated) the `CSVDataset`? And as a follow-up: should we do the same with the pandas datasets? Basically, the `pandas.GenericDataset` replaces the CSV, Excel, Feather, JSON, Parquet and XML ones, which should make the code base leaner. This is beyond the scope of this PR obviously...

astrojuanlu:
> Shouldn't we then also consider removing (or mark as deprecated) the CSVDataset? And as a follow-up: should we do the same with the pandas datasets? Basically, the pandas.GenericDataset replaces the CSV, Excel, Feather, JSON, Parquet and XML ones, which should make the code base leaner. This is beyond the scope of this PR obviously...

This is a discussion worth having but it has some implications in the docs, starters etc, could you open a separate issue about it?

kedro-datasets/kedro_datasets/polars/__init__.py
```diff
@@ -10,11 +10,14 @@
 CSVDataset: Any
 GenericDataSet: type[GenericDataset]
 GenericDataset: Any
+PolarsDataSet: type[PolarsDataset]
+PolarsDataset: Any
```
merelcht, line 14: If this is a new dataset there's no reason to use both the old and new spelling. It should just use the new `Dataset` ending.

kedro-datasets/kedro_datasets/polars/polars_dataset.py
```diff
@@ -0,0 +1,272 @@
...
+                f"Saving {self.__class__.__name__} to a directory is not supported."
+            )
+
+        if "partition_cols" in self._save_args:
+            raise DatasetError(
+                f"{self.__class__.__name__} does not support save argument "
+                f"'partition_cols'. Please use 'kedro.io.PartitionedDataset' instead."
+            )
+
+        BytesIO()
```
merelcht, line 217: What does this do?

kedro-datasets/kedro_datasets/polars/polars_dataset.py
```diff
@@ -0,0 +1,272 @@
...
+    if name in _DEPRECATED_CLASSES:
+        alias = _DEPRECATED_CLASSES[name]
+        warnings.warn(
+            f"{repr(name)} has been renamed to {repr(alias.__name__)}, "
+            f"and the alias will be removed in Kedro-Datasets 2.0.0",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return alias
+    raise AttributeError(f"module {repr(__name__)} has no attribute {repr(name)}")
```
merelcht, line 272: Double checking again that `PolarsDataSet` doesn't actually exist right? These deprecation warning were added to the older datasets that were originally named with the capital "s" in `DataSet`. Any new datasets should just use the `Dataset` ending and there's no need to add this deprecation warning.

kedro-datasets/setup.py
```diff
@@ -59,6 +59,10 @@ def _collect_requirements(requires):
     [
         POLARS, "pyarrow>=4.0", "xlsx2csv>=0.8.0", "deltalake >= 0.6.2"
     ],
+    "polars.PolarsDataSet":
```
merelcht, line 62:
```suggestion
    "polars.PolarsDataset":
```

kedro-datasets/tests/polars/test_polars_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+        fs_args=fs_args,
+    )
+
+
+@pytest.fixture
+def dummy_dataframe():
+    return pl.DataFrame({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
+
+
+class TestPolarsDataSet:
```
merelcht, line 51:
```suggestion
class TestPolarsDataset:
```

kedro-datasets/tests/polars/test_polars_dataset.py
```diff
@@ -0,0 +1,159 @@
...
+    """Test class for PolarsDataSet tests"""
```
merelcht, line 52:
```suggestion
    """Test class for PolarsDataset tests"""
```

kedro-datasets/kedro_datasets/polars/polars_dataset.py
```diff
@@ -0,0 +1,272 @@
+"""``PolarsDataset`` loads/saves data from/to a data file using an underlying
+filesystem (e.g.: local, S3, GCS). It uses polars to handle the
+type of read/write target.
```
merelcht, line 3: I think this doc string should mention this dataset uses lazy evaluation.

merelcht (commented):
Thanks a lot for this contribution @MatthiasRoels! I've left some comments, the main being around the naming syntax (classes ending in `DataSet` are old, any new ones should end in `Dataset`). 

In terms of naming the datasets, how about: 
- `LazyPolarsDataset`
- `EagerPolarsDataset`

kedro-datasets/kedro_datasets/polars/polars_dataset.py
MatthiasRoels: Nothing, that's probably a left-over of something (or a copy-paste error)

MatthiasRoels:
> Thanks a lot for this contribution @MatthiasRoels! I've left some comments, the main being around the naming syntax (classes ending in `DataSet` are old, any new ones should end in `Dataset`).
> 
> In terms of naming the datasets, how about:
> 
> * `LazyPolarsDataset`
> * `EagerPolarsDataset`

@merelcht That a nice suggestion! But that would introduce a breaking change (renaming the GenericDataset to EagerPolarsDataset). Should I already rename it while keeping the old names with deprecation warnings (like we did for DataSet vs Dataset)?

kedro-datasets/setup.py
MatthiasRoels: Fixed!

kedro-datasets/tests/polars/test_polars_dataset.py
MatthiasRoels: Fixed!

merelcht:
> @merelcht That a nice suggestion! But that would introduce a breaking change (renaming the GenericDataset to EagerPolarsDataset). Should I already rename it while keeping the old names with deprecation warnings (like we did for DataSet vs Dataset)?

Yes that sounds good! And then we can remove the alias in the next breaking kedro-datasets release `2.0.0` 🙂

kedro-datasets/setup.py
```diff
@@ -59,6 +59,10 @@ def _collect_requirements(requires):
     [
         POLARS, "pyarrow>=4.0", "xlsx2csv>=0.8.0", "deltalake >= 0.6.2"
     ],
+    "polars.LazyPolarsDataset":
```
merelcht, line 62: Should `EagerPolarsDataset` be added here as well? If someone uses that directly I'm guessing the requirements would now not be picked up?

merelcht (commented):
I haven't tried running this, but the code looks good! Can you also add this change to the release note + mention that `polars.GenericDataSet` will be deprecated and be replaced by `polars.EagerPolarsDataset`?

MatthiasRoels:
Read the Docs build failed because of a dependency conflict with Dask in test vs docs. Weird that this does not occur in other PR's (or is it?)...

MatthiasRoels:
> > Read the Docs build failed because of a dependency conflict with Dask in test vs docs. Weird that this does not occur in other PR's (or is it?)...
> 
> We are aware of this issue, it has nothing to do with your PR!

Is there an open issue already?

astrojuanlu:
It got fixed already in https://github.com/kedro-org/kedro-plugins/pull/396 :sparkles

astrojuanlu:
This is using https://github.com/astrojuanlu/workshop-jupyter-kedro/tree/workshop-steps/data by the way

MatthiasRoels:
@astrojuanlu: You were correct that the load args were not properly passed when loading files from object stores. However, this will not fix the issue you had because for object stores, we leverage Arrow to load the dataset. This means that you actually have to pass an Arrow schema (instead of polars `dtypes`). Unfortunately, you have to pass in the full schema, not just the types you want to manually specify... This is downside from the fact that older Polars version could not read directly from Object stores (which, in the newer versions, is in beta).

astrojuanlu:
I see, thanks! it's a bit annoying to specify the full schema for remote filepaths, do you think this is enough reason to embrace the new method, even if it's beta?

MatthiasRoels:
@astrojuanlu: I think there are pro's and cons of both approaches 

Current approach 
Pros:

- It is based on PyArrow which is well-established
- I have been using it for 3 months with 100+ daily kedro runs on different datasets daily (well tested?)

Cons:

- you cannot use convenient Polars methods and you have to in the Arrow docs if you want to do something special like overwriting the schema
- `load_args` look different when you switch from a file stored to locally to one stored in an object store
- If we change later on, this change could be considered breaking (everyone will potentially need to change `load_args`).

Approach when we force a newer version of Polars
Pros: 

- Uniform syntax 
- No further (breaking?) changes required in the future

Cons: 

- We have to force users to use a more recent version of Polars (not sure if that's necessarily a bad thing?)
- reading from cloud storage using polars is experimental so we might encounter unexpected behaviour (I have also never tested it so I have no idea what to expect).

Given the pros/cons of each, we need to decide how we proceed. I think it comes down to stability vs convenience, no?

astrojuanlu:
Good assessment, thanks for the writeup. I keep coming back and forth about this, because I'm pretty confident the new approach will stabilise, but to be honest there's no 100 % guarantee.

Let's proceed with what you have created here, and we can revisit when Polars 1.0 is out 👍🏽

merelcht (approved):
Tiny comment, but otherwise LGTM! Thank you so much for your contribution @MatthiasRoels it's truly awesome ⭐ 😄

astrojuanlu:
Congrats on your first contribution @MatthiasRoels ! 🎉

PR #352, opened 2023-09-28
Add support for `databricks-connect>=13.0`

### LINKED ISSUE #351: Support for remote spark sessions and databricks-connect
## Description
Since Spark 3.4, [spark-connect](https://spark.apache.org/docs/latest/spark-connect-overview.html) (and the equivalent [databricks-connect v2](https://docs.databricks.com/en/dev-tools/databricks-connect-ref.html)) were introduced for seamless development with remote Spark sessions. 

This is extremely useful for interactive debugging of kedro pipelines from an IDE

However, the syntax for creating remote Spark Sessions was changed, and current kedro spark implementations have hard-coded the legacy SparkSession creation: `SparkSession.builder.getOrCreate()`.

Therefore kedro is currently not supporting remote sessions.

I don't know about on premise spark setups, but with databricks-connect>=13.1 the current kedro spark datasets can't be instantiated on a remote session. When trying to instantiate a kedro spark dataset with databricks-connect>=13.1 the following error is raised:

```
RuntimeError: Only remote Spark sessions using Databricks Connect are supported. Could not find connection parameters to start a Spark remote session.
```

## Context

I suggest extending all spark-related code in kedro to support remote sessions

## Possible Implementation

I work with a Databricks environment and this code works fine to instantiate a `spark` object either in a remote databricks-connect>=13.0 session or a Databricks notebook directly from the web UI:

```python
import pyspark
from pyspark.sql import SparkSession


def get_spark() -> Any:
    """
    Retuns the SparkSession, we need this wrapper because the SparkSession
    is retrieved differently in databricks-connect vs a Notebook in web UI
    """
    # When in databriks-connect pyspark version is equal to DBR version
    # instead of the actual pyspark version
    pyspark_major_version = int(pyspark.__version__.split(".")[0])
    if pyspark_major_version >= 13:
        # In this case we are in a databricks-connect >= 13.0.0 (a.k.a databricks-connect-v2)
        # remote session, and therefore spark is initialized differently
        from databricks.connect import DatabricksSession

        spark = DatabricksSession.builder.getOrCreate()
    else:
        # For sessions in Notebook web UI or previous versions on databricks-connect
        # we get spark normally
        spark = SparkSession.builder.getOrCreate()

    return spark
```

This could probably be extended or improved to support all different environments (remote Session to on premise Spark cluster, databricks-connect<=12.2, etc.)

A quick fix would be to update kedro's [get_spark()](https://github.com/kedro-org/kedro/blob/main/kedro/extras/datasets/spark/spark_dataset.py#L383) function to this one (or a more flexible and better implementation), and maybe consider moving it to an independent module instead of each dataset having its own implementation of `get_spark()` 

## Possible Alternatives

I would love to hear kedro developers opinions on this and design a robust solution that fully supports remote spark-connect and databricks-connect sessions

MigQ2:
## Description
Closes #351 

## Checklist

- [ ] Added tests to cover my changes (I don't think there's an available Databricks environment to test on)

### DIFF
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset.py
@@ -24,6 +24,28 @@ from kedro_datasets._io import AbstractVersionedDataset, DatasetError
 
 logger = logging.getLogger(__name__)
 
+def _get_spark() -> Any:
+    """
+    Returns the SparkSession, we need this wrapper because the SparkSession
+    is retrieved differently in databricks-connect vs a Notebook in web UI
+    """
+    # When in databricks-connect pyspark version is equal to DBR version
+    # instead of the actual pyspark version
+    pyspark_major_version = int(pyspark.__version__.split(".")[0])
+    if pyspark_major_version >= 13:
+        # In this case we are in a databricks-connect >= 13.0.0 (a.k.a databricks-connect-v2)
+        # remote session, and therefore spark is initialized differently
+
+        from databricks.connect import DatabricksSession
+
+        spark = DatabricksSession.builder.getOrCreate()
+    else:
+        # For sessions in Notebook web UI or previous versions on databricks-connect
+        # we get spark normally
+        spark = SparkSession.builder.getOrCreate()
+
+    return spark
+
 
 def _parse_glob_pattern(pattern: str) -> str:
     special = ("*", "?", "[")
@@ -314,7 +336,7 @@ class SparkDataset(AbstractVersionedDataset[DataFrame, DataFrame]):
         elif filepath.startswith("/dbfs/"):
             # dbfs add prefix to Spark path by default
             # See https://github.com/kedro-org/kedro-plugins/issues/117
-            dbutils = _get_dbutils(self._get_spark())
+            dbutils = _get_dbutils(_get_spark())
             if dbutils:
                 glob_function = partial(_dbfs_glob, dbutils=dbutils)
                 exists_function = partial(_dbfs_exists, dbutils=dbutils)
@@ -382,13 +404,11 @@ class SparkDataset(AbstractVersionedDataset[DataFrame, DataFrame]):
             "version": self._version,
         }
 
-    @staticmethod
-    def _get_spark():
-        return SparkSession.builder.getOrCreate()
+
 
     def _load(self) -> DataFrame:
         load_path = _strip_dbfs_prefix(self._fs_prefix + str(self._get_load_path()))
-        read_obj = self._get_spark().read
+        read_obj = _get_spark().read
 
         # Pass schema if defined
         if self._schema:
@@ -404,7 +424,7 @@ class SparkDataset(AbstractVersionedDataset[DataFrame, DataFrame]):
         load_path = _strip_dbfs_prefix(self._fs_prefix + str(self._get_load_path()))
 
         try:
-            self._get_spark().read.load(load_path, self._file_format)
+            _get_spark().read.load(load_path, self._file_format)
         except AnalysisException as exception:
             # `AnalysisException.desc` is deprecated with pyspark >= 3.4
             message = exception.desc if hasattr(exception, "desc") else str(exception)
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,5 +1,7 @@
 # Upcoming Release
 ## Major features and improvements
+* Spark and Databricks based datasets now support [databricks-connect>=13.0](https://docs.databricks.com/en/dev-tools/databricks-connect-ref.html)
+
 ## Bug fixes and other changes
 ## Community contributions
 
--- a/kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
+++ b/kedro-datasets/kedro_datasets/databricks/managed_table_dataset.py
@@ -14,6 +14,7 @@ from pyspark.sql.types import StructType
 from pyspark.sql.utils import AnalysisException, ParseException
 
 from kedro_datasets._io import AbstractVersionedDataset, DatasetError
+from kedro_datasets.spark.spark_dataset import _get_spark
 
 logger = logging.getLogger(__name__)
 
@@ -262,10 +263,6 @@ class ManagedTableDataset(AbstractVersionedDataset):
             exists_function=self._exists,
         )
 
-    @staticmethod
-    def _get_spark() -> SparkSession:
-        return SparkSession.builder.getOrCreate()
-
     def _load(self) -> Union[DataFrame, pd.DataFrame]:
         """Loads the version of data in the format defined in the init
         (spark|pandas dataframe)
@@ -281,7 +278,7 @@ class ManagedTableDataset(AbstractVersionedDataset):
         if self._version and self._version.load >= 0:
             try:
                 data = (
-                    self._get_spark()
+                    _get_spark()
                     .read.format("delta")
                     .option("versionAsOf", self._version.load)
                     .table(self._table.full_table_location())
@@ -289,7 +286,7 @@ class ManagedTableDataset(AbstractVersionedDataset):
             except Exception as exc:
                 raise VersionNotFoundError(self._version.load) from exc
         else:
-            data = self._get_spark().table(self._table.full_table_location())
+            data = _get_spark().table(self._table.full_table_location())
         if self._table.dataframe_type == "pandas":
             data = data.toPandas()
         return data
@@ -327,7 +324,7 @@ class ManagedTableDataset(AbstractVersionedDataset):
             update_data (DataFrame): the Spark dataframe to upsert
         """
         if self._exists():
-            base_data = self._get_spark().table(self._table.full_table_location())
+            base_data = _get_spark().table(self._table.full_table_location())
             base_columns = base_data.columns
             update_columns = update_data.columns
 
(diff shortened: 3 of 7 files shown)


kedro-datasets/kedro_datasets/spark/spark_dataset.py
```diff
@@ -26,6 +27,27 @@
...
+        # the remote session is instantiated using the databricks module 
+        # If the databricks-connect module is installed, we use a remote session 
+        from databricks.connect import DatabricksSession
+
+        spark = DatabricksSession.builder.getOrCreate()
+
+    except ImportError:
+        # For "normal" spak sessions that don't use databricks-connect
+        # we get spark normally
+        spark = SparkSession.builder.getOrCreate()
```
cdkrot, line 46: For even further improved compatibility, add
```
from pyspark.sql.connect.session import SparkSession as RemoteSparkSession
```

and replace this line with `RemoteSparkSession.builder.getOrCreate()`.

MigQ2: Thanks a lot for looking into this @cdkrot , super appreciated

Do you mean another try except? So the fallback waterfall would be:

1. DatabricksSession
2. RemoteSparkSession
3. SparkSession

I guess adding RemoteSparksession would helo adding support for non-Databricks Spark-connect sessions?

cdkrot: Sorry, my bad, the logic in code is good as is 👍. sql.SparkSession will figure out if it needs to create sql.SparkSession or connect.SparkSession.

Possibly it's best to rewrite the comments though. Essentially there is no reason to fallback from dbconnect session creation to pyspark's since dbconnect does it automatically, the only reason to fallback is if dbconnect is not installed.

kedro-datasets/kedro_datasets/spark/spark_dataset.py
```diff
@@ -26,6 +27,27 @@
 logger = logging.getLogger(__name__)
 
 
+def _get_spark() -> Any:
+    """
+    Returns the SparkSession, we need this wrapper because the SparkSession
+    is retrieved differently in databricks-connect vs a Notebook in web UI
```
cdkrot, line 33:
```suggestion
    Returns the SparkSession. In case databricks-connect is available we use it for
    extended configuration mechanisms and notebook compatibility,
    otherwise we use classic pyspark.
```

MigQ2:
@astrojuanlu could you please re-run the CI/CD checks to see if the current implementation passes all checks and is ready to merge?

astrojuanlu:
Thanks for the ping @MigQ2 ! Hope everything is good to go now. Feel free to ping me again if needed.

astrojuanlu:
And also thanks @cdkrot for reviewing, much appreciated 🙌🏽

MigQ2:
Hi @astrojuanlu , @noklam I think I need some help with figuring out the best way to keep spark tests working fine here. 

I see there's some wrangling with `mocker.patch` and `_get_spark` so the way I built it breaks the tests as each dataset wouldn't have its own `_get_spark()` method. 

I have thought about the following workaround but I'd like to know what you think before implementing it: 

We could move the logic of `_get_spark()` to a `utils.py` (or keep it in `spark_dataset.py`, if you prefer that), and keep a `self._get_spark()` staticmethod in all DataSet classes which would be just a wrapper for `utils._get_spark()`. 

This way we could keep tests as they are because each DataSet class would have its own `self._get_spark()` method and I think `mocker.patch` would work fine as it is withhout changing the test

WDYT?

noklam:
> I see there's some wrangling with mocker.patch and _get_spark so the way I built it breaks the tests as each dataset wouldn't have its own _get_spark() method.

In this case we could update the test to mock `get_spark` instead of `mocker.object`

> We could move the logic of _get_spark() to a utils.py (or keep it in spark_dataset.py, if you prefer that), and keep a self._get_spark() staticmethod in all DataSet classes which would be just a wrapper for utils._get_spark().

I think it's fine to keep it in `spark_dataset.py`, if we ever move it I may prefer it stay in the `spark/__init__.py` instead  of another new utils.py.

merelcht:
Hi @MigQ2 , are you able to address the review comments soon? We'd like to do a `kedro-datasets` release this week and it would be great to include your contribution if possible. Otherwise, it will be part of our next release.

MigQ2:
Hi @merelcht , I'd love to see it on the next release so I'll work on it ASAP. 

Is there an easy way I can run the test suite and CI/CD locally to verify everything is working fine without pinging someone from the kedro team to manually trigger it im GitHub?

astrojuanlu:
@MigQ2 Can you try `make test plugin=kedro-datasets` or even `cd kedro-datasets && pytest -k databricks`?

MigQ2:
Thanks @astrojuanlu , I managed to run tests locally and they should succeed now. Can you pls trigger CI/CD again to see if everything is alright?

I struggled a bit to get stuff working, so here go some ideas that maybe could help some wannabe-kedroids to contribute more easily:

1. Document how to run tests (basically document the `cd kedro-datasets && pytest -k databricks` you just mentioned before), I found it by myself for the kedro core, but I didn't find it so obvious for kedro-plugins

2. Help to setup the local environment to run the tests. I had to install all dependencies using `pip install ./kedro-datasets/[test]`, which took a while because of the big requirements list (especially pyspark and tensorflow).
And then I realized I had to install java8 ([which is a bit hard because it's been removed from package manager](https://stackoverflow.com/questions/57031649/how-to-install-openjdk-8-jdk-on-debian-10-buster)) for spark tests to run locally. 
I like using [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers) in vscode to have a consistent environment. Do you think it would be a good idea to provide a devcontainer with all dependencies (pypi, java, other system libraries, etc.) installed out of the box so that people can just open their repo in that container and run tests easily without worrying about installing all stuff in their system? I have a Dockerfile that worked for me, if you think it's a good idea we can talk and I can share

noklam:
@MigQ2 We have Gitpod which is similar to codes pace and you can have it on browser, local Pycharm or VScode.

MigQ2:
There seemed to be a dask conflict in library requirements. I guess that's not coming from my changes. 

LMK when that's fixed and I'll update the branch and let's re-trigger the CI/CD

merelcht:
> There seemed to be a dask conflict in library requirements. I guess that's not coming from my changes.
> 
> LMK when that's fixed and I'll update the branch and let's re-trigger the CI/CD

That issue has been resolved! Now the test builds are failing because of missing coverage: 
`kedro_datasets/spark/spark_dataset.py                 186      1    99%   46`

DimedS:
Thanks for the PR, @MigQ2. Can you clarify a few points for me? If I'm understanding this correctly, this PR enables remote connections to Databricks, prioritising the Databricks connector in all scenarios. However, when it comes to remote Spark session connections, is there a different syntax? For instance, would this work for a remote connection to an AWS EMR Spark cluster?

MigQ2:
> Thanks for the PR, @MigQ2. Can you clarify a few points for me? If I'm understanding this correctly, this PR enables remote connections to Databricks, prioritising the Databricks connector in all scenarios. However, when it comes to remote Spark session connections, is there a different syntax? For instance, would this work for a remote connection to an AWS EMR Spark cluster?

Hi @DimedS,

I haven't used standard spark-connnect sessions, and I don't have an AWS EMR or similar setup to test on. 
As per [spark-connect documentation](https://spark.apache.org/docs/latest/spark-connect-overview.html) I think syntax would be different, and something like the following code should be used:

```python
from pyspark.sql import SparkSession
spark = SparkSession.builder.remote("some-host").getOrCreate()
```

This PR only attempts to fix for Databricks setups, cause that's the one we use and the one we can test on, but I think it would be great to add support for non-databricks remote spark-connect sessions

MigQ2:
@merelcht I added `# pragma: no cover` to the databricks-connect session and I don't think there's a Databricks environment available to test on.

When the release is out I'll start using it with Databricks in our environment and come back here if I find anything that needs to be fixed os that we can improve

MigQ2:
As a side comment, I tried out gitpod to setup the environment but there were a couple things still weren't clear for me: 

1. When running `make plgun=kedro-datasets install-test-requirements` I still got a `dask` dependency error
2. I still don't know what's the recommended way to install java8 in the gitpod environment so that pyspark test can run properly

MigQ2:
> > > Thanks for the PR, @MigQ2. Can you clarify a few points for me? If I'm understanding this correctly, this PR enables remote connections to Databricks, prioritising the Databricks connector in all scenarios. However, when it comes to remote Spark session connections, is there a different syntax? For instance, would this work for a remote connection to an AWS EMR Spark cluster?
> > 
> > 
> > Hi @DimedS,
> > I haven't used standard spark-connnect sessions, and I don't have an AWS EMR or similar setup to test on. As per [spark-connect documentation](https://spark.apache.org/docs/latest/spark-connect-overview.html) I think syntax would be different, and something like the following code should be used:
> > ```
> > from pyspark.sql import SparkSession
> > spark = SparkSession.builder.remote("some-host").getOrCreate()
> > ```
> > 
> > 
> > This PR only attempts to fix for Databricks setups, cause that's the one we use and the one we can test on, but I think it would be great to add support for non-databricks remote spark-connect sessions
> 
> yes, looks like this, that we need to add additional issue for making possible also remote spark connects. It looks unclear that our default choice will be trying to connect to Databricks in any case, but I think for now it's ok to fastly allow Databricks remote connection. Maybe later we should rewrite it with more general solution.

Yes, agree. I think the try-except is not that bad cause I don't think anyone would install the databricks-connect package in a non-databricks setup.

Another point that I think would be nice is to think about remote sessions authentication. For now this PR doesn't have any authentication options (or any other spark options), so it can only authenticate with environment variables. 

However there are several other different ways to authenticate to databricks (and probably to spark-connect) that maybe other users would like to use and this implementation wouldn't support

noklam:
@MigQ2 can you open a separate issue for GitPod? I may have time to look at that, pretty sure this was working before we migrate to `kedro-plugins` so I can find it in older commits.

DimedS (approved):
@MigQ2, thank you for your contribution. It enables the use of remote Spark sessions with Databricks. However, we should keep the initial issue open, as this PR doesn't facilitate remote Spark sessions with solutions other than Databricks.

noklam (approved):

Issue #331, opened 2023-09-05
Lint all plugins on python 3.11


ankatiyar:
## Description
Follow up to #265 

Linters on `kedro-airflow`, `kedro-docker` and `kedro-telemetry` run on python 3.8 for now.

Trying to bump versions on ci causes this issue with `pre-commit` - 
```
Trim Trailing Whitespace.................................................An unexpected error has occurred: TypeError: Random.shuffle() got an unexpected keyword argument 'random'
```
Check if #320 solves the issue 

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

Issue #320, opened 2023-08-29
Add `ruff` to `kedro-plugins`


merelcht:
## Description
Add ruff to `kedro-plugins` just like we did on the core Kedro repo.

## Context
https://github.com/kedro-org/kedro-plugins/issues/317

## Possible Alternatives
Fix version issues with `pylint` for `kedro-airflow` and `kedro-docker` as described in https://github.com/kedro-org/kedro-plugins/issues/317. As discussed in backlog grooming we prefer to introduce `ruff`.

PR #354, opened 2023-09-28
style: Introduce `ruff` for linting in all plugins.

merelcht:
## Description
Addresses https://github.com/kedro-org/kedro-plugins/issues/320 and https://github.com/kedro-org/kedro-plugins/issues/331

## Development notes
Replaced `flake8`, `isort` and `pylint` setup with `ruff`.

## Checklist


noklam (approved):
⭐️ Heroic work for removing all the pylint comments

.pre-commit-config.yaml
```diff
@@ -14,178 +14,44 @@ repos:
...
-          - "--select=B,C,E,F,W,T4,B9"
-          - "--ignore=E203,E266,E501,W503"
-        exclude: "^kedro_airflow/dag_template.py|^template.py"
 
   - repo: local
     hooks:
-      # pylint quick checks
-      - id: pylint-quick-kedro-datasets
-        name: "Quick PyLint on kedro_datasets/*"
+      - id: ruff-kedro-datasets
```
noklam, line 20: https://github.com/kedro-org/kedro/blob/af2ba0ed7abaf004da45a443a0b3d37bf11dd30f/.pre-commit-config.yaml#L6C1-L14C69

    - repo: https://github.com/astral-sh/ruff-pre-commit
      # Ruff version.
      rev: v0.0.277


Can we use the ruff pre-commit hook? this way we don't have to install `ruff` locally and ensure CI always work the same as local.

merelcht: I don't know if it's possible to run it only for the plugin that has been changed with the pre-commit hook. @SajidAlamQB do you know if that could work? I know you put a lot of thought into this setup.

SajidAlamQB: Yes @merelcht is correct, the main issue is that we can't dynamically run hooks based on the specific plugin environment on the CI. The setup will trigger all the defined ruff linting hooks on all plugins, regardless of which plugin has changes.

A workaround would be to create a script that checks for changed plugins, and then conditionally executes ruff on those plugins.

ankatiyar (approved):
💯

Issue #321, opened 2023-08-30
kedro-airflow crashes with kedro<=0.18.3


luizvbo:
## Description

I was trying to use kedro-airflow with kedro==0.18.2 and was getting the error `AttributeError: 'KedroContext' object has no attribute 'config_loader'`.

## Context

I tried different versions of kedro and the bug was solved by installing kedro==0.18.4.

I noticed "kedro>=0.17.5" in the requirements list for kedro-airflow. The requirement could be bumped or the bug could be fixed.

## Steps to Reproduce
```
pip install kedro==0.18.3 kedro-airflow
kedro new --starter=spaceflights # Use default values
cd spaceflights
# Remove kedro-viz from src/requirements.txt (to avoid conflicts)
pip install -r src/requirements.txt
kedro airflow create
```

## Actual Result

```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /tmp/venv/bin/kedro:8 in <module>                                                                │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/kedro/framework/cli/cli.py:211 in main                     │
│                                                                                                  │
│   208 │   """                                                                                    │
│   209 │   _init_plugins()                                                                        │
│   210 │   cli_collection = KedroCLI(project_path=Path.cwd())                                     │
│ ❱ 211 │   cli_collection()                                                                       │
│   212                                                                                            │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/click/core.py:1157 in __call__                             │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/kedro/framework/cli/cli.py:139 in main                     │
│                                                                                                  │
│   136 │   │   )                                                                                  │
│   137 │   │                                                                                      │
│   138 │   │   try:                                                                               │
│ ❱ 139 │   │   │   super().main(                                                                  │
│   140 │   │   │   │   args=args,                                                                 │
│   141 │   │   │   │   prog_name=prog_name,                                                       │
│   142 │   │   │   │   complete_var=complete_var,                                                 │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/click/core.py:1078 in main                                 │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/click/core.py:1688 in invoke                               │
│                                                                                                  │
│ /tmp/venv/lib/python3.8/site-packages/click/core.py:1688 in invoke                               │
[...]

PR #381, opened 2023-10-09
fix: backwards compatibility for `kedro-airflow`

sbrugman:
## Description
Closes https://github.com/kedro-org/kedro-plugins/issues/321
Classified as `fix` as the `pyproject.toml` specifies that the plugin is compatible with Kedro 0.17.5, which without this fix it is not.

## Development notes
Tested with internal setup.

## Checklist


ankatiyar:
@sbrugman Thank you for the quick PR! I think this fix looks good but it also needs a test to satisfy 100% code coverage. Could you add that too?

sbrugman:
> @sbrugman Thank you for the quick PR! I think this fix looks good but it also needs a test to satisfy 100% code coverage. Could you add that too? + I think the commits need to be signed off

Added a test (now the coverage is 100% again) and signed off the commits

merelcht (approved):
Thanks for this fix @sbrugman ! 👍

ankatiyar (approved):
Thank you @sbrugman!

PR #392, opened 2023-10-16
chore: Drop Python 3.7 support for kedro-plugins

lrcouto:
## Description

Part of https://github.com/kedro-org/kedro/issues/2158

## Checklist


merelcht (commented):
The changes to `kedro-datasets` will need to wait until `1.8.0` is released, because dropping python 3.7 support is a breaking change and will have to go into `2.0.0`. So you can either wait with merging this or do the `kedro-datasets` changes in a separate PR. 

I also had a question around `kedro-datasets/setup.py`. I noticed you added a lower bound for `python_version>=3.8` to some requirements, why is that? This should be needed.

merelcht (approved):
LGTM, thanks @lrcouto ! I would also add the removal of `3.7` support to all release notes under the next breaking release so:
- `0.7.0` for `kedro-airflow`
- `0.4.0` for `kedro-docker`
- `0.3.0` for `kedro-telemetry`

PR #400, opened 2023-10-20
style: Add shared CSS and meganav to datasets docs

stichbury:
## Description
Adding CSS to datasets docs

Preview is here: https://kedro--400.org.readthedocs.build/projects/kedro-datasets/en/400/ and doesn't look quite right 😆 
Please could you take a look @vladimir-mck or @tynandebold ?

## Development notes
Updates to datasets docs files

## Checklist


vladimir-mck:
I see, we need to reduce the font size for the logo/title from 4 to 3rem for it to fit. We can either add a class to it and and reduce the font size for that particular class or we can reduce the font size by 1 for all the docs and keep it consistent. 

@tynandebold  wdyt ?

tynandebold:
Tough one. It's because `kedro-datasets` is a very long string and is giving us a headache.

For it to fit properly and also accompany the words `stable` or even `latest`, the font size will have to be smaller than `3em`:

image

I don't have a good solution for this. I suppose we can have it this size for datasets and I don't think we should change the size of the other titles.

tynandebold (approved):
Styles are looking good and consistent with the other two projects. Heap data is coming through fine, too.

kedro-datasets/docs/source/conf.py
```diff
@@ -25,7 +25,7 @@
 
 # -- Project information -----------------------------------------------------
 
-project = "kedro-datasets"
+project = "kedro"
```
astrojuanlu, line 28: Wondering if this should still be `kedro-datasets` (sorry I didn't catch that early on)

stichbury: Oops! Well, next time. At least it's now merged ! And https://docs.kedro.org/projects/kedro-datasets/en/latest/ is looking splendid.

PR #416, opened 2023-10-24
test(datasets): run doctests to check examples run

deepyaman:
## Description

Set up framework to address https://github.com/kedro-org/kedro/issues/2287. Will work with @Gundalai-Batkhuu to resolve skipped doctests in future PRs.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,11 @@ test:
 	cd $(plugin) && pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile
 
 # Run test_tensorflow_model_dataset separately, because these tests are flaky when run as part of the full test-suite
-dataset-tests:
+dataset-tests: dataset-doctests
 	cd kedro-datasets && pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile --ignore tests/tensorflow --ignore tests/databricks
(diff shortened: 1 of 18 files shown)


noklam:
separate `make datasets-doctests`, I think this is useful for purely documentation change sometimes.

merelcht:
Should this doctest be part of the `unit-tests` builds or would it be better to create separate builds for this?

deepyaman:
> Should this doctest be part of the `unit-tests` builds or would it be better to create separate builds for this?

@merelcht What would be the benefit of doing it in a separate build? The trigger isn't any different. Unit tests have to be run if source code is modified, as do doctests (unless we do something fancy to check that only the docstring is changed).

If anything, I could see it all happening as part of one `pytest` command--which is more the default behavior, where `--doctest-modules` flag makes it run doctests in addition to the rest of the suite.

merelcht (approved):

noklam:
> I was wondering if the results might be easier to read if they are separate. But I don't have a huge preference.

I thought about this too, but I don't think it is a good idea in this case because:
1. It need to re-installed the requirements and it's slow
2. GitHub Action has a concurrent job limit, we are hitting it already so it will further slow down the time to merge a PR.

Issue #404, opened 2023-10-23
Document and apply Python version support policy for all plugins


astrojuanlu:
## Description
In https://github.com/kedro-org/kedro/issues/2815 we agreed to

- Adopt [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) for `kedro-datasets`
- In the absence of a stricter policy for the rest of the plugins, we treat removing Python version support as non-breaking when it's no longer officially supported (see https://devguide.python.org/versions/)

![Python versions support chart](https://github.com/kedro-org/kedro-plugins/assets/316517/a6ff910a-ab63-4339-894a-0fe30609b8d5)

We should put this policy into practice by

- Dropping Python 3.8 support in `kedro-datasets` ASAP, hence in the next breaking release (end of NEP 29 window as of Apr 14, 2023)
- Dropping Python 3.8 support for the rest of the plugins in October 2024

PR #419, opened 2023-10-25
chore: Drop support for Python 3.7 on kedro-datasets

lrcouto:
## Description

Drops support for Python 3.7 on `kedro-datasets`. Part of https://github.com/kedro-org/kedro/issues/2158.

## Checklist

### DIFF
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -155,7 +155,7 @@ extras_require["docs"] = [
     "Jinja2<3.1.0",
 ]
 extras_require["test"] = [
-    "adlfs~=2023.1; python_version >= '3.8'",
+    "adlfs~=2023.1",
     "bandit>=1.6.2, <2.0",
     "behave==1.2.6",
     "biopython~=1.73",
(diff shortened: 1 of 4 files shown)


kedro-datasets/setup.py
```diff
@@ -155,7 +155,6 @@ def _collect_requirements(requires):
     "Jinja2<3.1.0",
 ]
 extras_require["test"] = [
-    "adlfs>=2021.7.1, <=2022.2; python_version == '3.7'",
```
astrojuanlu, line 158: 👍🏽 

And `>=3.8` markers are redundant and could be dropped too.

astrojuanlu (commented):
Is the plan to have one last release with 3.8+ support before applying https://github.com/kedro-org/kedro-plugins/issues/404 ? Or should we go directly to it and drop 3.8 for kedro-datasets 2.0? @merelcht

merelcht:
> Is the plan to have one last release with 3.8+ support before applying #404 ? Or should we go directly to it and drop 3.8 for kedro-datasets 2.0? @merelcht

IMO we should also drop 3.8 now we have decided to use NEP 29 for `kedro-datasets`. We can do it in a separate PR if we want to have a better discussion there.

merelcht (approved):
LGTM! 👍

astrojuanlu (approved):
💯

PR #422, opened 2023-10-29
fix(telemetry): remove double execution by moving to after catalog created hook

### LINKED ISSUE #382: Double execution of `after_catalog_created` hook if opted into `kedro-telemetry`
## Description

As part of investigating an issue in kedro-azureml (https://github.com/getindata/kedro-azureml/issues/79) We came across the fact that any `after_catalog_created` hooks will be executed twice if the user has opted in `kedro-telemetry`. This is because in telemetry `after_context_created` hook the [context.catalog](https://github.com/kedro-org/kedro-plugins/blob/080c73a8b5dfd5688750d60081f3e754668a927b/kedro-telemetry/kedro_telemetry/plugin.py#L126) is accessed which in turn triggers the `after_catalog_created` [hook](https://github.com/kedro-org/kedro/blob/05002ea96534504887aef7ae18ad529a116c2142/kedro/framework/context/context.py#L287) for the first and I'd assume unexpected time.

This might lead to bugs and unintended behaviours. Is this by design or could one open a PR that as a potential solution. E.g. one could create a stateful KedroTelemetryProjectHooks that only sends the Project data as part of the after_catalog_created hook. Better solutions might be found when thinking about it more than 5 min 😅 

## Steps to Reproduce
1. install kedro spaceflight starter
2. add a hook that prints something in a `after_catalog_created` hook
```
import logging

from kedro.framework.hooks import hook_impl
from kedro.io import DataCatalog


class DataCatalogHooks:
    @property
    def _logger(self):
        return logging.getLogger(self.__class__.__name__)

    @hook_impl
    def after_catalog_created(self, catalog: DataCatalog) -> None:
        # self._logger.info("************************I was executed!")
        print("************************I was executed!")
```
4. run the kedro pipeline

## Expected Result
I would have expected `************************I was executed!` to only appear once but it might not be so by design.

## Actual Result
```
❯ kedro run
[10/09/23 21:41:59] INFO     Kedro project space-telemetry                                                                                                          session.py:364
************************I was executed!
************************I was executed!
[10/09/23 21:42:01] INFO     Loading data from 'companies' (CSVDataset)...
```

## Your Environment

python=3.10

kedro==0.18.13
kedro-datasets==1.7.1
kedro-telemetry==0.2.5
kedro-viz==6.5.0

fdroessler:
## Description
This PR fixes #382 by moving the sending of the heap to the `after_catalog_created` hook instead of the `after_context_created` hook. This prevents the creation of the catalog twice and thus double execution of `after_catalog_created` hooks as part of any downstream project.

## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,4 +1,5 @@
 # Upcoming release
+* Fixed double execution of `after_catalog_created` hook by moving the logic of determining and sending of project statistics from `after_context_created` to the `after_catalog_created` hook.
 
 # Release 0.3.0
 * Added support for Python 3.11
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -113,8 +113,11 @@ class KedroTelemetryProjectHooks:
     @hook_impl
     def after_context_created(self, context):
         """Hook implementation to send project statistics data to Heap"""
-        consent = _check_for_telemetry_consent(context.project_path)
-        if not consent:
+        self.consent = _check_for_telemetry_consent(context.project_path)
+
+    @hook_impl
+    def after_catalog_created(self, catalog):
+        if not self.consent:
             logger.debug(
                 "Kedro-Telemetry is installed, but you have opted out of "
                 "sharing usage analytics so none will be collected.",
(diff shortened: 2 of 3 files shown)


astrojuanlu:
Thanks a lot @fdroessler! Could you please signoff your commit so the DCO check passes? If you click on it you will find instructions. If it was only one commit you can do `git commit --amend --signoff --no-edit`, if you had more, `git rebase --signoff`. After rewriting the history you will need to do `git push --force`.

And if you keep working on this branch, please use the `-s/--signoff` flag when committing, for example `git commit -s -m 'Add feature'`.

noklam (approved):
Nice fix!

ankatiyar (approved):
Thank you @fdroessler 💯

PR #428, opened 2023-11-10
docs(datasets): Improve documentation index

astrojuanlu:
## Description

Now all the classes are listed in the index page, saving the reader one click.

## Checklist


stichbury (approved):
Nice one, yes, thanks!

Do you want to fix that project name in `conf.py` while you're in there?

stichbury:
I like it! Let's keep it but maybe add a comment in the README so anyone editing it knows that it'll be mirrored into the datasets docs.

I think we should hide the long list of classes with `:hidden:` though as it's pretty ugly and doesn't seem that necessary as they're in the sidebar

image

kedro-datasets/docs/source/index.md
```diff
@@ -1,12 +1,20 @@
...
 
-.. rubric:: Description
+## API documentation
 
+```{rubric} Description
+```
+
+```{eval-rst}
 .. automodule:: kedro_datasets
+```
```
stichbury, line 13: Maybe don't need all this as shown? Just hide them?

astrojuanlu: In that case, let me go back to the previous structure, but with the nicer README!

astrojuanlu:
Apparently, one can't prevent items on the navbar from being collapsed due to an ancient bug in sphinx-rtd-theme https://github.com/readthedocs/sphinx_rtd_theme/issues/455

astrojuanlu:
I was defeated by the Sphinx TOC this time, so this ended up being much less ambitious. But at least now the documentation index shows some info instead of a barebones link.

_Also_, I used this opportunity to move the files autogenerated by `sphinx.ext.autosummary` to `api/`. Since we are using `intersphinx` anyway, this should not matter, and also moves us closer to https://github.com/kedro-org/kedro/issues/2483

stichbury:
I made a couple of amends to docs links in a commit, plus changed the language slightly.
This is good to go from my perspective. What about adding a shield for the docs build to the page?

merelcht (approved):

astrojuanlu:
Terrible mistake from my side, API pages are broken. I'm partially reverting this rn.

astrojuanlu:
False alarm ✔️

PR #447, opened 2023-11-26
feat(datasets): support versioning data partitions

deepyaman:
## Description

Support versioning of the underlying dataset (for `PartitionedDataset` and `IncrementalDataset`)

## Development notes

Resolves kedro-org/kedro#521 in Kedro-Datasets

Context: https://kedro-org.slack.com/archives/C054U0LJLAC/p1690985599167069

image

## Checklist


kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
```diff
@@ -28,6 +28,18 @@
 S3_PROTOCOLS = ("s3", "s3a", "s3n")
 
 
+def _grandparent(path: str) -> str:
```
merelcht, line 31: It would be helpful to have a docstring here explaining what the method is doing.

kedro-datasets/tests/partitions/test_partitioned_dataset.py
```diff
@@ -28,7 +28,7 @@ def partitioned_data_pandas():
 
 @pytest.fixture
 def local_csvs(tmp_path, partitioned_data_pandas):
-    local_dir = Path(str(tmp_path / "csvs"))
+    local_dir = local_dir = tmp_path / "csvs"
```
merelcht, line 31:
```suggestion
    local_dir = tmp_path / "csvs"
```

merelcht (commented):
Left some minor comments. Generally the code looks good 👍 I'm wondering if we should add some documentation on how to use this functionality, or do you think it will be clear enough for users just looking at the dataset?

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
deepyaman, line 31:
```suggestion
def _grandparent(path: str) -> str:
    "Validate that the parent's parent has the same name, and return."""
```

deepyaman:
> I'm wondering if we should add some documentation on how to use this functionality, or do you think it will be clear enough for users just looking at the dataset?

I personally don't think it's wholly necessary; you place the versioned key at the top, just like you would with any other dataset, if you want a versioned `PartitionedDataset`.

What's more "complex" is how the versioning actually looks under the hood. However, I think you'd expect 90% of users to just take that for granted, so I think it's fine to not try to explain that right now.

merelcht (approved):
Thanks @deepyaman 👍  Fair point on the docs.

PR #457, opened 2023-11-29
test(datasets): skip hanging doctest in Windows CI

deepyaman:
## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/pillow/image_dataset.py
+++ b/kedro-datasets/kedro_datasets/pillow/image_dataset.py
@@ -22,8 +22,14 @@ class ImageDataset(AbstractVersionedDataset[Image.Image, Image.Image]):
 
     .. code-block:: pycon
 
+        >>> import sys
+        >>>
+        >>> import pytest
         >>> from kedro_datasets.pillow import ImageDataset
         >>>
+        >>> if sys.platform.startswith("win"):
+        ...     pytest.skip("this doctest hangs on Windows CI runner")
+        ...
         >>> dataset = ImageDataset(filepath="https://storage.googleapis.com/gtv-videos-bucket/sample/images/ForBiggerBlazes.jpg")
         >>> image = dataset.load()
         >>> image.show()


deepyaman:
@merelcht There seem to be some CI issues running doctests on Windows; tagging you, because a couple of them are related to image/video, which I think you looked at.

There's also the Spark thing...

Do you want me to just update this PR to skip doctests on Windows in these cases (2 current failures, plus the one already in this PR)?

merelcht:
> @merelcht There seem to be some CI issues running doctests on Windows; tagging you, because a couple of them are related to image/video, which I think you looked at.
> 
> There's also the Spark thing...
> 
> Do you want me to just update this PR to skip doctests on Windows in these cases (2 current failures, plus the one already in this PR)?

Very odd it didn't fail when I pushed.. Yes let's ignore them. I would even suggest maybe not running the doctests checks on Windows at all, because it's not really necessary to check they work on both operating systems especially if this is going to be another Windows pain..

.github/workflows/unit-tests.yml
```diff
@@ -50,6 +50,15 @@ jobs:
...
         run: pip freeze
+      - name: Get native Hadoop libraries
+        if: inputs.os == 'windows-latest'
+        run: |
+          # https://cwiki.apache.org/confluence/display/HADOOP2/WindowsProblems
+          git clone https://github.com/cdarlint/winutils.git
+
+          # https://github.com/cdarlint/winutils#how-to-use
+          HADOOP_HOME=$PWD/winutils/hadoop-2.7.3
+          Add-Content $env:GITHUB_PATH "$HADOOP_HOME/bin"
```
deepyaman, line 61:
```suggestion
          echo "$HADOOP_HOME/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```

.github/workflows/unit-tests.yml
```diff
@@ -50,6 +50,18 @@ jobs:
...
+        if: inputs.os == 'windows-latest'
+        shell: bash
+        run: |
+          # https://cwiki.apache.org/confluence/display/HADOOP2/WindowsProblems
+          git clone https://github.com/cdarlint/winutils.git
+
+          # https://github.com/cdarlint/winutils#how-to-use
+          PWD="${PWD:2}"
+          echo "$PWD"
+          HADOOP_HOME="D:${PWD//\//\\\\}\\winutils\\hadoop-2.7.3"
```
deepyaman, line 63:
```suggestion
          HADOOP_HOME="D:${PWD//\//\\}\winutils\hadoop-2.7.3"
```

.github/workflows/unit-tests.yml
```diff
@@ -61,7 +73,9 @@ jobs:
         run: |
           cd ${{ inputs.plugin }}
           pytest tests
+      - name: Show PATH
+        run: python -c'import os;print(os.environ["PATH"])'
```
deepyaman, line 77:
```suggestion
```

deepyaman:
> > @merelcht There seem to be some CI issues running doctests on Windows; tagging you, because a couple of them are related to image/video, which I think you looked at.
> > There's also the Spark thing...
> > Do you want me to just update this PR to skip doctests on Windows in these cases (2 current failures, plus the one already in this PR)?
> 
> Very odd it didn't fail when I pushed.. Yes let's ignore them. I would even suggest maybe not running the doctests checks on Windows at all, because it's not really necessary to check they work on both operating systems especially if this is going to be another Windows pain..

The image thing was failing, but otherwise the video issue was just on an assertion that I think isn't very important (in a doctest; it wasn't showing anything I can imagine a user wanting to do). Beyond that, I tried a lot, but I can't get Spark to work; there's already precedent for skipping Spark tests on Windows CI here.

Other than that, everything works.

merelcht (approved):
Thanks for fixing this @deepyaman ⭐

PR #460, opened 2023-12-01
fix(datasets): correct pandas-gbq as py311 dependency

kuruonur1:
## Description
pandas-gbq dependency is only updated in _extras_ **tests** but not in **all**.
If a package is using a dependency manager (say poetry) on py311 it is not possible to install kedro-datasets because of the missing specification.

## Development notes
just updated the dependency in **all** section in  setup.py

## Checklist


astrojuanlu (approved):
Thank you!

AhdraMeraliQB (approved):
Great catch! Thanks 🌟

PR #459, opened 2023-12-01
build: Debug bump s3fs

merelcht:
## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #458, opened 2023-11-30
build: Experiment aws patching

merelcht:
## Description
Debug problems in https://github.com/kedro-org/kedro-plugins/pull/348

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #463, opened 2023-12-05
build(datasets): Bump s3fs

merelcht:
## Description
Attempt to solve issues with tests in:

- https://github.com/kedro-org/kedro-plugins/pull/348
- https://github.com/kedro-org/kedro-plugins/pull/458

The problem that surfaced in #348 is described here: https://github.com/aio-libs/aiobotocore/issues/755 Lots of people have provided patches, so I have used one of the most up to date ones and it seems to work almost perfectly.

## Development notes
Tests that are still broken and have been commented out for the time being:
- `tests/dask/test_parquet_dataset.py` test_pass_credentials
- `tests/spark/test_spark_dataset.py` test_no_version

See https://github.com/kedro-org/kedro-plugins/pull/459 for some more things I tried to get these tests passing. We just need to change the way the mocking is done in these, but I'm out of ideas at the moment. Probably won't be too hard to fix with a fresh pair of eyes.

## Checklist

### DIFF
--- a/kedro-datasets/tests/conftest.py
+++ b/kedro-datasets/tests/conftest.py
@@ -5,8 +5,78 @@ discover them automatically. More info here:
 https://docs.pytest.org/en/latest/fixture.html
 """
 
+from typing import Callable
+from unittest.mock import MagicMock
+
+import aiobotocore.awsrequest
+import aiobotocore.endpoint
+import aiohttp
+import aiohttp.client_reqrep
+import aiohttp.typedefs
+import botocore.awsrequest
+import botocore.model
 from kedro.io.core import generate_timestamp
 from pytest import fixture
+
+BUCKET_NAME = "test_bucket"
+IP_ADDRESS = "127.0.0.1"
+PORT = 5555
+ENDPOINT_URI = f"http://{IP_ADDRESS}:{PORT}/"
+
+
+"""
+Patch aiobotocore to work with moto
+See https://github.com/aio-libs/aiobotocore/issues/755
+"""
+
+
+class MockAWSResponse(aiobotocore.awsrequest.AioAWSResponse):
+    def __init__(self, response: botocore.awsrequest.AWSResponse):
+        self._moto_response = response
+        self.status_code = response.status_code
+        self.raw = MockHttpClientResponse(response)
+
+    # adapt async methods to use moto's response
+    async def _content_prop(self) -> bytes:
+        return self._moto_response.content
+
+    async def _text_prop(self) -> str:
+        return self._moto_response.text
+
+
+class MockHttpClientResponse(aiohttp.client_reqrep.ClientResponse):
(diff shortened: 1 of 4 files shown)


kedro-datasets/tests/dask/test_parquet_dataset.py
```diff
@@ -91,23 +91,22 @@ def test_empty_credentials_load(self, bad_credentials):
...
-
-        assert client_mock.call_count == 1
-        args, kwargs = client_mock.call_args_list[0]
-        assert args == ("s3",)
-        assert kwargs["aws_access_key_id"] == AWS_CREDENTIALS["key"]
-        assert kwargs["aws_secret_access_key"] == AWS_CREDENTIALS["secret"]
-
-    @pytest.mark.usefixtures("mocked_s3_bucket")
-    def test_save_data(self, s3_dataset):
+    # def test_pass_credentials(self, mocker):
```
astrojuanlu, line 94: Could we do

```
@pytest.mark.xfail
```

? https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail

MatthiasRoels:
Thanks @merelcht for picking it up and fixing the failing tests! This approach is cleaner than what I initially thought it would. 

I still believe my approach (with moto in server mode) could work, but I wasn't able to figure out why it didn't. The `s3fs` package is using it for their tests, so if I have time, I will look into their approach to see what they do differently than what I had.

kedro-datasets/tests/conftest.py
```diff
@@ -5,9 +5,79 @@
...
+
+    async def _text_prop(self) -> str:
+        return self._moto_response.text
+
+
+class MockHttpClientResponse(aiohttp.client_reqrep.ClientResponse):
+    def __init__(self, response: botocore.awsrequest.AWSResponse):
+        async def read(self, n: int = -1) -> bytes:
+            # streaming/range requests. used by s3fs
+            return response.content
```
SajidAlamQB, line 51: Should this be in here? I would have thought it would be in `MockHttpClientResponse` and not in the `__init__`.

SajidAlamQB (approved):
LGTM overall just left a minor comment, thank you @merelcht! 👍

kedro-datasets/tests/conftest.py
merelcht: Good question, tbh I didn't try other variations of this specific patch. I used the most recent and clean snippet I could find in the threads here https://github.com/aio-libs/aiobotocore/issues/755 https://gist.github.com/giles-betteromics/12e68b88e261402fbe31c2e918ea4168, assuming that writing my own wasn't going to be better than all the iterations from the community.

DimedS (approved):
Thank you, @merelcht , good solution!

PR #473, opened 2023-12-12
build(docker): Release 0.5.0

lrcouto:
## Description

Release `kedro-docker` 0.5.0

## Checklist


astrojuanlu:
https://github.com/yaml/pyyaml/issues/601#issuecomment-1813963845

kedro-telemetry/kedro_telemetry/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Kedro plugin for collecting Kedro usage data."""
 
-__version__ = "0.3.0"
+__version__ = "0.5.0"
```
merelcht, line 3: No need to also bump `kedro-telemetry`here.

lrcouto: Ah this one was an accident, I was on the wrong directory

astrojuanlu:
Somehow the logs are wrapped in a more narrow max width now and that's why some assertions fail. For example one test expects `INFO - Pipeline execution completed successfully` but instead the logs show

```
                          INFO     Pipeline execution completed          runner.py:112
                                   successfully.
```

(notice how the line is broken)

https://github.com/kedro-org/kedro-plugins/actions/runs/7186100161/job/19570771296?pr=473#step:7:168

AhdraMeraliQB:
> Somehow the logs are wrapped in a more narrow max width now and that's why some assertions fail. For example one test expects `INFO - Pipeline execution completed successfully` but instead the logs show
> 
> ```
>                           INFO     Pipeline execution completed          runner.py:112
>                                    successfully.
> ```
> 
> (notice how the line is broken)
> 
> https://github.com/kedro-org/kedro-plugins/actions/runs/7186100161/job/19570771296?pr=473#step:7:168

Might be helpful: https://stackoverflow.com/questions/14446861/matching-discontinuous-interrupted-strings

noklam:
I starting to suspect if this PR accidentally bring in a bug #92, I am not certain because before that our e2e-test never try to write on mounted storage. `pandas-iris` only read from it but now we changer to `spaceflights-pandas` which write to the mounted volume.

The problem only appear when using with the `--uid` and `--gid` options.

merelcht:
> I starting to suspect if this PR accidentally bring in a bug #92, I am not certain because before that our e2e-test never try to write on mounted storage. `pandas-iris` only read from it but now we changer to `spaceflights-pandas` which write to the mounted volume.
> 
> The problem only appear when using with the `--uid` and `--gid` options.

I propose we do the following:
1. Create a simpler test started for `kedro-docker` e2e tests that doesn't need to write to mounted volume
2. In a separate task fix the issue with mounted volume in combination with the `--uid` and `--gid` options

It would be a shame to have the compatibility with Kedro 0.19.0 blocked by this issue which might have been broken for a much longer time already.

noklam:
@merelcht After reading this carefully, I have changed my mind and I don't think it is a broken feature, it is just a bad test.

> Create a simpler test started for kedro-docker e2e tests that doesn't need to write to mounted volume

I agree with this. The permission problem is not a bug because UID/GID need to match with the host. For someone who set this manually should be aware of the consequence. 

I propose we update the test. Instead of testing with a pipeline run, we can use `kedro docker ipython`, and assert PID and UID to see if it match with the CLI arguments `kedro docker ipython`. The original test isn't really testing UID/GID, it could still work even if the UID and GID wasn't set properly.

merelcht (commented):
The changes look good! Great work fixing it ⭐ 

I would suggest doing the release in a separate PR though, to make it easier to trace back the different changes. The release PR should ideally just be the version bump + release notes.

PR #498, opened 2023-12-21
fix: airflow metadata

AhdraMeraliQB:
## Description

Fix failing airflow build: https://github.com/kedro-org/kedro-plugins/issues/494
Tests blocking https://github.com/kedro-org/kedro/pull/3462

## Development notes

Instead of calling cookiecutter directly, the test setup has been changed to use `kedro new` (not unlike the existing kedro-docker e2e setup), without introducing new dependencies, and eliminating the direct reliance of kedro-plugins on on cookiecutter.

Honestly, still unsure why `cookiecutter()` would fail but `cookiecutter` wrapped in `kedro new` is fine 🤔. It looks like the original fixture was copied over from kedro-mlflow, might 0.19.0 be breaking there?

## Checklist

### DIFF
--- a/kedro-airflow/tests/conftest.py
+++ b/kedro-airflow/tests/conftest.py
@@ -52,7 +52,7 @@ def kedro_project(cli_runner):
         "repo_name": "fake-project",
         "python_package": "fake_project",
         "include_example": True,
-        "tools": ['1', '2', '3', '4', '5']
+        "tools": ["1", "2", "3", "4", "5"],
     }
     cookiecutter(
         str(TEMPLATE_PATH),
@@ -107,6 +107,6 @@ def metadata(kedro_project):
         project_name="Hello world !!!",
         kedro_init_version=kedro_version,
         project_path=project_path,
-        tools=['None'],
+        tools=["None"],
         example_pipeline="No",
     )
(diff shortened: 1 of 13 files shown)


noklam:
< However, pyproject.toml is not found when an airflow project is generated, hence trying to remove requirements from a non-existent file causes errors.

How was this airflow project generated? Is it the `kedro-starter` starters?

AhdraMeraliQB:
CC: @noklam 
> < However, pyproject.toml is not found when an airflow project is generated, hence trying to remove requirements from a non-existent file causes errors.
> 
> How was this airflow project generated? Is it the `kedro-starter` starters?

It uses a link to the kedro template in framework's `kedro/templates/project`

DimedS (approved):
Thank you, @AhdraMeraliQB , it's a great work! I agree that it's more clear to use common testing pattern with 'kedro new' instead of manually use cookiecutter.

SajidAlamQB (approved):
This implementation makes sense, but I'm wondering why the original fixture didn't work are we just missing the tools and example fields? But happy to go with this 👍

AhdraMeraliQB:
@SajidAlamQB Unsure, maybe @ankatiyar can shed some light on this when she's back. Ultimately I do think using kedro new instead of cookiecutter aligns more closely with the rest of our testing practices, but it is odd that it resolved that particular issue.

PR #504, opened 2024-01-04
Dataset Preview Refactor 

rashidakanchwala:
## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/matplotlib/matplotlib_writer.py
+++ b/kedro-datasets/kedro_datasets/matplotlib/matplotlib_writer.py
@@ -2,9 +2,10 @@
 files to an underlying filesystem (e.g. local, S3, GCS)."""
 
 import io
+import base64
 from copy import deepcopy
 from pathlib import PurePosixPath
-from typing import Any, NoReturn, Union
+from typing import Any, NoReturn, Union, NewType
 from warnings import warn
 
 import fsspec
@@ -113,6 +114,8 @@ class MatplotlibWriter(
     """
 
     DEFAULT_SAVE_ARGS: dict[str, Any] = {}
+    
+    Image = NewType('Image', str)
 
     def __init__(  # noqa: PLR0913
         self,
@@ -245,3 +248,9 @@ class MatplotlibWriter(
         """Invalidate underlying filesystem caches."""
         filepath = get_filepath_str(self._filepath, self._protocol)
         self._fs.invalidate_cache(filepath)
+
+    def _preview(self) -> Image:
+        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+        with self._fs.open(load_path, mode="rb") as img_file:
+            base64_bytes = base64.b64encode(img_file.read())
+        return base64_bytes.decode("utf-8")
--- a/kedro-datasets/kedro_datasets/pandas/csv_dataset.py
+++ b/kedro-datasets/kedro_datasets/pandas/csv_dataset.py
@@ -5,7 +5,7 @@ import logging
 from copy import deepcopy
 from io import BytesIO
 from pathlib import PurePosixPath
-from typing import Any
+from typing import Any, NewType
 
 import fsspec
 import pandas as pd
@@ -67,6 +67,8 @@ class CSVDataset(AbstractVersionedDataset[pd.DataFrame, pd.DataFrame]):
 
     DEFAULT_LOAD_ARGS: dict[str, Any] = {}
     DEFAULT_SAVE_ARGS: dict[str, Any] = {"index": False}
+    
+    DataFrame = NewType('DataFrame', dict)
 
     def __init__(  # noqa: PLR0913
         self,
@@ -194,7 +196,7 @@ class CSVDataset(AbstractVersionedDataset[pd.DataFrame, pd.DataFrame]):
         filepath = get_filepath_str(self._filepath, self._protocol)
         self._fs.invalidate_cache(filepath)
 
-    def _preview(self, nrows: int = 40) -> dict:
+    def _preview(self, nrows: int = 40) -> DataFrame:
         # Create a copy so it doesn't contaminate the original dataset
         dataset_copy = self._copy()
         dataset_copy._load_args["nrows"] = nrows
--- a/kedro-datasets/kedro_datasets/pandas/excel_dataset.py
+++ b/kedro-datasets/kedro_datasets/pandas/excel_dataset.py
@@ -5,7 +5,7 @@ import logging
 from copy import deepcopy
 from io import BytesIO
 from pathlib import PurePosixPath
-from typing import Any, Union
+from typing import Any, Union, NewType
 
 import fsspec
 import pandas as pd
@@ -107,6 +107,8 @@ class ExcelDataset(
 
     DEFAULT_LOAD_ARGS = {"engine": "openpyxl"}
     DEFAULT_SAVE_ARGS = {"index": False}
+    
+    DataFrame = NewType('DataFrame', dict)
 
     def __init__(  # noqa: PLR0913
         self,
@@ -262,7 +264,7 @@ class ExcelDataset(
         filepath = get_filepath_str(self._filepath, self._protocol)
         self._fs.invalidate_cache(filepath)
 
-    def _preview(self, nrows: int = 40) -> dict:
+    def _preview(self, nrows: int = 40) -> DataFrame:
         # Create a copy so it doesn't contaminate the original dataset
         dataset_copy = self._copy()
         dataset_copy._load_args["nrows"] = nrows
--- a/kedro-datasets/kedro_datasets/plotly/json_dataset.py
+++ b/kedro-datasets/kedro_datasets/plotly/json_dataset.py
@@ -3,7 +3,7 @@ filesystem (e.g.: local, S3, GCS).
 """
 from copy import deepcopy
 from pathlib import PurePosixPath
-from typing import Any, Union
+from typing import Any, Union, NewType
 
 import fsspec
 import plotly.io as pio
@@ -52,6 +52,8 @@ class JSONDataset(
 
     DEFAULT_LOAD_ARGS: dict[str, Any] = {}
     DEFAULT_SAVE_ARGS: dict[str, Any] = {}
+    
+    Plot = NewType('Plot', str)
 
     def __init__(  # noqa: PLR0913
         self,
@@ -167,3 +169,7 @@ class JSONDataset(
     def _invalidate_cache(self) -> None:
         filepath = get_filepath_str(self._filepath, self._protocol)
         self._fs.invalidate_cache(filepath)
+        
+    def _preview(self) -> Plot:
+        plotly_data = self.load()  # Assuming this returns a plotly figure as a string
+        return plotly_data
--- a/kedro-datasets/kedro_datasets/plotly/plotly_dataset.py
+++ b/kedro-datasets/kedro_datasets/plotly/plotly_dataset.py
@@ -3,7 +3,7 @@ file using an underlying filesystem (e.g.: local, S3, GCS). It loads the JSON in
 plotly figure.
 """
 from copy import deepcopy
-from typing import Any
+from typing import Any, NewType
 
 import pandas as pd
 import plotly.express as px
@@ -65,6 +65,8 @@ class PlotlyDataset(JSONDataset):
         >>> assert px.scatter(df_data, x="x1", y="x2") == reloaded
 
     """
+    
+    Plot = NewType('Plot', str)
 
     def __init__(  # noqa: PLR0913
         self,
@@ -148,3 +150,7 @@ class PlotlyDataset(JSONDataset):
         fig.update_layout(template=self._plotly_args.get("theme", "plotly"))
         fig.update_layout(self._plotly_args.get("layout", {}))
         return fig
+    
+    def _preview(self) -> Plot:
+        plotly_data = self.load()  # Assuming this returns a plotly figure as a string
+        return plotly_data
--- a/kedro-datasets/kedro_datasets/tracking/json_dataset.py
+++ b/kedro-datasets/kedro_datasets/tracking/json_dataset.py
@@ -2,12 +2,12 @@
 filesystem (e.g.: local, S3, GCS). It uses native json to handle the JSON file.
 The ``JSONDataset`` is part of Kedro Experiment Tracking. The dataset is versioned by default.
 """
-from typing import NoReturn
+import json
+from typing import NoReturn, NewType
 
-from kedro.io.core import DatasetError
+from kedro.io.core import DatasetError, get_filepath_str
 
 from kedro_datasets.json import json_dataset
-
 
 class JSONDataset(json_dataset.JSONDataset):
     """``JSONDataset`` saves data to a JSON file using an underlying
@@ -41,6 +41,14 @@ class JSONDataset(json_dataset.JSONDataset):
     """
 
     versioned = True
+    
+    JSONData = NewType('JSONData', str)
 
     def _load(self) -> NoReturn:
         raise DatasetError(f"Loading not supported for '{self.__class__.__name__}'")
+
+    def _preview(self) -> JSONData:
+        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+
+        with self._fs.open(load_path, **self._fs_open_args_load) as fs_file:
+            return json.load(fs_file)
--- a/kedro-datasets/kedro_datasets/tracking/metrics_dataset.py
+++ b/kedro-datasets/kedro_datasets/tracking/metrics_dataset.py
@@ -4,7 +4,7 @@ The ``MetricsDataset`` is part of Kedro Experiment Tracking. The dataset is vers
 and only takes metrics of numeric values.
 """
 import json
-from typing import NoReturn
+from typing import NoReturn, NewType
 
 from kedro.io.core import DatasetError, get_filepath_str
 
@@ -43,6 +43,7 @@ class MetricsDataset(json_dataset.JSONDataset):
     """
 
     versioned = True
+    JSONData = NewType('JSONData', str)
 
     def _load(self) -> NoReturn:
         raise DatasetError(f"Loading not supported for '{self.__class__.__name__}'")
@@ -65,3 +66,9 @@ class MetricsDataset(json_dataset.JSONDataset):
             json.dump(data, fs_file, **self._save_args)
 
         self._invalidate_cache()
+        
+    def _preview(self) -> JSONData:
+        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+
+        with self._fs.open(load_path, **self._fs_open_args_load) as fs_file:
+            return json.load(fs_file)


astrojuanlu:
This is the docstring generated by Sphinx (found after running Sphinx with `-vvv` locally)

```
...
[autodoc] => <function JSONDataset.preview at 0x2aa6e7c10>
[app] emitting event: 'autodoc-before-process-signature'(<function JSONDataset.preview at 0x2aa6e7c10>, True)
[app] emitting event: 'autodoc-process-signature'('method', 'kedro_datasets.tracking.JSONDataset.preview', <function JSONDataset.preview at 0x2aa6e7c
[app] emitting event: 'autodoc-process-docstring'('method', 'kedro_datasets.tracking.JSONDataset.preview', <function JSONDataset.preview at 0x2aa6e7c
...
[autodoc] output:

.. py:class:: JSONDataset(*, filepath, save_args=None, version=None, credentials=None, fs_args=None, metadata=None)
   :module: kedro_datasets.tracking
   :canonical: kedro_datasets.tracking.json_dataset.JSONDataset

   ``JSONDataset`` saves data to a JSON file using an underlying
...

   .. py:method:: JSONDataset.preview()
      :module: kedro_datasets.tracking

      Load the JSON tracking dataset used in Kedro-viz experiment tracking.


      :rtype: :py:func:`~typing.NewType`\(``JSONTracking``, :py:class:`str`)
```

Which appears properly listed in the Sphinx inventory of the CPython docs (see https://webknjaz.github.io/intersphinx-untangled/docs.python.org/)

But! Only in Python 3.9! See https://docs.python.org/3.10/library/typing.html#typing.NewType

> Changed in version 3.10: NewType is now a class rather than a function.

(A reminder that CPython doesn't follow Semantic Versioning, but rather a rolling deprecation window - and also, this was merged quite hastily, see https://github.com/python/cpython/pull/27250 for context)

Anyway, this is important because it teaches us how to fix the issue:

```diff
diff --git a/kedro-datasets/docs/source/conf.py b/kedro-datasets/docs/source/conf.py
index 12e861fd..6e729152 100644
--- a/kedro-datasets/docs/source/conf.py
+++ b/kedro-datasets/docs/source/conf.py
@@ -93,6 +93,7 @@ exclude_patterns = [
 
 intersphinx_mapping = {
     "kedro": ("https://docs.kedro.org/en/stable/", None),
+    "python": ("https://docs.python.org/3.9/", None),
 }
 
 type_targets = {
```

and also explains why you couldn't reproduce this locally @rashidakanchwala (since you were using Python 3.10).

Please try my patch above and let me know how it goes.

kedro-datasets/kedro_datasets/constants.py
```diff
@@ -0,0 +1,11 @@
+from typing import NewType
```
merelcht, line 1: I think it will be helpful to have a short explanation at the top of the file of what these constants are for.

kedro-datasets/kedro_datasets/constants.py
```diff
@@ -0,0 +1,11 @@
...
+
+Dataframe = NewType("Dataframe", dict)
+Image = NewType("Image", str)
+Plot = NewType("Plot", dict)
+JSON = NewType("JSON", str)
+
+
+# experiment_tracking datasets
```
merelcht, line 9:
```suggestion
# experiment tracking dataset types
```

kedro-datasets/test.pkl
merelcht, line 1: Does this test file and `test.xlsx` need to be added to the repo?

merelcht (commented):
I'm very excited for these changes 👏 ! I left some small comments and questions.

kedro-datasets/test.pkl
rashidakanchwala: oh, i need to remove this. thanks for flagging.

kedro-datasets/tests/pandas/test_csv_dataset.py
```diff
@@ -174,10 +175,14 @@ def test_storage_options_dropped(self, load_args, save_args, caplog, tmp_path):
...
-        """Test _preview returns the correct data structure."""
+        """Test preview returns the correct data structure."""
         csv_dataset.save(dummy_dataframe)
-        previewed = csv_dataset._preview(nrows=nrows)
+        previewed = csv_dataset.preview(nrows=nrows)
         assert previewed == expected
+        assert (
+            inspect.signature(csv_dataset.preview).return_annotation.__name__
+            == "Dataframe"
+        )
```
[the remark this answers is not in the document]
rashidakanchwala: exactly `type(preview)` will return a dict or a string. this is the only way to figure what the alias is.

kedro-datasets/kedro_datasets/constants.py
noklam: Agree. Is there going to an example in the documentation to show that you can extend custom dataset for this preview function?

kedro-datasets/kedro_datasets/constants.py
```diff
@@ -0,0 +1,16 @@
...
+`constants.py` defines custom data types for Kedro-viz integration. It uses NewType from the typing module.
+These types are used to facilitate data rendering in the Kedro-viz front-end.
+"""
+
+from typing import NewType
+
+Dataframe = NewType("Dataframe", dict)
+Image = NewType("Image", str)
+Plot = NewType("Plot", dict)
+JSON = NewType("JSON", str)
```
noklam, line 11: Can you explain why `Dataframe` is a subclass of `dict` here? I am not too familiar with this `typing`, why do we choose `NewType` over `TypeVar`?

noklam (commented):
A nice change to enable extending `preview` for generic datasets! I have some concern about the `NewType`, is this defined in `Kedro-viz` too? This should be a specification coming from kedro-viz, `kedro-datasets` is only implementing it. 

I also feel `constants.py` is not the right name for this, argubably it shouldn't be a public module because we don't expect usage like from `kedro_datasets.constants import Image` outside of `kedro-datasets`.However we cannot import from `kedro-viz` because this will introduce a dependency issue.

maybe something like `[_typing](https://github.com/unionai-oss/pandera/blob/main/pandera/typing/dask.py)` or `_internal`? 

I also wish there are some way to indicate that `preview` is a `kedro-viz` specific things, it feels incorrect to have plugin specific implementation here, particularly `load` and `preview` sounds very similar but `preview` isn't that useful without viz. For example, PlotlyDataset preview method does not return a figure but JSON.

astrojuanlu:
> particularly load and preview sounds very similar but preview isn't that useful without viz. For example, PlotlyDataset preview method does not return a figure but JSON.

Agreed. In fact, a user suggested overloading `_describe` for this https://github.com/kedro-org/kedro-viz/issues/1714

merelcht:
> > particularly load and preview sounds very similar but preview isn't that useful without viz. For example, PlotlyDataset preview method does not return a figure but JSON.
> 
> Agreed. In fact, a user suggested overloading `_describe` for this [kedro-org/kedro-viz#1714](https://github.com/kedro-org/kedro-viz/issues/1714)

The main reason why we didn't go for `_describe` is that it's meant to _describe_ the dataset and not return anything that shows part of the data, like preview is meant to do and because describe is an abstract method that needs to be implemented, so if we want consistency across all datasets we'd have to update all of them.

rashidakanchwala:
> A nice change to enable extending `preview` for generic datasets! I have some concern about the `NewType`, is this defined in `Kedro-viz` too? This should be a specification coming from kedro-viz, `kedro-datasets` is only implementing it.
> 
> I also feel `constants.py` is not the right name for this, argubably it shouldn't be a public module because we don't expect usage like from `kedro_datasets.constants import Image` outside of `kedro-datasets`.However we cannot import from `kedro-viz` because this will introduce a dependency issue.
> 
> maybe something like `[_typing](https://github.com/unionai-oss/pandera/blob/main/pandera/typing/dask.py)` or `_internal`?
> 
> I also wish there are some way to indicate that `preview` is a `kedro-viz` specific things, it feels incorrect to have plugin specific implementation here, particularly `load` and `preview` sounds very similar but `preview` isn't that useful without viz. For example, PlotlyDataset preview method does not return a figure but JSON.

I have addressed concerns in another post. I like the idea of having it under `typing` folder ...would this have an underscore before indicating private ?

noklam:
Thanks @rashidakanchwala.
```
    def preview(self, nrows: int = 5) -> Dataframe:
        """
        Generate a preview of the dataset with a specified number of rows.
        ...
        return data.to_dict(orient="split")
```
I think we need to make it clear to users, because it is a bit unintuitive now. This is the definition of `CSVDataset`, which return a `Dataframe` which is a dictionary itself.

I think it's nice to address notebook's user experience, I wonder can we use something more native like `__repr__` that notebook used for rendering. So when a user want to see the chart, they just do `dataset` instead of `dataset.preview()`. Right now I am not convinced that `preview` is a nice public interface, because there are hidden kedro-viz internal implementation to decide what it should return.


>I have addressed concerns in another post. I like the idea of having it under typing folder ...would this have an underscore before indicating private ?

I prefer `_typing` because this is not `kedro-datasets` specfication, but `kedro-viz`. I don't expect these types get used outside of `kedro-datasets` (they are not meant to be imported). On the other hand, these types should exist in `kedro-viz` and make sense to be public. I wish there is a way that we don't need to duplicate the definition, but I don't have an idea here.

Would like to get some opinions from @merelcht @astrojuanlu

antonymilne (commented):
This generally looks like the right idea 👍  I agree with pretty much all of @noklam's comments though. `_typing.py` sounds like a better name for the new file.

Fully agree that it's not very nice defining these types here given that they are kedro-viz specific here but unfortunately there's not another way to do it without making kedro-viz a dependency of kedro-datasets, which would not be good.

Having previews in Jupyter notebooks is something I've always wanted too. See https://ipython.readthedocs.io/en/stable/config/integrating.html#custom-methods and https://stackoverflow.com/questions/67253015/ipython-priority-order-of-repr-methods-using-display for useful information on this.

Probably the right way to do it now would be to keep these `preview` methods similar to how they are now and then write some new `_repr_`/`_ipython_display_` method that maybe calls `preview`.

For now I think that this should all be kept separate from `describe` (which tbh I've never really understood the point of, but given how things currently stand I don't think should be changed right now anyway).

astrojuanlu:
> The main reason why we didn't go for _describe is that it's meant to describe the dataset and not return anything that shows part of the data, like preview is meant to do and because describe is an abstract method that needs to be implemented, so if we want consistency across all datasets we'd have to update all of them.

I understand, but at the same time, what's wrong with doing this?

```python
def preview():
    return Image(...)  # Current API as the time of writing

def _describe(self) -> dict[str, Any]:
    return {
        "filepath": self._filepath,
        "protocol": self._protocol,
        "load_args": self._load_args,
        "save_args": self._save_args,
        "version": self._version,
        "preview": self.preview(),  # Reusing the result from the method!
    }
```

This is not different from how IPython `_repr_*` methods are implemented in the wild. See Pillow for example:

https://github.com/python-pillow/Pillow/blob/1b6723967440cf8474a9bd1e1c394c90c5c2f986/src/PIL/Image.py#L659-L664

We could go further and namespace the preview in `_describe`, since it's a Kedro-Viz thing at least for now:

```python
def _describe(self) -> dict[str, Any]:
    return {
        "filepath": self._filepath,
        "protocol": self._protocol,
        "load_args": self._load_args,
        "save_args": self._save_args,
        "version": self._version,
        "kedro_viz": {
            "preview": self.preview(),  # Reusing the result from the method!
        }
    }
```

or going _even further_ and enabling downstream plugins to add stuff to `_describe`:

```python
def _describe(self) -> dict[str, Any]:
    description = {
        "filepath": self._filepath,
        "protocol": self._protocol,
        "load_args": self._load_args,
        "save_args": self._save_args,
        "version": self._version,
    }

    # description |= (  # PEP 584, Python 3.9+ if we think "flat is better than nested"
    description["extras"] = (  # If we think "Namespaces are one honking great idea"
        self._extra_describe_fields()  # Figure an extension mechanism for datasets
    )

    return description
```

The advantage of using `_describe` for all this is that

1. Unlike `preview() ` as proposed here, it's a private method. With an upcoming redesign of datasets, it's very likely there will be some breakage to private methods - which will be annoying, sure, but at least we would not be breaking public API.
   - A possible retort to this is "let's keep `_preview` private", which I'd totally support. But then Viz would continue using private APIs. Then why not use `.describe()`, which is already public? Hence my proposal 🙃
2. `_describe` is already there, it declares an arbitrary dictionary `dict[str, t.Any]`, and so it doesn't require extra typing hoops. It can be easily extended - dare I say it was _designed for extensibility_ in the first place?
   - I didn't invent this idea https://github.com/kedro-org/kedro-viz/issues/1714

These are my two cents, won't push back any further.

merelcht:
> I understand, but at the same time, what's wrong with doing this?
> 
> ```python
> def preview():
>     return Image(...)  # Current API as the time of writing
> 
> def _describe(self) -> dict[str, Any]:
>     return {
>         "filepath": self._filepath,
>         "protocol": self._protocol,
>         "load_args": self._load_args,
>         "save_args": self._save_args,
>         "version": self._version,
>         "preview": self.preview(),  # Reusing the result from the method!
>     }
> ```

How is this any different though? We'd still need to add preview methods, but now we'd also call it in `_describe`?

I think a fundamental point that might be missing is that the need for the `NewType` comes from viz needing to know something more about the preview other than just the return type. If it all preview methods return "str" or "dict", viz can't figure out how to render the previews. That's why, at the moment, we rely on exact dataset type checking in kedro-viz e.g. 

```python
def is_metric_node(self):
        """Check if the current node is a metrics node."""
        return self.dataset_type in (
            "tracking.metrics_dataset.MetricsDataset",
            "tracking.metrics_dataset.MetricsDataSet",
        )
```

I don't see how using `_describe` would solve that, without still adding preview methods.

kedro-datasets/kedro_datasets/constants.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+# experiment tracking datasets types
+MetricsTracking = NewType("MetricsTracking", str)
+JSONTracking = NewType("JSONTracking", str)
```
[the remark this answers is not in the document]
rashidakanchwala: you are right. i think it should be a dict.

astrojuanlu:
> I think a fundamental point that might be missing is that the need for the NewType comes from viz needing to know something more about the preview other than just the return type. If it all preview methods return "str" or "dict", viz can't figure out how to render the previews. That's why, at the moment, we rely on exact dataset type checking in kedro-viz e.g.

Fair point. There are other possible solutions for this problem though. One could be returning a wrapper object. For example

```python
@dataclass
class PreviewResult:
    data: t.Any
    repr_type: str


def is_metric_node(preview: PreviewResult):
    return preview.repr_type in (
        "tracking.metrics_dataset.MetricsDataset",
        "tracking.metrics_dataset.MetricsDataSet",
    )
```

Another one could be essentially reusing IPython repr system https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display and implement `_repr_html_`, `_repr_png_` etc with some logic that chooses the "richest" representation.

---

I think (1) we should avoid reinventing the wheel and (2) we should be careful with using Python `typing` at runtime. It evolves really (too?) fast, makes incompatible changes between Python versions, and it has questionable design choices. Plus, it was only meant for static type checking (then projects like Pydantic started abusing it for runtime behavior too).

kedro-datasets/kedro_datasets/constants.py
```diff
@@ -0,0 +1,16 @@
+"""
+`constants.py` defines custom data types for Kedro-viz integration. It uses NewType from the typing module.
+These types are used to facilitate data rendering in the Kedro-viz front-end.
+"""
+
+from typing import NewType
+
+Dataframe = NewType("Dataframe", dict)
```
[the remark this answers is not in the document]
noklam: I like this change, I get confused when I read the signature and Table is sounds more familiar in most experiment tracking tools.

i.e. https://clear.ml/docs/latest/docs/guides/frameworks/pytorch/notebooks/table/download_and_preprocessing/#plots-tables

rashidakanchwala:
We had a discussion and reached several decisions regarding this refactoring PR:

- **NewType Usage**: We acknowledged that using NewType was not the ideal solution, but it remains the best option to ensure that kedro-viz and kedro-datasets are not dependent on each other.

- **Type Renaming**: Ivan proposed renaming the types by adding a "Preview" suffix. For example, renaming Table to TablePreview. This change will be implemented to improve clarity.

- **preview() Method Accessibility**: It was decided to keep the preview() method public. This decision has minimal impact on kedro since it's not part of the AbstractDataset, and it doesn't affect the core functionality of kedro.

- **Specificity of Preview**: We acknowledged that certain datasets, like Plotly preview, are currently tailored to kedro-viz and may not be suitable for notebooks as they return JSON instead of Plotly figures. The priority right now is to address the decoupling of kedro-viz and kedro-datasets. Extending preview functionality to notebooks is not a top priority at the moment, and we can explore solutions for that in the future.

-- **describe():method**  It was noted that the describe() function serves a different purpose from previewing a dataset and is not the preferred option for this scenario.

These decisions will help us maintain the modularity and independence of the kedro-viz and kedro-datasets projects while addressing specific use cases and improving clarity in our codebase.

@merelcht

merelcht (approved):
This looks great ⭐ I think the only thing missing is just a note in the release notes about the changes.

ravi-kumar-pilla (approved):
LGTM !!

antonymilne:
I was too slow to respond before this got merged, but I'm still happy enough with this approach anyway so wouldn't have raised any objections to it. Just leaving some comments here in case we want to return to the issue in future... 🙂 

@astrojuanlu 
> Fair point. There are other possible solutions for this problem though. One could be returning a wrapper object. For example
> 
> ```python
> @dataclass
> class PreviewResult:
>     data: t.Any
>     repr_type: str
> 

Yeah, I also considered this. Or even more simply just return a tuple `(repr_type, data)` [like the ipython `_repr_*_` methods do](https://ipython.readthedocs.io/en/stable/config/integrating.html#metadata).

The main advantage of this approach IMO is that it enables you to use an enum for the `repr_type` so that someone who writes their own dataset knows what all the supported types are. Currently they just have to look at a list of `NewType`s to figure it out, which is much less nice.

The main disadvantage of this is that it feels less obvious/intuitive to me (which is very subjective, I know). I absolutely 100% share your two big concerns about using runtime `typing` to solve this and have general distaste for the Python typing system, but for me practical beats purity here, and I think that a type hint is a better UX than wrapping into a dataclass or tuple. But maybe that's just because I use pydantic a lot (and kedro-viz uses FastAPI so it also part of that stack). It's a flagrant abuse of the static type hints indeed, but for better or worse it's a well established and recognised one by now.

An alternative would also be to have an attribute `preview_return_type = "table"`. This feels unpleasantly detached from the `preview` function itself though, hence my preference to directly affect the `preview` function, either through type hints or the actual returned type as you suggest.

> Another one could be essentially reusing IPython repr system https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display and implement `_repr_html_`, `_repr_png_` etc with some logic that chooses the "richest" representation.

As in https://github.com/kedro-org/kedro-plugins/pull/504#pullrequestreview-1863161114 I very much agree we should try to tie in with this 👍 And the way the priority is decided depends on the frontend and is [kind of documented](https://stackoverflow.com/questions/67253015/ipython-priority-order-of-repr-methods-using-display).

PR #511, opened 2024-01-09
chore(airflow): Bump `apache-airflow` version

### LINKED ISSUE #315: Make `kedro-airflow` compatible with latest version of `apache-airflow`
## Description

With the release of `apache-airflow 2.7.0` the end to end tests for `kedro-airflow` started failing. 
As a temporary fix, we're pinning the version of `apache-airflow ` to `<2.7.0`(https://github.com/kedro-org/kedro-plugins/pull/314) in the test requirements but we should investigate if there's something that needs fixing on Kedro side to make it work.

ankatiyar:
## Description

Fix #315 

## Development notes

- Unpin `apache-airflow` in test requirements
- Changed the starter used in e2e tests to `spaceflights-pandas` to `astro-airflow-iris`

This is because `apache-airflow` has moved to `pydantic v2` in 2.8.0. The spaceflights starters have `kedro-viz` in their requirements and `kedro-viz` still uses `pydantic v1`. So, in the e2e-tests, the installation of the kedro project, which also installs `kedro-viz` was downgrading pydantic to v1 which was causing failures. `astro-airflow-iris` doesn't have `kedro-viz` in the requirements. 

## Checklist

### DIFF
--- a/kedro-airflow/features/airflow.feature
+++ b/kedro-airflow/features/airflow.feature
@@ -11,8 +11,8 @@ Feature: Airflow
     When I execute the airflow command "tasks list project-dummy"
     Then I should get a successful exit code
     And I should get a message including "split"
-    And I should get a message including "make-predictions"
-    And I should get a message including "report-accuracy"
+    And I should get a message including "train"
+    And I should get a message including "predict"
 
   Scenario: Run Airflow task locally with latest Kedro
     Given I have installed kedro version "latest"
@@ -22,4 +22,4 @@ Feature: Airflow
     And I have installed the kedro project package
     When I execute the airflow command "tasks test project-dummy split"
     Then I should get a successful exit code
-    And I should get a message including "Loading data from companies"
+    And I should get a message including "Loading data from example_iris_data"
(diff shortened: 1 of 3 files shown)


kedro-airflow/features/airflow.feature
```diff
@@ -10,16 +10,16 @@ Feature: Airflow
...
     Given I have installed kedro version "latest"
     And I have prepared a config file
     And I have run a non-interactive kedro new
     And I have executed the kedro command "airflow create -t ../airflow/dags/"
     And I have installed the kedro project package
-    When I execute the airflow command "tasks test project-dummy preprocess-companies-node"
+    When I execute the airflow command "tasks test project-dummy split"
     Then I should get a successful exit code
-    And I should get a message including "Loading data from companies"
+    And I should get a message including "Loading data"
```
ankatiyar, line 25: I've also clipped this message because the output in the e2e test is not wide enough for the whole message.

DimedS:
Thank you @ankatiyar , it's a good insight! So right now is it impossible to use airflow in projects where kedro-viz used?

ankatiyar:
@DimedS Using `kedro-airflow` does not require `apache-airflow`, it's just a test requirement, so everything should work with `kedro-viz`. The incompatibility is between `apache-airflow` and `kedro-viz` which both use pydantic but different versions. I think users can use a lower version of `apache-airflow` if needed but I don't think there's anything we can do about it.

DimedS (approved):
@ankatiyar , thank you for PR and explanation!

merelcht (approved):
Thanks @ankatiyar 👍

PR #538, opened 2024-02-05
fix(datasets): Drop pyarrow constraint when using snowpark

felipemonroy:
## Description
Fix Snowflake dataset cannot installed with latest pyarrow 14.0.1 #453 

## Development notes
I believe the contraint was put there when snowpark was not available for python 3.11. Nor snowpark nor the dataset needs pyarrow as dependency.

## Checklist

### DIFF
--- a/kedro-datasets/tests/pandas/test_deltatable_dataset.py
+++ b/kedro-datasets/tests/pandas/test_deltatable_dataset.py
@@ -66,7 +66,7 @@ class TestDeltaTableDataset:
         """Test saving by appending new data."""
         deltatable_dataset_from_path.save(dummy_df)
         new_df = pd.DataFrame({"col1": [0, 0], "col2": [1, 1], "col3": [2, 2]})
-        appended = pd.concat([dummy_df, new_df], ignore_index=True)
+        appended = pd.concat([new_df, dummy_df], ignore_index=True)
         deltatable_dataset_from_path.save(new_df)
         reloaded = deltatable_dataset_from_path.load()
         assert_frame_equal(appended, reloaded)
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Happy to merge if the tests pass 👍🏽 

In the meantime, could you please signoff your commit so the DCO check passes? If you click on it you will find instructions. If it was only one commit you can do `git commit --amend --signoff --no-edit`, if you had more, `git rebase --signoff`. After rewriting the history you will need to do `git push --force`.

And if you keep working on this branch, please use the `-s/--signoff` flag when committing, for example `git commit -s -m 'Add feature'`.

kedro-datasets/tests/pandas/test_deltatable_dataset.py
```diff
@@ -66,7 +66,7 @@ def test_append(self, deltatable_dataset_from_path, dummy_df):
         """Test saving by appending new data."""
         deltatable_dataset_from_path.save(dummy_df)
         new_df = pd.DataFrame({"col1": [0, 0], "col2": [1, 1], "col3": [2, 2]})
-        appended = pd.concat([dummy_df, new_df], ignore_index=True)
+        appended = pd.concat([new_df, dummy_df], ignore_index=True)
```
felipemonroy, line 69: Unsure why with the new pyarrow version the append behaviour changes and if this breaks the expected behaviour

noklam: This seems to be a bug and it should have a fix instead of changing the test.

noklam:
Would you be able to help checking which version of pyarrow introduce this regression? Would be great if you can confirm the bug is coming from pandas or kedro-datasets

noklam:
#541 is a temporary fix, I pulled the `main` branch here. I think we can revert the change in tests here but let's wait for the test result.

astrojuanlu (approved):
Thanks @felipemonroy!

noklam (approved):
Appreciate the help! Thanks!

PR #541, opened 2024-02-06
fix(datasets): Debug and fix `kedro-datasets` nightly build failures

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-plugins/issues/540

The latest release of `deltalake 0.15.2` broke one our tests after digging into their [changes](https://github.com/delta-io/delta-rs/releases/tag/python-v0.15.2), I believe some changes in their Rust code has altered the way data is appended and retrieved from the Delta table which is causing different order of rows in the DataFrame. If we aren't to fussed about the ordering I added sorting before comparing the DataFrames in our tests.

## Development notes

Added sorting to the DF before comparison.

## Checklist

### DIFF
--- a/kedro-datasets/setup.py
+++ b/kedro-datasets/setup.py
@@ -70,19 +70,19 @@ polars_require = {
         POLARS,
         "pyarrow>=4.0",
         "xlsx2csv>=0.8.0",
-        "deltalake >= 0.6.2",
+        "lake >= 0.6.2",
     ],
     "polars.EagerPolarsDataset": [
         POLARS,
         "pyarrow>=4.0",
         "xlsx2csv>=0.8.0",
-        "deltalake >= 0.6.2",
+        "lake >= 0.6.2",
     ],
     "polars.LazyPolarsDataset": [
         # Note: there is no Lazy read Excel option, so we exclude xlsx2csv here.
         POLARS,
         "pyarrow>=4.0",
-        "deltalake >= 0.6.2",
+        "lake >= 0.6.2",
     ],
 }
 redis_require = {"redis.PickleDataset": ["redis~=4.1"]}
@@ -96,7 +96,7 @@ spark_require = {
     "spark.SparkDataset": [SPARK, HDFS, S3FS],
     "spark.SparkHiveDataset": [SPARK, HDFS, S3FS],
     "spark.SparkJDBCDataset": [SPARK, HDFS, S3FS],
-    "spark.DeltaTableDataset": [SPARK, HDFS, S3FS, "delta-spark>=1.0, <3.0"],
+    "spark.TableDataset": [SPARK, HDFS, S3FS, "-spark>=1.0, <3.0"],
 }
 svmlight_require = {"svmlight.SVMLightDataset": ["scikit-learn>=1.0.2", "scipy~=1.7.3"]}
 tensorflow_require = {
@@ -184,7 +184,7 @@ extras_require["test"] = [
     "coverage[toml]",
     "dask[complete]>=2021.10",
     "delta-spark>=1.0, <3.0",
-    "deltalake>=0.10.0, !=0.15.2",
+    "deltalake>=0.10.0, <0.15.2",
     "dill~=0.3.1",
     "filelock>=3.4.0, <4.0",
     "gcsfs>=2023.1, <2023.3",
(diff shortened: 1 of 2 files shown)


SajidAlamQB:
Seems like some new changes in `deltalake 0.15.2` is causing one of our tests to fail.

kedro-datasets/setup.py
```diff
@@ -184,7 +184,7 @@ def _collect_requirements(requires):
     "coverage[toml]",
     "dask[complete]>=2021.10",
     "delta-spark>=1.0, <3.0",
-    "deltalake>=0.10.0",
+    "deltalake>=0.10.0, <0.15.2",
```
astrojuanlu, line 187: Doing one test here, one moment

```suggestion
    "deltalake>=0.10.0, !=0.15.2",
```

astrojuanlu:
Problem still exists with 0.15.3. I can't find a changelog.

SajidAlamQB:
> Problem still exists with 0.15.3. I can't find a changelog.

Here is the changelog: https://github.com/delta-io/delta-rs/releases/tag/python-v0.15.2

SajidAlamQB:
> Here is the changelog: https://github.com/delta-io/delta-rs/releases/tag/python-v0.15.2

I've been digging into their changes, and I believe some changes in their Rust code has altered the way data is appended and retrieved from the Delta table which is causing different order of rows in the DataFrame. If we aren't to fussed about the ordering I added sorting before comparing the DataFrames in our tests.

noklam (commented):
Thanks for digging, it is interesting that you find the issue is from `deltalake`.

Is it `deltalake`, `pyarrow` or actually the combination of them? I'd argue the change should be either pinning the upper bound or update the DeltaTableDataset, because now the `append` function is appending at the beginning of the file instead of end of the file. Fixing the test is hiding the issue.

[In this PR ](https://github.com/kedro-org/kedro-plugins/pull/538)users notice that the `pyarrow` version is affecting the DeltaTableDataset.

SajidAlamQB:
> Thanks for digging, it is interesting that you find the issue is from `deltalake`.
> 
> Is it `deltalake`, `pyarrow` or actually the combination of them? I'd argue the change should be either pinning the upper bound or update the DeltaTableDataset, because now the `append` function is appending at the beginning of the file instead of end of the file. Fixing the test is hiding the issue.
> 
> [In this PR ](https://github.com/kedro-org/kedro-plugins/pull/538)users notice that the `pyarrow` version is affecting the DeltaTableDataset.

Hey @noklam I am pretty sure the issue is due to a specific change in the deltalake between versions `0.15.1` and `0.15.2`. I confirmed it by swapping the `_internal.abi3.so` (shared object file for calling rust code from python) from `0.15.2` back to the one from `0.15.1` which fixes the the test failures. 

I am not apposed to pinning deltalake to version `0.15.1` as a temporary measure to unblock some PRs. I think we'll have to look deeper into the rust code on deltalake to understand the changes and see if we can adapt our `DeltaTableDataset`.

As for the pyarrow PR from what I can see in the pip freeze their environment also has deltalake `0.15.2` switching to `0.15.1` will resolve the issue.

AhdraMeraliQB (approved):
Thanks for digging into this!

I'm happy to adapt the test to the changes; the data is confirmed to be the same (and in the limited data manipulation I've done there's always been some sort of ordering after fetching the data - I don't think adding this behaviour to the test deviates from its expected use)

> As for the pyarrow PR from what I can see in the pip freeze their environment also has deltalake 0.15.2 switching to 0.15.1 will resolve the issue.

@SajidAlamQB just a quick confirm: pinning is an alternative, but this change should also unblock the PR, right?

astrojuanlu:
> Is it deltalake, pyarrow or actually the combination of them? I'd argue the change should be either pinning the upper bound or update the DeltaTableDataset, because now the append function is appending at the beginning of the file instead of end of the file. Fixing the test is hiding the issue.

> I am not apposed to pinning deltalake to version 0.15.1 as a temporary measure to unblock some PRs. I think we'll have to look deeper into the rust code on deltalake to understand the changes and see if we can adapt our DeltaTableDataset.

Should we then pin deltalake to <0.15.2 or even <0.15 for a short-term fix (instead of fixing the test), and buy us some time to look at this more peacefully?

SajidAlamQB:
> Should we then pin deltalake to <0.15.2 or even <0.15 for a short-term fix (instead of fixing the test), and buy us some time to look at this more peacefully?

Reverted changes to the tests and pinned deltalake to <0.15.2.

noklam (approved):

noklam:
Approved, can you add a comment for the pin?

PR #552, opened 2024-02-09
fix: kedro-telemetry masking

DimedS:
## Description
This PR modifies the algorithm for masking CLI inputs:

1. It retains the initial segment of the command, permitting only commands from the Kedro CLI, up until the point where parameter sections start.
2. It masks all inputs except for the permitted parameter keywords specific to that Kedro CLI command.

Additionally, the sequence of operations in KedroTelemetryCLIHooks has been altered: we now first check for the user's consent. If it has not been provided, we take no further action. Previously, we would parse the Kedro CLI structure first, which consumed time unnecessarily.

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/masking.py
+++ b/kedro-telemetry/kedro_telemetry/masking.py
@@ -44,6 +44,10 @@ def _recurse_cli(
                 io_dict[element_name],
                 get_help,
             )
+        nested_parameter_list = [option.opts for option in cli_element.get_params(ctx)]
+        for item in (item for sublist in nested_parameter_list for item in sublist):
+            if item not in io_dict[element_name]:
+                io_dict[element_name][item] = None
 
     elif isinstance(cli_element, click.Command):
         if get_help:  # gets formatted CLI help incl params for printing
@@ -89,7 +93,7 @@ def _mask_kedro_cli(cli_struct: Dict[str, Any], command_args: List[str]) -> List
         current_CLI = current_CLI[command_args[arg_index]]
         arg_index += 1
 
-    # Mask everything except recognized parameter keywords
+    # Mask everything except parameter keywords
     for arg in command_args[arg_index:]:
         if arg.startswith("-"):
             if "=" in arg:
--- a/kedro-telemetry/kedro_telemetry/masking.py
+++ b/kedro-telemetry/kedro_telemetry/masking.py
@@ -72,39 +72,39 @@ def _get_cli_structure(
     return output
 
 
-def _mask_unrecognized_arg(arg: str, vocabulary: Set[str]) -> str:
-    return arg if arg in vocabulary else MASK
-
-
 def _mask_kedro_cli(cli_struct: Dict[str, Any], command_args: List[str]) -> List[str]:
     """Takes a dynamic vocabulary (based on `KedroCLI`) and returns
     a masked CLI input"""
     output = []
-    vocabulary = _get_vocabulary(cli_struct)
-    mask_next = False
-    for arg in command_args:
+
+    # Preserve the initial part of the command until parameters sections begin
+    arg_index = 0
+    current_CLI = cli_struct.get("kedro", {})
+    while (
+        arg_index < len(command_args)
+        and not command_args[arg_index].startswith("-")
+        and command_args[arg_index] in current_CLI
+    ):
+        output.append(command_args[arg_index])
+        current_CLI = current_CLI[command_args[arg_index]]
+        arg_index += 1
+
+    # Mask everything except recognized parameter keywords
+    for arg in command_args[arg_index:]:
         if arg.startswith("-"):
             if "=" in arg:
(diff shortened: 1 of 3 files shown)


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -78,6 +71,14 @@ def before_command_run(
...
                 return
 
+            # get KedroCLI and its structure from actual project root
+            cli = KedroCLI(project_path=Path.cwd())
+            cli_struct = _get_cli_structure(cli_obj=cli, get_help=False)
+            masked_command_args = _mask_kedro_cli(
+                cli_struct=cli_struct, command_args=command_args
+            )
+            main_command = masked_command_args[0] if masked_command_args else "kedro"
+
```
astrojuanlu, line 81: +1 on moving this here

kedro-telemetry/kedro_telemetry/masking.py
```diff
@@ -72,22 +72,30 @@ def _get_cli_structure(
...
 
+def _mask_unrecognized_arg(arg: str, vocabulary: Set[str]) -> str:
+    return arg if arg in vocabulary else MASK
+
+
 def _mask_kedro_cli(cli_struct: Dict[str, Any], command_args: List[str]) -> List[str]:
     """Takes a dynamic vocabulary (based on `KedroCLI`) and returns
     a masked CLI input"""
     output = []
     vocabulary = _get_vocabulary(cli_struct)
```
astrojuanlu, line 83: Potential opportunity for refactor: looks like `cli_struct` is only used to extract the vocabulary. Maybe `_mask_kedro_cli` should take `vocabulary` as a parameter instead?

kedro-telemetry/kedro_telemetry/masking.py
```diff
@@ -72,22 +72,30 @@ def _get_cli_structure(
...
-                if arg_part in vocabulary:
-                    output.append(arg_part)
-                elif arg_part:
-                    output.append(MASK)
-        elif arg in vocabulary:
-            output.append(arg)
-        elif arg:
+            if "=" in arg:
+                arg_left = arg.split("=")[0]
+                output.append(_mask_unrecognized_arg(arg_left, vocabulary))
```
astrojuanlu, line 89: Q: Do we still want to use the vocabulary approach? Or instead "mask any value given to a parameter"? In other words, should we be looking only at the structure and the order of the CLI to mask instead of the vocabulary?

astrojuanlu (commented):
Left a few comments

kedro-telemetry/kedro_telemetry/masking.py
DimedS: I've considered this as well. I retained the previous algorithm for safety reasons—to be 100% sure that only Kedro CLI keywords are not masked. However, scanning the CLI each time is time-consuming. I will be happy simplify the process by adopting the approach of masking words following parameter keywords only.

kedro-telemetry/kedro_telemetry/masking.py
DimedS: I agree, that's right, but let's remove that part in general according to your next comment.

ankatiyar:
I was testing these changes, while this works well with CLI options eg. `kedro run --pipeline kedro` will be properly masked but it doesn't mask commands like `kedro pipeline create pipeline`. Might be changing the approach to check the command against `cli_struct` instead of `vocabulary` (which is just a set) would work?

kedro-telemetry/kedro_telemetry/masking.py
```diff
@@ -76,27 +83,35 @@ def _mask_kedro_cli(cli_struct: Dict[str, Any], command_args: List[str]) -> List
     """Takes a dynamic vocabulary (based on `KedroCLI`) and returns
     a masked CLI input"""
     output = []
-    vocabulary = _get_vocabulary(cli_struct)
-    for arg in command_args:
+
+    # Preserve the initial part of the command until parameters sections begin
+    arg_index = 0
+    current_CLI = cli_struct.get("kedro", {})
```
ankatiyar, line 89: My IDE is saying the variable name should be lowercase, maybe rename this to `current_cli`? 😆

ankatiyar (approved):
Tested it manually, works great! Thanks @DimedS, will you also add this to the release notes?

astrojuanlu:
Could you share what's the best way to test this? We're doing quite deep introspection of the Click object and I'm a bit lost (the fact that we need such complex logic doesn't speak positively about Click to be honest...)

ankatiyar:
> Could you share what's the best way to test this? We're doing quite deep introspection of the Click object and I'm a bit lost (the fact that we need such complex logic doesn't speak positively about Click to be honest...)

I tested with a spaceflights project and comment out `_send_heap_event()` lines just print the `cli_properties`

astrojuanlu (approved):
Tested, and it works!

PR #557, opened 2024-02-15
fix: telemetry data and add example_pipeline

DimedS:
## Description

This PR addresses the following issues: 
1. It removes parameters from dataset counts, 
2. ensures proper collection of tools by converting them to strings,
3. includes an example_pipeline in the data collection process.

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -159,11 +159,16 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
         with open(pyproject_path) as file:
             pyproject_data = toml.load(file)
 
-        if build_config := pyproject_data.get("tool", {}).get("kedro", {}):
-            if tools := build_config.get("tools"):
-                properties["tools"] = ", ".join(tools)
-            if example_pipeline := build_config.get("example_pipeline"):
-                properties["example_pipeline"] = example_pipeline
+        if "tool" in pyproject_data and "kedro" in pyproject_data["tool"]:
+            if "tools" in pyproject_data["tool"]["kedro"]:
+                # convert list of tools to comma-separated string
+                properties["tools"] = ", ".join(
+                    pyproject_data["tool"]["kedro"]["tools"]
+                )
+            if "example_pipeline" in pyproject_data["tool"]["kedro"]:
+                properties["example_pipeline"] = pyproject_data["tool"]["kedro"][
+                    "example_pipeline"
+                ]
 
     return properties
 
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -159,10 +159,10 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
         with open(pyproject_path) as file:
             pyproject_data = toml.load(file)
 
-        if (build_config := pyproject_data.get("tool", {}).get("kedro", {})):
-            if (tools := build_config.get("tools")):
+        if build_config := pyproject_data.get("tool", {}).get("kedro", {}):
+            if tools := build_config.get("tools"):
                 properties["tools"] = ", ".join(tools)
-            if (example_pipeline := build_config.get("example_pipeline")):
+            if example_pipeline := build_config.get("example_pipeline"):
                 properties["example_pipeline"] = example_pipeline
 
     return properties
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -159,8 +159,16 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
         with open(pyproject_path) as file:
             pyproject_data = toml.load(file)
 
-        if "tools" in pyproject_data["tool"]["kedro"]:
-            properties["tools"] = pyproject_data["tool"]["kedro"]["tools"]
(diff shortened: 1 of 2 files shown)


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -159,8 +159,16 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
...
+        if "tool" in pyproject_data and "kedro" in pyproject_data["tool"]:
+            if "tools" in pyproject_data["tool"]["kedro"]:
+                # convert list of tools to comma-separated string
+                properties["tools"] = ", ".join(
+                    pyproject_data["tool"]["kedro"]["tools"]
+                )
+            if "example_pipeline" in pyproject_data["tool"]["kedro"]:
+                properties["example_pipeline"] = pyproject_data["tool"]["kedro"][
+                    "example_pipeline"
+                ]
```
astrojuanlu, line 171:
```suggestion
        if (build_config := pyproject_data.get("tool", {}).get("kedro", {}):
            if (tools := build_config.get("tools")):
                properties["tools"] = ", ".join(tools)
            if (example_pipeline := build_config.get("example_pipeline")):
                properties["example_pipeline"] = example_pipeline
```

? (didn't test the code, I believe it should work)

DimedS: Thank you, @astrojuanlu , it looks even better now! It seems like there was one parenthesis missing in the first line. I'll correct that and commit the changes.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -185,8 +193,10 @@ def _format_project_statistics_data(
 ):
     """Add project statistics to send to Heap."""
     project_statistics_properties = properties.copy()
-    project_statistics_properties["number_of_datasets"] = len(
-        catalog.datasets.__dict__.keys()
+    project_statistics_properties["number_of_datasets"] = sum(
+        1
+        for c in catalog.list()
+        if not c.startswith("parameters") and not c.startswith("params:")
```
noklam, line 199: `catalog.list("^(?!params\:)\w+$")` will gives you the same thing I think. but I think this is also fine.

My question is should we also count the dataset generated from pattern? This will underestimate the datasets a lot if it's generated by patterns.

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: @astrojuanlu, I rolled back that change because I encountered an error in mocker related to that syntax and wasn't sure how to resolve the issue with the mocker. I suggest we leave it as it is for now. Error attached:

Screenshot 2024-02-20 at 13 01 43

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: @noklam , thank you for highlighting this issue. I've verified that our telemetry currently does not account for datasets created via dataset factories. 

While attempting to address this by scanning node.inputs and node.outputs, I encountered a challenge with intermediate datasets such as 'X_test', 'y_train', and 'y_test', which introduced noise. I also explored using catalog._datasets to identify Dataset types, but this approach doesn't reveal datasets instantiated by factories. 

Given these challenges, I propose we maintain the current state in this PR and open a separate issue dedicated to accurately counting datasets generated from dataset factories in telemetry.

astrojuanlu:
I'm thinking: if the hook is calling `_get_project_properties` and that function opens the `pyproject.toml`...

...does this work with packaged projects? 🤔 Meaning: the result of `kedro package`, then installing the resulting artifact, and then doing `python -m <package_name>`.

@DimedS do you mind trying it out if you have the time? No need to fix it in this PR, but still relevant I'd say.

kedro-telemetry/kedro_telemetry/plugin.py
astrojuanlu: I agree with @DimedS 👍🏽 this is relevant for https://github.com/kedro-org/kedro/issues/1978 which is also in progress.

astrojuanlu:
Also any chance to add a regression test for the dataset counts?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -159,8 +159,16 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
         with open(pyproject_path) as file:
             pyproject_data = toml.load(file)
 
-        if "tools" in pyproject_data["tool"]["kedro"]:
-            properties["tools"] = pyproject_data["tool"]["kedro"]["tools"]
+        if "tool" in pyproject_data and "kedro" in pyproject_data["tool"]:
+            if "tools" in pyproject_data["tool"]["kedro"]:
+                # convert list of tools to comma-separated string
+                properties["tools"] = ", ".join(
```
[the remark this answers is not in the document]
DimedS: Currently, we're sending a list, but Heap isn't recognizing it. Therefore, this PR casting the list to a string. Why do you believe viewing the list in Heap would be preferable?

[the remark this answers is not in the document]
ankatiyar: Ah I see, that's alright. I assumed that processing it as a list would be easier for analytics on Heap.

noklam (approved):

ankatiyar (approved):

DimedS:
> Also any chance to add a regression test for the dataset counts?

Regression test implemented, and follow-up issue regarding dataset factories and pyproject.toml in packaged projects created: #566

Issue #535, opened 2023-06-23
How to maintain external datasets contributions


noklam:
### Why this is raised?
With more incoming datasets PR, it become harder to maintain all the datasets. Particularly for the exotic datasets, we don't have the setup for every possible environment (e.g. snowflake/databricks). This create challenge for maintaining all the datasets since we don't have the re

This also lead to the question "Does every datasets belongs to `kedro-datasets`?

The answer is no, since there are few popular datasets maintained separately in `kedro-mlflow` as well.

### Possible Action
* CSVDataSet is more robust than say ManagedTableDataSet, can we signal this better through our docs? We did something  similar to [Deployment docs](https://github.com/kedro-org/kedro/issues/2435)

### More Discussion
How to we want to maintain the contributions? How do we draw the line that something should be a separate plugins or going into `kedro-datasets` Cc @astrojuanlu 

Idea raised during retro:
1. datasets could be maintained as a separate plugins. i.e. `kedro-mlflow` has its own datasets.

datajoely:
I do keep wondering if we could have a Low-code dataset contribution workflow on the website that allowed us to accept contributions and manage the test suite for users.

astrojuanlu:
A user literally _ran out of disk space_ when trying to install kedro-datasets test dependencies while troubleshooting a pip conflict https://github.com/kedro-org/kedro-plugins/issues/597#issuecomment-1981138414

lrcouto:
> A user literally _ran out of disk space_ when trying to install kedro-datasets test dependencies while troubleshooting a pip conflict [#597 (comment)](https://github.com/kedro-org/kedro-plugins/issues/597#issuecomment-1981138414)

That happened to me this week while running tests to figure out the issues with the kedro-datasets dependencies 😬

astrojuanlu:
I think the main concerns have been addressed with the experimental datasets

Issue #401, opened 2023-10-20
Plugins in this monorepo are hard to find from search engines


astrojuanlu:
## Description
Searching for the plugins living in this monorepo yields old or unrelated links. For example, `kedro-airflow`.

_Note: This is not the case anymore but still an issue, see below_

Ecosia:
![image](https://github.com/kedro-org/kedro-plugins/assets/316517/8b57dc31-0809-42f8-91c8-6d8aa0c9a5f5)

Google:
![image](https://github.com/kedro-org/kedro-plugins/assets/316517/ca73003d-cb4f-4acb-b93e-c7315f5a6076)

astrojuanlu:
This doesn't happen anymore!

astrojuanlu:
I was a bit too fast. This is still an issue, although not so bad as 1 year ago.

Some of _our_ docs can be located in the first results, but not the source code.

Connected with: https://github.com/kedro-org/kedro/issues/3741

astrojuanlu:
The issue has definitely improved, closing

PR #560, opened 2024-02-21
feat(datasets): add dataset to load/save with Ibis

deepyaman:
## Description

Officially add Ibis `TableDataset` introduced in https://kedro.org/blog/building-scalable-data-pipelines-with-kedro-and-ibis.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -14,11 +14,52 @@ if TYPE_CHECKING:
 class TableDataset(AbstractDataset[ir.Table, ir.Table]):
     """``TableDataset`` loads/saves data from/to Ibis table expressions.
 
-    Example:
+    Example usage for the
+    `YAML API <https://kedro.readthedocs.io/en/stable/data/\
+    data_catalog_yaml_examples.html>`_:
+
+    .. code-block:: yaml
+
+        cars:
+          type: ibis.TableDataset
+          filepath: data/01_raw/company/cars.csv
+          file_format: csv
+          table_name: cars
+          connection:
+            backend: duckdb
+            database: company.db
+          load_args:
+            sep: ","
+            nullstr: "#NA"
+          save_args:
+            materialized: table
+
+        motorbikes:
+          type: ibis.TableDataset
+          table_name: motorbikes
+          connection:
+            backend: duckdb
+            database: company.db
+
+    Example usage for the
+    `Python API <https://kedro.readthedocs.io/en/stable/data/\
+    advanced_data_catalog_usage.html>`_:
 
     .. code-block:: pycon
 
+        >>> import ibis
         >>> from kedro_datasets.ibis import TableDataset
+        >>>
+        >>> data = ibis.memtable({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
+        >>>
+        >>> dataset = TableDataset(
+        ...     table_name="test",
+        ...     connection={"backend": "duckdb", "database": tmp_path / "file.db"},
+        ...     save_args={"materialized": "table"},
+        ... )
+        >>> dataset.save(data)
+        >>> reloaded = dataset.load()
+        >>> assert data.execute().equals(reloaded.execute())
 
     """
 
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -145,7 +145,7 @@ class TableDataset(AbstractDataset[ir.Table, ir.Table]):
             if isinstance(value, dict):
                 return tuple((k, hashable(v)) for k, v in sorted(value.items()))
             if isinstance(value, list):
-                return tuple(hashable(x) for x in value)  # pragma: no cover
+                return tuple(hashable(x) for x in value)
             return value
 
         cls = type(self)
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -1,7 +1,6 @@
 """Provide data loading and saving functionality for Ibis's backends."""
 from __future__ import annotations
 
-import json
 from copy import deepcopy
 from typing import TYPE_CHECKING, Any, ClassVar
 
@@ -143,7 +142,7 @@ class TableDataset(AbstractDataset[ir.Table, ir.Table]):
     @property
     def connection(self) -> BaseBackend:
         cls = type(self)
-        key = json.dumps(self._connection_config, sort_keys=True)
+        key = tuple(sorted(self._connection_config.items()))
         if key not in cls._connections:
             import ibis
 
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -1,0 +1,124 @@
+"""Provide data loading and saving functionality for Ibis's backends."""
+from __future__ import annotations
+
+from copy import deepcopy
+from typing import TYPE_CHECKING, Any, ClassVar
+
+import ibis.expr.types as ir
+from kedro.io import AbstractDataset, DatasetError
+
+if TYPE_CHECKING:
+    from ibis import BaseBackend
+
+
+class TableDataset(AbstractDataset[ir.Table, ir.Table]):
+    """``TableDataset`` loads/saves data from/to Ibis table expressions.
+
+    Example:
+
+    .. code-block:: pycon
+
+        >>> from kedro_datasets.ibis import TableDataset
+
+    """
+
+    DEFAULT_LOAD_ARGS: ClassVar[dict[str, Any]] = {}
+    DEFAULT_SAVE_ARGS: ClassVar[dict[str, Any]] = {
+        "materialized": "view",
+        "overwrite": True,
+    }
+
+    _connections: ClassVar[dict[tuple[tuple[str, str]], BaseBackend]] = {}
+
+    def __init__(  # noqa: PLR0913
+        self,
+        *,
+        filepath: str | None = None,
+        file_format: str | None = None,
+        table_name: str | None = None,
+        connection: dict[str, Any] | None = None,
+        load_args: dict[str, Any] | None = None,
+        save_args: dict[str, Any] | None = None,
+    ) -> None:
+        """Creates a new ``TableDataset`` pointing to a table (or file).
+
+        Args:
+            filepath: Path to a file to register as a table. Most useful
+                for loading data into your data warehouse (for testing).
+            file_format: Specifies the input file format for `filepath`.
+            table_name: The name of the table or view to read or create.
+            connection: Configuration for connecting to an Ibis backend.
+            load_args: Additional arguments passed to the Ibis backend's
+                `read_{file_format}` method.
+            save_args: Additional arguments passed to the Ibis backend's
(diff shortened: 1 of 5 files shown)


astrojuanlu:
:fire:

astrojuanlu:
Before I start reviewing: I'm fine pushing this to `kedro-datasets` for now but I want to make it clear that at some point I'd like to seriously discuss the idea of breaking up `kedro-datasets` into domain-specific subpackages https://github.com/kedro-org/kedro-plugins/issues/535#issuecomment-1923539103 given that our monorepo approach is actually obstructing discoverability #401

datajoely:
I would go further @astrojuanlu and say this should be in a `kedro-ibis` package from day 1 - especially since (1) we know it's going to end up there today (2) the dependency conflicts with the rest of the datasets are going to be numerous.

datajoely:
but also thank you for the push this is great!

kedro-datasets/pyproject.toml
```diff
@@ -49,6 +49,28 @@ huggingface-hfdataset = ["datasets", "huggingface_hub"]
 huggingface-hftransformerpipelinedataset = ["transformers"]
 huggingface = ["kedro-datasets[huggingface-hfdataset,huggingface-hftransformerpipelinedataset]"]
 
+ibis-bigquery = ["ibis-framework[bigquery]"]
```
datajoely, line 52: I wonder if we could do a CI script to keep this bit in sync

deepyaman:
> Before I start reviewing: I'm fine pushing this to `kedro-datasets` for now but I want to make it clear that at some point I'd like to seriously discuss the idea of breaking up `kedro-datasets` into domain-specific subpackages [#535 (comment)](https://github.com/kedro-org/kedro-plugins/issues/535#issuecomment-1923539103) given that our monorepo approach is actually obstructing discoverability #401

@astrojuanlu Intuitively, what would domain-specific subpackages look like? `kedro-datasets-spark`, `kedro-datasets-tensorflow`, `kedro-datasets-ibis` or something? (Maybe just `kedro-spark`, `kedro-tensorflow`, `kedro-ibis`.) I can see this making sense, but it's a much broader effort (with it's own complications, like figuring out the extent to which datasets need to be compatible with each other), and I'd be supportive of doing it down the road.

> I would go further @astrojuanlu and say this should be in a `kedro-ibis` package from day 1 - especially since (1) we know it's going to end up there today (2) the dependency conflicts with the rest of the datasets are going to be numerous.

1. Do we know it's going to end up in a separate package? At this point, `ibis.TableDataset` is the only concrete contribution.
2. I don't believe so; `ibis-framework` itself doesn't have such tight dependencies, and then you choose the backend you'd like as an extra. Ibis is pretty good about dependencies—better than the majority of other libraries—in order to be able to resolve across every backend (except Flink 😞 but that's Flink's fault), so I don't see the introduction of this causing resolution issues.

For both @astrojuanlu and @datajoely (and anybody else on the TSC): If we do go down the route of `kedro-ibis` as a separate package, would it fall under `kedro-org`?

datajoely:
I think `kedro-plugins` -> `kedro-ibis` or `kedro-datasets-ibis`

noklam:
I get some weird error when I try to run the test maunally, CI seems to be fine with it.
image

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class TableDataset(AbstractDataset[ir.Table, ir.Table]):
+    """``TableDataset`` loads/saves data from/to Ibis table expressions.
+
+    Example:
+
+    .. code-block:: pycon
+
+        >>> from kedro_datasets.ibis import TableDataset
```
noklam, line 21: Would it be possible to add an local example here with some dummy dataframe / duckdb? I know there is a blog post, but just looking at the docs/docstring I think it's not that easy to get how an user should use this.

merelcht: Ideally we'd have both a `yaml` and `python` example.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _load(self) -> ir.Table:
+        if self._filepath is not None:
+            if self._file_format is None:
+                raise NotImplementedError
+
+            reader = getattr(self.connection, f"read_{self._file_format}")
+            return reader(self._filepath, self._table_name, **self._load_args)
+        else:
+            return self.connection.table(self._table_name)
```
inigohidalgo, line 101: currently when loading tables thru the `mssql` backend I am doing `conn.table(table_name="XXX", schema="A").` 

an easy solution would be to add the load args into the `conn.table` call and specify the schema as a `load_arg`, but imo this is suboptimal, as the write and save schema should usually (in my experience, my assertion could be wrong) be the same

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -0,0 +1,124 @@
...
+        self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+        self._materialized = self._save_args.pop("materialized")
+
+    @property
+    def connection(self) -> BaseBackend:
+        cls = type(self)
+        key = tuple(sorted(self._connection_config.items()))
```
inigohidalgo, line 83: I get an error `unhashable type: 'dict'` when using this connection method

```python
conn_table = TableDataset(
    connection={
        "resource": "mssql://xxx.sql.azuresynapse.net/xxx",
        "query": {"driver": "ODBC Driver 17 for SQL Server"},
    },
    table_name="SIGNAL",
    load_args={"schema": "xxx"},
)
```

kedro-datasets/kedro_datasets/ibis/table_dataset.py
inigohidalgo: @gforsyth not sure if you have any insight here wrt the upcoming changes in ibis around schema and db hierarchy

gforsyth: so starting in 9.0, `schema` will be deprecated as a kwarg.  We will be using the word "database" to refer to a collection of tables and the word "catalog" to refer to a collection of "database".

For specifying only the `database`, you would do:

```
conn.table(table_name="XX", database="A")
```

if the database is located in a catalog (like, say `dbo` or `sys`) you would do either of:

```
conn.table(table_name="XX", database="dbo.A")
```

or

```
conn.table(table_name="XX", database=("dbo", "A"))
```

(`schema` will still work in 9.0, it will just raise a FutureWarning)

kedro-datasets/kedro_datasets/ibis/table_dataset.py
inigohidalgo: Just fyi, when connecting to ibis there are two major "paradigms", either a filebased connection or a db connection. From my limited knowledge I think the configuration arguments are quite different.

For reference, I am currently connecting to mssql dbs like this:

```python
conn_table = TableDataset(
    connection={
        "host": "xxx.sql.azuresynapse.net",
        "database": "xxx",
        "query": {"driver": "ODBC Driver 17 for SQL Server"},
        "backend": "mssql",
    },
    table_name="xxx",
    load_args={"schema": "xxx"},
)
```

inigohidalgo: the documentation here should probably clarify what the different arguments map to in the ibis connection object. 

in the case of a `filepath` being provided, it will map to the backend method `read_{file_format}`, e.g. https://ibis-project.org/backends/duckdb#ibis.backends.duckdb.Backend.read_parquet

otherwise it will get the table from `ibis.{backend}.Backend.table` where the Backend object is obtained through `ibis.{backend}.connect` e.g. https://ibis-project.org/backends/mssql

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -0,0 +1,124 @@
+"""Provide data loading and saving functionality for Ibis's backends."""
+from __future__ import annotations
+
+from copy import deepcopy
+from typing import TYPE_CHECKING, Any, ClassVar
+
+import ibis.expr.types as ir
```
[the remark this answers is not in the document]
deepyaman: This is the standard alias in the Ibis codebase. Users shouldn't need this; it's basically just added for typing (but can't go in a `TYPE_CHECKING` block).

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: Sorry, I forgot to add an example, it seems. 😅

deepyaman:
> I get some weird error when I try to run the test maunally, CI seems to be fine with it. Maybe it's the `ibis` example dataset fetching from gcs.
> 
> image
> Is there a way to make this test run without network? Seems like it is fetching data from bucket, not sure about that. After re-run CI it also failed.

@noklam I've updated it to use our canonical `[[1, 2], [4, 5], [5, 6]]` example. Still using DuckDB, since it's probably worth highlighting DB interactions (including the default Ibis backend).

noklam (approved):

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: @inigohidalgo Good catch! I've done some digging, and opted to use `json.dumps(self._connection_config, sort_keys=True)` instead. I think this should handle the case you brought up.

deepyaman: Ugh, that doesn't work, for non- serialization keys...

deepyaman: OK, fixed and added some test cases!

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Args:
+            filepath: Path to a file to register as a table. Most useful
+                for loading data into your data warehouse (for testing).
+            file_format: Specifies the input file format for `filepath`.
+            table_name: The name of the table or view to read or create.
+            connection: Configuration for connecting to an Ibis backend.
+            load_args: Additional arguments passed to the Ibis backend's
+                `read_{file_format}` method.
+            save_args: Additional arguments passed to the Ibis backend's
```
merelcht, line 115: Any docs we can link to that specify which values a user can supply to `materialized`?

merelcht (approved):
I haven't tried to use the dataset yet, but the code looks all good to me! Thanks for adding the detailed examples and doc-strings, they're great! ⭐

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: I'm not aware of anywhere it's documented centrally (i.e. not on a backend-specific basis) in Ibis. @lostmygithubaccount any chance you know?

lostmygithubaccount: we do not, this is on my TODO list to add API docs for alongside the other `read_*` and `to_*` methods

deepyaman: Sounds good! @merelcht if it's OK then, we'll improve the documentation on the Ibis side, and then whenever that happens, we can make the dataset docs reference that.

astrojuanlu:
So happy to see this merged 🔥 let's make some noise!

PR #565, opened 2024-02-22
feat(telemetry): Add is_ci_env environment to reported project telemetry

### LINKED ISSUE #483: Update `kedro-telemetry` to use the `CI` environment variable to indicate that data comes from a CI environment
### Description
Use the `CI` environment variable to differentiate runs between a CI/CD environment and a real kedro project run for data collection.

https://github.com/kedro-org/kedro/issues/1640#issuecomment-1162000302
> I think it would be good to take this opportunity to distinguish CI runs in telemetry. Maybe can read an environment variable; most CI systems ([GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables), [CircleCI](https://circleci.com/docs/2.0/env-vars#built-in-environment-variables), [Travis CI](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables), etc.) follow the convention of setting CI=true, so I would recommend detecting presence of the environment variable, solving the telemetry prompt automatically, and recording commands as belonging to a CI run--and ideally not requiring any additional configuration or hacks to avoid this telemetry prompt.

Whoever pick up this ticket should double check if `CI` is a standard convention we can use. Alternatively, we can refer to how Dagester do it:
https://github.com/dagster-io/dagster/blob/e31e577b8841e0ff54745e206d621d9bf934a063/python_modules/dagster/dagster/_core/telemetry.py#L92

AhdraMeraliQB:
## Description

Closes #483 

## Development notes

This implementation diverges from the Dagster in several ways:
* `KNOWN_CI_ENV_VAR_KEYS` is implemented as a list, not a set, as lists have better iteration time on average
* We use `os.getenv()` instead of `os.environ.get()` for consistency within our codebase
* `KNOWN_CI_ENV_VAR_KEYS` does not contain keys for tools that set the CI env in our implementation, all keys exempted belong to tools that would otherwise set the CI env var to true

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -149,7 +149,7 @@ class KedroTelemetryProjectHooks:
         )
 
 
-def _check_is_ci_env():
+def _check_is_known_ci_env():
     # Most CI tools will set the CI environment variable to true
     if os.environ.get("CI") is True:
         return True
@@ -171,7 +171,7 @@ def _get_project_properties(hashed_username: str, project_path: str) -> Dict:
         "telemetry_version": TELEMETRY_VERSION,
         "python_version": sys.version,
         "os": sys.platform,
-        "is_ci_env": _check_is_ci_env(),
+        "is_ci_env": _check_is_known_ci_env(),
     }
     pyproject_path = Path(project_path) / "pyproject.toml"
     if pyproject_path.exists():
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -35,7 +35,7 @@ KNOWN_CI_ENV_VAR_KEYS = {
     "GITHUB_ACTION",  # https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
     "BITBUCKET_BUILD_NUMBER",  # https://support.atlassian.com/bitbucket-cloud/docs/variables-and-secrets/
     "JENKINS_URL",  # https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
-    "CODEBUILD_BUILD_ID"  # https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+    "CODEBUILD_BUILD_ID",  # https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
     "CIRCLECI",  # https://circleci.com/docs/variables/#built-in-environment-variables
     "TRAVIS",  # https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
     "BUILDKITE",  # https://buildkite.com/docs/pipelines/environment-variables
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -30,7 +30,7 @@ from kedro_telemetry.masking import _get_cli_structure, _mask_kedro_cli
 HEAP_APPID_PROD = "2388822444"
 HEAP_ENDPOINT = "https://heapanalytics.com/api/track"
 HEAP_HEADERS = {"Content-Type": "application/json"}
-KNOWN_CI_ENV_VAR_KEYS = (
+KNOWN_CI_ENV_VAR_KEYS = {
     "GITLAB_CI",  # https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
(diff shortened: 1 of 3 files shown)


kedro-telemetry/tests/test_plugin.py
```diff
@@ -412,6 +421,21 @@ def test_confirm_consent_yaml_dump_error(self, mocker, fake_metadata, caplog):
         )
         assert msg in caplog.messages[-1]
 
+    @mark.parametrize(
```
AhdraMeraliQB, line 424: The examples for testing CI tools that don't set the CI env var to be true feel a little clunky, any ideas from the reviewers?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -30,6 +30,10 @@
 HEAP_APPID_PROD = "2388822444"
 HEAP_ENDPOINT = "https://heapanalytics.com/api/track"
 HEAP_HEADERS = {"Content-Type": "application/json"}
+KNOWN_CI_ENV_VAR_KEYS = [
```
noklam, line 33: The PR mentioned list is faster to iterate. The speed here doesn't have any impact, but we usually defined these constant values with `tuple`.

see: https://github.com/kedro-org/kedro/blob/941440aa4a5815e8d8157d167ec581524a1f6e20/kedro/io/core.py#L31

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -145,6 +149,19 @@ def after_catalog_created(self, catalog):
         )
 
 
+def _check_is_known_ci_env():
```
noklam, line 152:
```suggestion
def _is_known_ci_env():
```
nit

kedro-telemetry/kedro_telemetry/plugin.py
astrojuanlu: I'm sure this is a negligible micro optimization, but in any case +1 to turning this to a `tuple` 👍🏽

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -30,6 +30,10 @@
 HEAP_APPID_PROD = "2388822444"
 HEAP_ENDPOINT = "https://heapanalytics.com/api/track"
 HEAP_HEADERS = {"Content-Type": "application/json"}
+KNOWN_CI_ENV_VAR_KEYS = [
+    "CODEBUILD_BUILD_ID",  # https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+    "JENKINS_URL",  # https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
+]
```
astrojuanlu, line 36: This will need more variables. Dagster is Apache license, we can copy https://github.com/dagster-io/dagster/blob/e31e577b8841e0ff54745e206d621d9bf934a063/python_modules/dagster/dagster/_core/telemetry.py#L92-L101

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -145,6 +149,19 @@ def after_catalog_created(self, catalog):
         )
 
 
+def _check_is_known_ci_env():
+    # Most CI tools will set the CI environment variable to true
+    if os.getenv("CI") == "true":
```
astrojuanlu, line 154: Are we confident it will be `"true"` and not `"True"` or `1`?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -145,6 +149,19 @@ def after_catalog_created(self, catalog):
...
+        return True
+
+    # Not all CI tools follow this convention, we can check through those that don't
+    for env_var_key in KNOWN_CI_ENV_VAR_KEYS:
+        if env_var_key in os.environ:
+            return True
+
+    return False
```
astrojuanlu, line 162: Maybe

```suggestion
    # Not all CI tools follow this convention, we can check through those that don't
    return not set(KNOWN_CI_ENV_VAR_KEYS).isdisjoint(os.environ)
```

? (Didn't test this code, offering it as a suggestion)

kedro-telemetry/kedro_telemetry/plugin.py
astrojuanlu: (Although see below for a way of using a `set` instead)

kedro-telemetry/kedro_telemetry/plugin.py
ankatiyar: I checked, it's "true" for most places.

astrojuanlu (approved):
LGTM!

noklam (approved):

PR #595, opened 2024-03-05
feat(datasets): Extend preview mechanism

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/1622

This PR introduces an update to extend the preview functionality across more kedro-datasets, extending support for different data formats.

List of Datasets we are aiming to support in this PR:

- [x] `ParquetDataset`
- [x] `SQLTableDataset`
- [x] `pandas.JSONDataset`
- [x] `JSONDataset`

## Checklist


noklam:
Is this still a draft or waiting for something?

SajidAlamQB:
> @SajidAlamQB Is this still a draft or waiting for something?

Its ready for review, just some coverage tests that are needed.

kedro-datasets/kedro_datasets/json/json_dataset.py
```diff
@@ -160,3 +162,17 @@ def _invalidate_cache(self) -> None:
...
+
+        Returns:
+            JSON: The JSON data for preview purposes.
+        """
+        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+
+        with self._fs.open(load_path, mode="r", **self._fs_open_args_load) as fs_file:
+            data = json.load(fs_file)
+
+        return json.dumps(data)
```
noklam, line 178: Is it possible to reuse `_load` as much as possible?
https://github.com/kedro-org/kedro-plugins/blob/9acc3a571b04ab3bab2e63e32c6e90d5bfa32ace/kedro-datasets/kedro_datasets/pandas/csv_dataset.py#L200-L215

kedro-datasets/kedro_datasets/pandas/parquet_dataset.py
```diff
@@ -214,3 +216,19 @@ def _invalidate_cache(self) -> None:
...
+
+        Args:
+            nrows: The number of rows to include in the preview. Defaults to 5.
+
+        Returns:
+            dict: A dictionary containing the data in a split format.
+        """
+        load_path = str(self._get_load_path())
+        data_preview = pd.read_parquet(load_path, **self._load_args)
+        preview_data = data_preview.head(nrows).to_dict(orient="split")
```
noklam, line 232: Does this read the full file and then filter? This will cause significant performance issue, there should be native parquet options that allows you to filter from source.

kedro-datasets/kedro_datasets/pandas/sql_dataset.py
```diff
@@ -90,6 +92,12 @@ def _get_sql_alchemy_missing_error() -> DatasetError:
     )
 
 
+def sanitise_table_name(table_name):
```
noklam, line 95: should this be private?

noklam (commented):
Few comments:
- I would try to stick with `_load` as much as possible to avoid re-implementation and edge case that run into difference load data. There may be some special case if `load` and `preview` expect different data, we discussed this in tech design before but that's another topic.
- Performance - the `preview` should be a lightweight solution and should never read the full data.

kedro-datasets/kedro_datasets/pandas/json_dataset.py
```diff
@@ -188,3 +190,34 @@ def _invalidate_cache(self) -> None:
...
+        # Create a copy
+        load_args = self._load_args.copy()
+
+        lines = load_args.pop("lines", False)
+
+        if lines:
+            data_preview = pd.read_json(load_path, lines=True, nrows=nrows, **load_args)
+        else:
+            full_data = pd.read_json(load_path, **load_args)
+            # if its nested try normalise
```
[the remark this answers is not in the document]
SajidAlamQB: The `lines` argument treats each line in the file as a separate JSON object. So for the `preview` method, using the lines limits the number of rows we read from the file so we can possibly avoid loading large data into memory. Its not needed in `_load` since we assume user needs full data.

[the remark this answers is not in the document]
noklam: I think this is unnecessary. You are reusing the `lines` argument from load args, which means the original dataset will use this via `return pd.read_json(load_path, **self._load_args)` anyway.

If this is a preview specific argument, shouldn't it go to `preview_args`?

kedro-datasets/kedro_datasets/pandas/parquet_dataset.py
SajidAlamQB: Just going through [pandas.read_parquet](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_parquet.html) I don't see an option to read in a specific number of rows and the filter parameter seems to be filtering the data based on column values and not the number of rows which isn't super helpful as this is a generic method.

Maybe we should go with a lower level approach and import `pyarrow`?

kedro-datasets/kedro_datasets/pandas/json_dataset.py
```diff
@@ -188,3 +190,27 @@ def _invalidate_cache(self) -> None:
...
+        Generate a preview of the dataset with a specified number of rows,
+        including handling for both flat and nested JSON structures.
+
+        Args:
+            nrows: Number of rows to include in the preview. Defaults to 5.
+
+        Returns:
+            dict: A dictionary in a split format for preview, if possible.
+        """
+        data = self._load()
```
rashidakanchwala, line 205: hi, can we pass nrows here itself so that we don't have to load all the data?

SajidAlamQB: The `nrows` param only works when `lines=True` through `pd.read_json()`. So this won't work for regular json without `lines=True`, we would need to load it all and manually get the rows.

kedro-datasets/tests/json/test_json_dataset.py
```diff
@@ -197,3 +199,22 @@ def test_versioning_existing_dataset(
...
+        preview_data = json_dataset.preview()
+
+        # Load the data directly for comparison
+        with json_dataset._fs.open(json_dataset._get_load_path(), mode="r") as fs_file:
+            full_data = json.load(fs_file)
+
+        expected_data = json.dumps(full_data)
+
+        assert (
+            preview_data == expected_data
```
rashidakanchwala, line 215: should we also have a success test case where preview_data == expected data?

rashidakanchwala (approved):
LGTM. thanks @SajidAlamQB <3

kedro-datasets/kedro_datasets/pandas/sql_dataset.py
```diff
@@ -275,6 +283,25 @@ def _exists(self) -> bool:
...
+        Args:
+            nrows: The number of rows to include in the preview. Defaults to 5.
+
+        Returns:
+            dict: A dictionary containing the data in a split format.
+        """
+
+        table_name = _sanitise_table_name(self._load_args["table_name"])
+
+        sql_query = f"SELECT * FROM {table_name} LIMIT {nrows}"  # nosec
```
noklam, line 299: This allow SQL injection and thus the security error? Do we have better alternatives?

noklam, line 299: is `LIMIT` universal to all SQL dataset? I remember SQL SERVER use `TOP` instead.

https://stackoverflow.com/questions/1891789/sql-select-first-10-rows-only

kedro-datasets/kedro_datasets/pandas/parquet_dataset.py
```diff
@@ -214,3 +216,24 @@ def _invalidate_cache(self) -> None:
...
+
+        Returns:
+            dict: A dictionary containing the data in a split format.
+        """
+        import pyarrow.parquet as pq
+
+        load_path = str(self._get_load_path())
+
+        table = pq.read_table(
+            load_path, columns=self._load_args.get("columns"), use_threads=True
```
noklam, line 235: is column the only applicable arg?

kedro-datasets/kedro_datasets/pandas/sql_dataset.py
rashidakanchwala: we did it like this in Kedro-viz - https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/repositories/runs.py#L51

rashidakanchwala: we could also do without creating a model - https://stackoverflow.com/questions/49005812/python-sqlalchemy-access-huge-db-data-without-creating-models

kedro-datasets/kedro_datasets/pandas/parquet_dataset.py
SajidAlamQB: For our current use case in Kedro-Viz, focusing on the columns args with the number of rows seems to enough.

kedro-datasets/kedro_datasets/pandas/sql_dataset.py
SajidAlamQB: Addressed the issue by using SQLAlchemy library instead.

noklam (approved):
Thanks for addressing all the comments, this is a great addition to the previews feature:sparkles:

Issue #333, opened 2023-09-08
User identity for telemetry events is not unique


astrojuanlu:
## Description
kedro-telemetry uses the hashed username for the identity of the Heap event:

https://github.com/kedro-org/kedro-plugins/blob/7d092dbdb9344af58435ed6a696331ecb80ae553/kedro-telemetry/kedro_telemetry/plugin.py#L84-L94

https://github.com/kedro-org/kedro-plugins/blob/7d092dbdb9344af58435ed6a696331ecb80ae553/kedro-telemetry/kedro_telemetry/plugin.py#L41-L44

However, this is not enough to disambiguate users. An extreme example is that most people launching Kedro in a container will have the username `root`, hence all their events might be collected under the same identity!

```
❯ docker run -it --rm --platform linux/amd64 python:3.11-slim bash
root@6ac694d81117:/# pip install kedro-telemetry
...
>>> from kedro_telemetry.plugin import _get_hashed_username
>>> _get_hashed_username()
'99adc231b045331e514a516b4b7680f588e3823213abe901738bc3ad67b2f6fcb3c64efb93d18002588d3ccc1a49efbae1ce20cb43df36b38651f11fa75678e8'
>>> 
root@6ac694d81117:/# whoami
root
```

## Context
If we are interested in number of users, we might be incurring in a dramatic under-reporting. 

## Possible solution
For example, Orchest used to generate a `TELEMETRY_UUID` and store it in a centralized location:

https://github.com/orchest/orchest/blob/560103a5c7dbdf615a2e0df435076f0ee78b6a1d/services/orchest-api/app/app/utils.py#L431-L432

astrojuanlu:
I really like Pants anonymous telemetry collection system https://www.pantsbuild.org/docs/anonymous-telemetry

astrojuanlu:
From https://github.com/kedro-org/kedro-plugins/issues/510#issuecomment-1885237973, we see that GE, DVC, Evidently, Reflex do track unique users, so we can draw some inspiration from it.

Final file naming still TBC. Since this would be a kedro-telemetry specific config (and not kedro), we could go for something like `~/.config/kedro-telemetry.toml` or similar.

astrojuanlu:
Let's make this

```
~/.config/kedro/telemetry.conf
```

equivalent home path using https://pypi.org/project/appdirs/. Format using `ConfigParser` https://docs.python.org/3/library/configparser.html 

Mechanics as follows:

[...]

PR #596, opened 2024-03-05
feat(telemetry): Unique User IDs in kedro-telemetry - merge only for kedro-telemetry release 0.4.0

DimedS:
## Description
This PR introduces a unique identifier (UUID) for each user of the kedro-telemetry.

- Instead of relying on hashed usernames, we now generate a unique UUID for each user. This UUID is generated using uuid.uuid4() function.
- Generated UUID is stored in a user-specific configuration file, leveraging the appdirs library to determine the appropriate OS-specific path for this file (~/Library/Application Support/kedro/telemetry.conf on MAC)
- If a UUID does not already exist, the function generates a new one, stores it, and then uses this UUID in subsequent telemetry events.

## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,4 +1,5 @@
 # Upcoming release
+* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
 
 # Release 0.3.2
 * Updated plugin to share if a project is being run in a ci environment.
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,4 +1,4 @@
-# Upcoming release
+# Upcoming release 0.4.0
 * Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
 
 # Release 0.3.2
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,5 +1,5 @@
 # Upcoming release
-* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
+* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`. (On release 0.4)
 
 # Release 0.3.2
 * Updated plugin to share if a project is being run in a ci environment.
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,5 +1,5 @@
 # Upcoming release
-* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`. (On release 0.4)
+* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
 
 # Release 0.3.2
 * Updated plugin to share if a project is being run in a ci environment.
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -56,31 +56,35 @@ def _get_or_create_uuid():
     """
     Reads a UUID from a configuration file or generates and saves a new one if not present.
     """
+    try:
+        config_path = user_config_dir("kedro")
+        full_path = os.path.join(config_path, CONFIG_FILENAME)
+        config = ConfigParser()
 
-    config_path = user_config_dir("kedro")
-    full_path = os.path.join(config_path, CONFIG_FILENAME)
-    config = ConfigParser()
+        if os.path.exists(full_path):
+            config.read(full_path)
 
-    if os.path.exists(full_path):
-        config.read(full_path)
+            if config.has_section("telemetry") and "uuid" in config["telemetry"]:
+                try:
+                    return uuid.UUID(config["telemetry"]["uuid"]).hex
+                except ValueError:
+                    pass  # Invalid UUID found, will generate a new one
 
-        if config.has_section("telemetry") and "uuid" in config["telemetry"]:
-            try:
-                return uuid.UUID(config["telemetry"]["uuid"]).hex
-            except ValueError:
-                pass  # Invalid UUID found, will generate a new one
+        # Generate a new UUID and save it to the config file
+        if not config.has_section("telemetry"):
+            config.add_section("telemetry")
+        new_uuid = uuid.uuid4().hex
+        config.set("telemetry", "uuid", new_uuid)
 
-    # Generate a new UUID and save it to the config file
-    if not config.has_section("telemetry"):
-        config.add_section("telemetry")
-    new_uuid = uuid.uuid4().hex
-    config.set("telemetry", "uuid", new_uuid)
+        os.makedirs(config_path, exist_ok=True)
+        with open(full_path, "w") as configfile:
+            config.write(configfile)
 
-    os.makedirs(config_path, exist_ok=True)
-    with open(full_path, "w") as configfile:
-        config.write(configfile)
+        return new_uuid
 
-    return new_uuid
+    except Exception as e:
+        logging.error(f"Failed to get or create UUID: {e}")
+        return ""
 
 
 class KedroTelemetryCLIHooks:
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -72,7 +72,18 @@ def _get_or_create_uuid():
 
         # Generate a new UUID and save it to the config file
         if not config.has_section("telemetry"):
-            config.add_section("telemetry")
+            new_uuid = _generate_new_uuid(config, config_path, full_path)
+
+        return new_uuid
+
+    except Exception as e:
+        logging.error(f"Failed to retrieve UUID: {e}")
+        return ""
+
+
+def _generate_new_uuid(config, config_path, full_path):
+    try:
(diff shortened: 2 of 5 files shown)


lrcouto (commented):
Did some small changes to the structure of the function to leave only the part that actually handles file operations inside the `try` block.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +52,38 @@ def _hash(string: str) -> str:
...
+        config.set("telemetry", "uuid", new_uuid)
+
+        os.makedirs(config_path, exist_ok=True)
+        with open(full_path, "w") as configfile:
+            config.write(configfile)
+
+        return new_uuid
+
+    except Exception as e:
+        logging.error(f"Failed to get or create UUID: {e}")
```
noklam, line 86: Can this be splitted into two function? doing both read and write in a function is a bit too much and hard to test.

For example, how are you testing for the case when config exist but `telemetry` section does not exist?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -15,6 +16,7 @@
 import requests
 import toml
 import yaml
+from appdirs import user_config_dir
```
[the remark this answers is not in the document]
DimedS: > is this a new dependency?

yes, we added it to the dependencies list

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -41,6 +43,7 @@
     "BUILDKITE",  # https://buildkite.com/docs/pipelines/environment-variables
 }
 TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+CONFIG_FILENAME = "telemetry.conf"
```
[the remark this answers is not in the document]
DimedS: > is it expected to be commited into the repo?

You mean this file?  It will be stored in the user's configuration directory, not in project directory, so looks like it will not be possible to commit it.

kedro-telemetry/RELEASE.md
```diff
@@ -1,4 +1,5 @@
 # Upcoming release
+* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
```
astrojuanlu, line 2: This will be `kedro-telemetry` 0.4 by the way (definitely not 0.3.3)

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +52,38 @@ def _hash(string: str) -> str:
     return hashlib.sha512(bytes(string, encoding="utf8")).hexdigest()
 
 
-def _get_hashed_username():
+def _get_or_create_uuid():
+    """
+    Reads a UUID from a configuration file or generates and saves a new one if not present.
+    """
+    config_path = user_config_dir("kedro")
```
astrojuanlu, line 59: ⭐

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +52,38 @@ def _hash(string: str) -> str:
...
+    full_path = os.path.join(config_path, CONFIG_FILENAME)
+    config = ConfigParser()
```
astrojuanlu, line 61: About the use of `ConfigParser`, more opinions needed https://github.com/kedro-org/kedro-plugins/issues/333#issuecomment-1981613714

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +52,38 @@ def _hash(string: str) -> str:
...
-            exc,
-        )
+        if os.path.exists(full_path):
+            config.read(full_path)
+
+            if config.has_section("telemetry") and "uuid" in config["telemetry"]:
+                try:
+                    return uuid.UUID(config["telemetry"]["uuid"]).hex
+                except ValueError:
+                    pass  # Invalid UUID found, will generate a new one
```
astrojuanlu, line 71: > If the file does exist but it's empty, or if the id is not a valid UUID4, or if the file cannot be written, or if anything else fails, then the identity in _send_heap_event must not be set (None or empty string).

The only situation in which we should generate a new UUID is if the file doesn't exist

astrojuanlu (commented):
Left a few comments, thanks @DimedS and @lrcouto !

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: @astrojuanlu , could you please clarify why it's preferable to always send a Null userId when encountering a broken UUID, rather than attempting to fix it by creating and saving a new ID once?

astrojuanlu: Because this leaves a way for the user to overwrite the UUID with something invalid or empty so it's not sent

astrojuanlu: (as a way to opt out)

kedro-telemetry/kedro_telemetry/plugin.py
astrojuanlu: We decided to go for TOML

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +51,48 @@ def _hash(string: str) -> str:
...
-        logger.warning(
-            "Something went wrong with getting the username. Exception: %s",
-            exc,
-        )
+        if os.path.exists(full_path):
+            with open(full_path) as f:
+                config = toml.load(f)
+
+                if "telemetry" in config and "uuid" in config["telemetry"]:
+                    try:
```
DimedS, line 67: looks like we don't need additional try more, because we are returning the same for any error on level upper

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -49,15 +51,45 @@ def _hash(string: str) -> str:
...
+        new_uuid = _generate_new_uuid(full_path)
+
+        return new_uuid
+
+    except Exception as e:
+        logging.error(f"Failed to retrieve UUID: {e}")
+        return ""
+
+
+def _generate_new_uuid(full_path):
```
merelcht, line 79: I just noticed we're not doing any typing in `kedro-telemetry`, but can we maybe start adding that for new functions like this? 🙂

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -90,7 +122,7 @@ def before_command_run(
             main_command = masked_command_args[0] if masked_command_args else "kedro"
 
             logger.debug("You have opted into product usage analytics.")
-            hashed_username = _get_hashed_username()
+            hashed_username = _get_or_create_uuid()
```
merelcht, line 125: Should the variable be renamed to `user_uuid` or something like that now it's not a hashed username anymore?

kedro-telemetry/tests/test_plugin.py
```diff
@@ -226,13 +226,17 @@ def test_before_command_run_empty_args(self, mocker, fake_metadata):
...
+            return_value="hashed_username",
+        )
 
         mocked_heap_call = mocker.patch("kedro_telemetry.plugin._send_heap_event")
         telemetry_hook = KedroTelemetryCLIHooks()
         command_args = []
         telemetry_hook.before_command_run(fake_metadata, command_args)
         expected_properties = {
-            "username": "digested",
+            "username": "hashed_username",
```
merelcht, line 239: NIT: This is the uuid now right? I know it's just test data, but maybe for clarity it would be helpful to specify the actual data so uuid in this case.

merelcht (commented):
I left some minor comments, but otherwise looks good! 👍 Also tested locally and seems like it's all working.

kedro-telemetry/RELEASE.md
```diff
@@ -1,4 +1,5 @@
 # Upcoming release
+* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`. (On release 0.4)
```
ankatiyar, line 2:
```suggestion
* Updated the plugin to generate an unique UUID for each user of `kedro-telemetry`.
```

ankatiyar: The heading will say the release number anyway

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -90,17 +122,17 @@ def before_command_run(
             main_command = masked_command_args[0] if masked_command_args else "kedro"
```
ankatiyar, line 122: I can't comment on the line above since it hasn't been changed but could line 117 be changed to -
`cli = KedroCLI(project_path=metadata.project_path)` 
Following the changes in https://github.com/kedro-org/kedro/pull/3683, I think this will allow us to capture commands called from within subdirectories also and mask them properly

ankatiyar (approved):
Left some comments but looks good to me! Thanks @lrcouto and @DimedS!

merelcht (approved):
LGTM! ⭐

astrojuanlu:
Tested this and tried to break it in various ways, didn't find anything weird ⭐ Thanks @DimedS and @lrcouto!

PR #598, opened 2024-03-07
dtypes support when reading csv files into Polars

grofte:
## Description
Polars needs Polars dtypes, rather than strings, when reading a csv if you want to set the dtypes rather than depend on schema inference. So this takes a string argument like "pl.Int64" in a sequence or mapping and substitutes it with the actual pl.Int64. 

## Development notes
I've tested it in the spaceflights tutorial, both that it works and fails as expected and that it works for all three ways of loading a CSV into Polars. For CSVDataset and Eager I did a `.to_pandas()` in the first line processing `companies` and for Lazy I did `.collect().to_pandas()`.
I got the same number of passes, fails, and errors running `make test_no_spark` before and after adding the new code. 
The code is basically the same for all three classes and could possibly be abstracted away. But I'm not sure about your preferences there.
I haven't written a test because I wasn't sure about requirements and standards. If you give me some hints I will try to write some tests.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/polars/csv_dataset.py
+++ b/kedro-datasets/kedro_datasets/polars/csv_dataset.py
@@ -2,6 +2,7 @@
 filesystem (e.g.: local, S3, GCS). It uses polars to handle the CSV file.
 """
 import logging
+from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from io import BytesIO
 from pathlib import PurePosixPath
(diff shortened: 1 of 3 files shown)


astrojuanlu:
Thanks for the PR @grofte ! Are you aware of this though? https://docs.kedro.org/en/stable/configuration/advanced_configuration.html#how-to-use-resolvers-in-the-omegaconfigloader

grofte:
I will have a look, thank you.

grofte:
Doh

```yaml
my_polars_dataset:
  type: polars.CSVDataset
  filepath: data/01_raw/my_dataset.csv
  load_args:
    dtypes:
      product_age: "${polars:Float64}"
      group_identifier: "${polars:Utf8}"
    try_parse_dates: true
```

noklam:
@grofte Is this PR still relevant? custom resolver should be the default solution for non-primitive type.

astrojuanlu:
Wondering: is there a way to ship a custom resolver with a plugin? So that users do `pip install kedro-datasets[polars]` and have the `$polars` resolver already available.

This is clearly the preferred option but I foresee this might become a common point of friction, maybe we can alleviate it somehow.

grofte:
I'll delete it - definitely not relevant. 
Maybe I'll make a PR on the docstring to add a link to the stuff about the omega config loader. There's already a link in load arguments to Polars so I assume it would be okay to add?

merelcht:
Hi @grofte are you still interested in doing something with this PR or can I close it?

astrojuanlu:
Closing and opened https://github.com/kedro-org/kedro-plugins/issues/687

PR #607, opened 2024-03-13
feat(airflow): add --tags option for node filtering in airflow create

DimedS:
## Description
This PR introduces an option to use the `--tags="ABC"` argument together with the `airflow create` command, enabling the filtering of nodes based on specific tags.

## Checklist

### DIFF
--- a/kedro-airflow/kedro_airflow/plugin.py
+++ b/kedro-airflow/kedro_airflow/plugin.py
@@ -10,7 +10,12 @@ import jinja2
 from click import secho
 from kedro.config import MissingConfigException
 from kedro.framework.cli.project import PARAMS_ARG_HELP
-from kedro.framework.cli.utils import ENV_HELP, KedroCliError, _split_params
+from kedro.framework.cli.utils import (
+    ENV_HELP,
+    KedroCliError,
+    _split_params,
+    split_string,
+)
 from kedro.framework.context import KedroContext
 from kedro.framework.project import pipelines
 from kedro.framework.session import KedroSession
@@ -120,6 +125,7 @@ def _get_pipeline_config(config_airflow: dict, params: dict, pipeline_name: str)
     type=str,
     default="",
     help=TAGS_ARG_HELP,
+    callback=split_string,
 )
 @click.option(
     "--params",
@@ -191,8 +197,7 @@ def create(  # noqa: PLR0913, PLR0912
         dag_filename = dags_folder / dag_name
 
         if tags:
-            tags_list = tags.split(",")
-            pipeline = pipeline.only_nodes_with_tags(*tags_list)  # noqa: PLW2901
+            pipeline = pipeline.only_nodes_with_tags(*tags)  # noqa: PLW2901
 
         # group memory nodes
         if group_in_memory:
--- a/kedro-airflow/kedro_airflow/plugin.py
+++ b/kedro-airflow/kedro_airflow/plugin.py
@@ -24,6 +24,9 @@ If not set, the '__default__' pipeline is used. This argument supports
 passing multiple values using `--pipeline [p1] --pipeline [p2]`.
 Use the `--all` flag to convert all registered pipelines at once."""
 ALL_ARG_HELP = """Convert all registered pipelines at once."""
+TAGS_ARG_HELP = """Tags to be used for filtering pipeline nodes.
+Multiple tags are supported. Use the following format:
+`--tags tag1,tag2`."""
 DEFAULT_RUN_ENV = "local"
 DEFAULT_PIPELINE = "__default__"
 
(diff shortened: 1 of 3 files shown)


DimedS:
@jw-cpnet , thank you for [your solution to adding a --tags filter to the CLI](https://github.com/jw-cpnet/kedro-airflow-fusion/commit/088d926599b63fade88f5fc41672936e50978b01). I've created a PR based on your commit with some minor adjustments. Feel free to review it if you'd like.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -180,6 +190,10 @@ def create(  # noqa: PLR0913
         dag_name += "_dag.py"
         dag_filename = dags_folder / dag_name
 
+        if tags:
+            tags_list = tags.split(",")
+            pipeline = pipeline.only_nodes_with_tags(*tags_list)  # noqa: PLW2901
+
         # group memory nodes
```
noklam, line 197: I suspect this is not consistent with `kedro run` or break with some edge cases. I think we can either import from `kedro` if it's not added dependencies or we can just copy paste the parsing logic.

We do `split_string`, which is roughly:
```python

def split_string(ctx: click.Context, param: Any, value: str) -> list[str]:
    """Split string by comma."""
    return [item.strip() for item in value.split(",") if item.strip()]
```

This logic should fails with whitespace (I don't have a test project, please try it)

noklam (commented):
leave a comment how we handles the new argument.

kedro-airflow/tests/test_plugin.py
```diff
@@ -263,6 +268,31 @@ def test_create_airflow_dag_env_parameter_exists(cli_runner, metadata):
         dag_code = [line.strip() for line in f.read().splitlines()]
     assert expected_airflow_dag in dag_code
 
+    _kedro_remove_env(Path.cwd())
+
+
+def test_create_airflow_dag_tags_parameter_exists(cli_runner, metadata):
+    """Test the `tags` parameter"""
+    dag_name = "hello_world"
+    command = ["airflow", "create", "--tags", "tag0", "--env", "remote"]
```
ElenaKhaustova, line 277: Shall we also add at least one test with two tags?

kedro-airflow/tests/test_plugin.py
```diff
@@ -263,6 +268,31 @@ def test_create_airflow_dag_env_parameter_exists(cli_runner, metadata):
...
+
+    _kedro_create_env(Path.cwd())
+
+    dag_file = Path.cwd() / "airflow_dags" / f"{dag_name}_remote_dag.py"
+    result = cli_runner.invoke(commands, command, obj=metadata)
+
+    assert result.exit_code == 0, (result.exit_code, result.stdout)
+    assert dag_file.exists()
+
+    expected_airflow_dag = 'tasks["node0"] >> tasks["node2"]'
```
ElenaKhaustova, line 287: I would suggest moving test parameters to `@pytest.mark.parametrize` to keep the test cleaner.  However, I've looked through other tests and found they're organised similarly, so please ignore this comment if there's a specific intent to do this.

kedro-airflow/kedro_airflow/plugin.py
DimedS: Yes, that's true. Thank you for pointing that out. It was already a dependency, so I added `split_string` to the imports and the callback.

kedro-airflow/tests/test_plugin.py
DimedS: Great idea, added!

kedro-airflow/tests/test_plugin.py
DimedS: Thank you. I moved them to `parametrize` while adding a new test that involves multiple DAGs. 
I think it's not organized to use numerous @pytest.mark.parametrize with the intention of providing a clear output that includes the test name, which specifies the exact problem, instead of put too many options into parametrize in a single test.

kedro-airflow/tests/test_plugin.py
```diff
@@ -245,6 +246,10 @@ def _kedro_create_env(project_root: Path):
     (project_root / "conf" / "remote").mkdir(parents=True)
 
 
+def _kedro_remove_env(project_root: Path):
```
noklam, line 249: Minor. Should this take env as arg?

noklam (approved):
Approved with a comment.

ankatiyar (approved):
Thanks @DimedS!

kedro-airflow/tests/test_plugin.py
DimedS: That's right, fixed it

ElenaKhaustova (approved):
LGTM, thank you, nice work!

PR #612, opened 2024-03-15
Improved HuggingFace Connectors

lordsoffallen:
## Description

Current HF connectors are too limited so I end up creating my own most of the time.

## Development notes

Similar code that I have been using in my projects. 

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
+++ b/kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
@@ -2,9 +2,14 @@ from __future__ import annotations
 
 from typing import Any
 
-from datasets import load_dataset
+from datasets import load_dataset, Dataset
 from huggingface_hub import HfApi
 from kedro.io import AbstractVersionedDataset
+
+import logging
+
+
+logger = logging.getLogger(__file__)
 
 
 class HFDataset(AbstractVersionedDataset):
@@ -36,20 +41,53 @@ class HFDataset(AbstractVersionedDataset):
 
     """
 
-    def __init__(self, *, dataset_name: str):
+    def __init__(
+        self,
+        *,
+        dataset_name: str,
+        filepath: str = None,
+        credentials: dict[Any] = None,
+        save_to_disk: bool = True,
+        save_to_hub: bool = False,
+    ):
         self.dataset_name = dataset_name
+        self.filepath = filepath
+        self.credentials = credentials
+        self.save_to_disk = save_to_disk
+        self.save_to_hub = save_to_hub
 
-    def _load(self):
-        return load_dataset(self.dataset_name)
+    def _load(self) -> Dataset:
+        try:
+            ds = Dataset.load_from_disk(self.filepath)
+        except FileNotFoundError:
+            ds = load_dataset(self.dataset_name)
+        return ds
 
-    def _save(self):
-        raise NotImplementedError("Not yet implemented")
+    def _save(self, data: Dataset):
+        if self.save_to_disk:
+            logger.info("Saving to local disk.")
+            data.save_to_disk(self._filepath)
+
+        if self.save_to_hub:
+            logger.info("Saving to HuggingFace Hub")
+
+            if isinstance(self.credentials, dict):
+                token = self.credentials.get('write')
+            elif isinstance(self.credentials, str):
+                token = self.credentials
+            else:
+                token = None
+
+            data.push_to_hub(self._dataset_name, token=token)
 
     def _describe(self) -> dict[str, Any]:
         api = HfApi()
         dataset_info = list(api.list_datasets(search=self.dataset_name))[0]
         return {
             "dataset_name": self.dataset_name,
+            "filepath": self.filepath,
+            "save_to_disk": self.save_to_disk,
+            "save_to_hub": self.save_to_hub,
             "dataset_tags": dataset_info.tags,
             "dataset_author": dataset_info.author,
         }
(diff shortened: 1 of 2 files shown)


lordsoffallen:
@merelcht 

I have created this draft PR to start a discussion on some points. Let me know what you people think so I can update it accordingly. Right now, test, documentation is missing cuz I wanted to get the design first.

merelcht:
Thanks @lordsoffallen , I'll bring it up with the team 🙂  In the meantime could you maybe explain a bit more on in what respect the other hugging face datasets are too limited?

lordsoffallen:
> Thanks @lordsoffallen , I'll bring it up with the team 🙂 In the meantime could you maybe explain a bit more in what respect the other hugging face datasets are too limited?

Sure. Current ones don't support local load/save operation, in fact HFDataset don't support save at all so checkpointing datasets at different stages isn't doable as is in kedro. Other one which is transformer oriented is using pipeline only. Having model and tokenizer separately brings more functionality to user than pipeline itself. 

I would say it would be good to think about the whole new pretrained model usecases in kedro. All of the connectors about the data and I am aware models somehow can be used in the same manner but it feels a bit more of hack. Perhaps we'd need something like `kedro_models` to put them all together? Bunch of tools to help manage models? Essentially when we work with huggingface models, a lot of configs are model related so organizing them in a catalog would be more kedro centric.

merelcht:
Thanks for sharing more context @lordsoffallen. I personally haven't worked with hugging face yet, but it does sound like kedro datasets aren't fully matching the models use case. Looping in @astrojuanlu who created the current hugging face dataset to hear his thoughts.

kedro-datasets/kedro_datasets/huggingface/transformer.py
```diff
@@ -0,0 +1,56 @@
...
+from collections import namedtuple
+
+
+TransformerModel = namedtuple("TransformerModel", ["model", "tokenizer"])
+
+
+logger = logging.getLogger(__file__)
+
+
+class HFTransformer(AbstractDataset):
```
astrojuanlu, line 17: What about `HFAutoTransformerDataset`? 

(About `...Dataset`, I know it's not technically a _dataset_ and that there's potential confusion with [Hugging Face Datasets](https://huggingface.co/datasets), but I'm wondering if we should keep consistency here. @merelcht ?

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -36,20 +41,53 @@ class HFDataset(AbstractVersionedDataset):
...
+        except FileNotFoundError:
+            ds = load_dataset(self.dataset_name)
+        return ds
+
+    def _save(self, data: Dataset):
+        if self.save_to_disk:
+            logger.info("Saving to local disk.")
+            data.save_to_disk(self._filepath)
+
+        if self.save_to_hub:
```
astrojuanlu, line 71: I think this branching here is kind of unusual. As an alternative, `kedro-mlflow` provides 2 different datasets for this purpose, `MlflowModelTrackingDataset` and `MlflowModelLocalFileSystemDataset`. Do you think we should do the same here @merelcht ?

astrojuanlu (commented):
Thanks for this contribution @lordsoffallen and sorry it took me so much to get back! I have a couple of questions on the implementation and naming.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
merelcht: Yes looking at this it might be an idea to split this into a dataset that does loading & saving to disk and one that does it remotely.

kedro-datasets/kedro_datasets/huggingface/transformer.py
merelcht: Good point, I'd prefer to keep it consistent even this isn't technically a dataset. `HFAutoTransformerDataset` sounds like a good name.

lordsoffallen:
Hey @merelcht @astrojuanlu,

Sorry I also forgot about this for a while. Now getting back to it, I share my two cents here:

LLM workflows is tricky to use with kedro at the moment. Because I like the framework and i wanted to make it work with kedro. I believe suggestions are to make it fit into the current style but I believe for LLMs we need a different style than normal one. I think I will work on creating a extension kind of version as I don't see high prio for that on kedro roadmap atm. I will share the updates later on with the community in case anyone find it useful.

astrojuanlu:
@lordsoffallen Just yesterday I opened this discussion https://github.com/kedro-org/kedro/discussions/3979 I'd love to know what makes LLM workflows difficult in Kedro. Do you mind elaborating a bit more there?

lordsoffallen:
> @lordsoffallen Just yesterday I opened this discussion [kedro-org/kedro#3979](https://github.com/kedro-org/kedro/discussions/3979) I'd love to know what makes LLM workflows difficult in Kedro. Do you mind elaborating a bit more there?

Sure, I'll share my workflow and how i image it there :+1:

merelcht:
Hi @lordsoffallen , do you still want to finish this PR and make it ready for review?

lordsoffallen:
> Hi @lordsoffallen , do you still want to finish this PR and make it ready for review?

Hi, thanks for the reminder, kind of forgot about this PR. I think the discussion thread wasn't too busy so I'd say nothing is concluded there yet. With that mind, I don't know the need nor the use case of other people good enough to have something here.  I think we can close the PR until we have something that is concluded that can be implemented. Meanwhile, I can play with my own implementation and see how that works in general.

merelcht:
That sounds good, thank you @lordsoffallen! I really appreciate you thinking with us on this topic ⭐

PR #619, opened 2024-03-19
fix(datasets): Use lazy `.sink_*` instead of collecting data

### LINKED ISSUE #519: polars.LazyPolarsdataset .collect() streaming
## Description
Enable Polars streaming by default when saving [polars.LazyPolarsDataset](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.collect.html#polars.LazyFrame.collect)

## Context
Enables larger-than-memory data processing, one of the main advantages of using Polars LazyFrames.

## Possible Implementation
`.collect(streaming=True)` in polars.LazyPolarsDataset

If streaming cannot be performed for whatever reason, Polars disables streaming automatically at runtime, so having streaming as the default behaviour should be okay.

## Possible Alternatives
Add a flag to enable/disable streaming through data catalog load_args. However, this may be problematic given streaming is not an argument of LazyFrame.sink_csv(), but rather an argument of LazyFrame.collect().

astrojuanlu:
Fix #519.

cc @cpinon-grd @butterlyn could you give this a try?

## Development notes

This doesn't yet have a regression test nor an entry in the release notes.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/polars/lazy_polars_dataset.py
+++ b/kedro-datasets/kedro_datasets/polars/lazy_polars_dataset.py
@@ -207,15 +207,15 @@ class LazyPolarsDataset(AbstractVersionedDataset[pl.LazyFrame, PolarsFrame]):
         save_path = get_filepath_str(self._get_save_path(), self._protocol)
 
         collected_data = None
-        if isinstance(data, pl.LazyFrame):
-            collected_data = data.collect()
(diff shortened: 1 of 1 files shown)


astrojuanlu:
Since `sink_*` methods are still marked as experimental and we will likely see changes very soon (Polars 2.0?) let's close this for now.

PR #647, opened 2024-04-10
chore(datasets): add `mypy` setup for datasets

ankatiyar:
## Description

Fix https://github.com/kedro-org/kedro/issues/2697

## Development notes

- add `mypy` check to the `make plugin=<plugin> lint` step
- Fix type errors flagged by `mypy`
- Bump RTD build workflow to python 3.10 (v 3.9 was failing because of the typing changes)

## Checklist


.github/workflows/kedro-datasets.yml
```diff
@@ -43,15 +43,15 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-      - name: Set up Python 3.9
```
ankatiyar, line 46: context for this change - https://github.com/sphinx-doc/sphinx/issues/8237

kedro-datasets/kedro_datasets/_typing.py
```diff
@@ -6,9 +6,9 @@
 from typing import NewType
 
 TablePreview = NewType("TablePreview", dict)
-ImagePreview = NewType("ImagePreview", bytes)
+ImagePreview = NewType("ImagePreview", str)
 PlotlyPreview = NewType("PlotlyPreview", dict)
-JSONPreview = NewType("JSONPreview", dict)
+JSONPreview = NewType("JSONPreview", str)
```
ankatiyar, line 12: @SajidAlamQB @rashidakanchwala please double check this

rashidakanchwala: It's a bit confusing because `tracking.JSONDataset` returns `json.loads()`, which is a `dict`, while `json.JSONDataset` returns `json.dumps()`, resulting in a `string`.

kedro-datasets/kedro_datasets/matplotlib/matplotlib_writer.py
```diff
@@ -17,14 +18,13 @@
...
+from matplotlib.figure import Figure
 
 from kedro_datasets._typing import ImagePreview
 
 
 class MatplotlibWriter(
-    AbstractVersionedDataset[
-        Union[plt.figure, list[plt.figure], dict[str, plt.figure]], NoReturn
-    ]
+    AbstractVersionedDataset[Union[Figure, list[Figure], dict[str, Figure]], NoReturn]
```
ElenaKhaustova, line 27: Is it intentional to keep `Union` here?

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
```diff
@@ -4,7 +4,7 @@
 """
 
 from copy import deepcopy
-from typing import Any, Optional
+from typing import Any, Optional, Union
```
ElenaKhaustova, line 7: Same question about `Union` - shall we use `|` instead for consistency?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -154,7 +154,8 @@ def hashable(value):
             import ibis
 
             config = deepcopy(self._connection_config)
-            backend = getattr(ibis, config.pop("backend"))
+            backend_attr = config.pop("backend") if config else None
```
ElenaKhaustova, line 157: Shall we use a default value for the `backend`, or are we okay with raising an error if the key is missing?

ElenaKhaustova (commented):
Massive amount of work! 🚀

Left a couple of minor comments.

kedro-datasets/kedro_datasets/_typing.py
ankatiyar: `tracking.JSONDataset.preview()` returns `JSONTrackingPreview` type which is still `dict`
`json.JSONDataset.preview()` returns `JSONPreview` which I've changed to `str`

Since `tracking.JSONDataset` is a subclass of `json.JSONDataset` `mypy` does complain about the overwriting of methods but I've added an ignore - is that okay?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
ankatiyar: I'm not sure about this, @DimedS, thoughts?

kedro-datasets/kedro_datasets/matplotlib/matplotlib_writer.py
ankatiyar: Updated!

kedro-datasets/kedro_datasets/pandas/deltatable_dataset.py
ankatiyar: Updated!

kedro-datasets/kedro_datasets/ibis/table_dataset.py
noklam: I think ibis have a default backend for duckdb. We should stick with the underlying libraries as much as we could. If ibis would fail without the backend argument we should probably do the same.

Do check the default as I only say  this from memory.

kedro-datasets/kedro_datasets/matplotlib/matplotlib_writer.py
ankatiyar: Turns out it is needed here, the python <=3.9 tests fail otherwise

ElenaKhaustova (approved):

Issue #655, opened 2024-04-16
`kedro-airflow`: Grouping in memory is not working properly


ankatiyar:
## Description

The `--group-in-memory` flag is a new addition https://github.com/kedro-org/kedro-plugins/pull/241 and hasn't been released yet.

Running `kedro airflow create --group-in-memory` and `kedro airflow create` with the spaceflights project creates the exact same DAGs - there's some issue in the determining whether a dataset is `MemoryDataset` 

https://github.com/kedro-org/kedro-plugins/blob/fbe545f63fe84b1c3610c091711429f02c7416e4/kedro-airflow/kedro_airflow/grouping.py#L14-L15

^ returns `None` for datasets that are not in the catalog (dataset factory datasets or `MemoryDatasets`

There might also be some problems with dataset factories - still investigating

astrojuanlu:
cc @sbrugman FYI!

PR #664, opened 2024-05-03
fix(airflow): Fix nodes grouping 

ElenaKhaustova:
## Description
Fix https://github.com/kedro-org/kedro-plugins/issues/655

## Checklist


kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -206,22 +206,19 @@ def create(  # noqa: PLR0913, PLR0912
...
+            nodes = {}
+            dependencies = {}
+            for node in pipeline.nodes:
+                nodes[node.name] = [node]
+                dependencies[node.name] = []
             for node, parent_nodes in pipeline.node_dependencies.items():
                 for parent in parent_nodes:
                     dependencies[parent.name].append(node.name)
 
-        # Sort both parent and child nodes to make sure it's deterministic
```
ElenaKhaustova, line 216: This sorting was replaced with the original topological order obtained from kedro; it is deterministic. Otherwise `nodes` and `dependencies` have a different order which is confusing.

ElenaKhaustova:
@ankatiyar and I have also discussed whether we need to add grouping logic to the e2e test. IMO, grouping logic is currently covered well by unit tests, and we can go without it. Unit tests already include `kedro airflow create --group-in-memory` command run, which is kind of the e2e test. But happy to hear others' thoughts.

ankatiyar (approved):
Manually tested, works well! 💯 
ETA: About the e2e test, I think it's still worth adding an e2e test because in the unit tests we are mocking the catalog etc. The bug would have been evident earlier if it there was an e2e test in place. I think we can do it as a follow up task though, I want to explore other grouping strategies as a part of the airflow milestone too.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -206,22 +205,19 @@ def create(  # noqa: PLR0913, PLR0912
...
         else:
-            nodes = {node.name: [node] for node in pipeline.nodes}
-
-            dependencies = defaultdict(list)
+            nodes = {}
+            dependencies = {}
+            for node in pipeline.nodes:
+                nodes[node.name] = [node]
+                dependencies[node.name] = []
             for node, parent_nodes in pipeline.node_dependencies.items():
```
DimedS, line 213: Why do you believe that this row will be deterministic? I mean it affects the order of `.append(node.name)`

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -25,74 +20,88 @@ def get_memory_datasets(catalog: DataCatalog, pipeline: Pipeline) -> set[str]:
...
+        node_output: node for node in pipeline.nodes for node_output in node.outputs
+    }
+
+    for node in pipeline.nodes:
+        for node_input in node.inputs:
+            if node_input in output_to_node:
+                if node_input in memory_datasets:
+                    adj_matrix[node.name].add(output_to_node[node_input].name)
+                    adj_matrix[output_to_node[node_input].name].add(node.name)
+                parents[output_to_node[node_input].name].add(node.name)
```
DimedS, line 45: To me, it seems like that's a `children` dictionary, because the values are children, as I understand it.

DimedS (approved):
Great work, @ElenaKhaustova ! It looks good to me. I just have a few small questions.

kedro-airflow/kedro_airflow/plugin.py
ElenaKhaustova: We are iterating nodes in the order they were resolved with topological sort, which is deterministic, and `nodes` and `dependencies` dictionaries are ordered.

kedro-airflow/kedro_airflow/grouping.py
ElenaKhaustova: Yeah, a key is a parent's name, while value is the children's list. Will rename it to `parent_to_children` for clarity. Thank you!

kedro-airflow/kedro_airflow/plugin.py
DimedS: `dependencies[parent.name].append(node.name)`
I understand that the parents will be deterministic because you initialised an empty dictionary previously. However, could you explain in more detail why the dictionary values containing the children will also be deterministic?

ElenaKhaustova: They're deterministic because all the traversals inside `group_memory_nodes` function are deterministic as well and based on `pipeline.nodes` order which is sorted already.

kedro-airflow/kedro_airflow/grouping.py
ElenaKhaustova: Done

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -25,74 +20,88 @@ def get_memory_datasets(catalog: DataCatalog, pipeline: Pipeline) -> set[str]:
...
+        node_output: node for node in pipeline.nodes for node_output in node.outputs
+    }
+
+    for node in pipeline.nodes:
+        for node_input in node.inputs:
+            if node_input in output_to_node:
+                if node_input in memory_datasets:
+                    adj_matrix[node.name].add(output_to_node[node_input].name)
+                    adj_matrix[output_to_node[node_input].name].add(node.name)
+                parent_to_children[output_to_node[node_input].name].add(node.name)
```
noklam, line 45:
```suggestion
        for node_input in node.inputs:
            if node_input in output_to_node:
                parent_to_children[output_to_node[node_input].name].add(node.name)
                if node_input in memory_datasets:
                    adj_matrix[node.name].add(output_to_node[node_input].name)
                    adj_matrix[output_to_node[node_input].name].add(node.name)
```

If rearranging it does not change the logic, I prefer to keep the related code closer.

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -25,74 +20,88 @@ def get_memory_datasets(catalog: DataCatalog, pipeline: Pipeline) -> set[str]:
...
+    catalog: DataCatalog, pipeline: Pipeline
+) -> tuple[dict[str, set], dict[str, set]]:
+    """
+    Builds adjacency matrix (adj_matrix) to search connected components - undirected graph,
+    and adjacency matrix (parent_to_children) to retrieve connections between new components using on
+    initial kedro topological sort - directed graph.
+    """
+    memory_datasets = get_memory_datasets(catalog, pipeline)
+
+    adj_matrix: dict[str, set] = {node.name: set() for node in pipeline.nodes}
```
noklam, line 33: Is this really an adjacency matrix? I was thinking https://en.wikipedia.org/wiki/Adjacency_matrix but it seems like it's just another dictionary

kedro-airflow/kedro_airflow/plugin.py
DimedS: > Can we add some comments about the intent of keeping it deterministic, it is not so obvious from the code. Ideally there should be tests for it but AFIAK it's a bit tricky to reproduce the randomness. This is only manually tested I assume.

I agree, impossible to test it

DimedS: > They're deterministic because all the traversals inside `group_memory_nodes` function are deterministic as well and based on `pipeline.nodes` order which is sorted already.

Is it means that previously code was not deterministic (and sorting was used) because of using `defaultdict`?

noklam (commented):
Left a couple of small comment. Is this tested manually?

The only concern for me is implementing a new `dfs` here. I recalled I have to implement a bfs for the `SoftFailRunner`, at the end I end up using some `Pipeline` API, which actually is doing the search already. I just want to make sure this is considered already, and do we see a need to extend on the Pipeline API side instead.

example of bfs:
https://github.com/noklam/kedro-softfail-runner/blob/970036ea8d5c969d02ed9150bfd4a2dc4baf967a/kedro_softfail_runner/core.py#L92

kedro-airflow/kedro_airflow/grouping.py
ElenaKhaustova: It's an adjacency list: https://en.wikipedia.org/wiki/Adjacency_list. Will rename it

ElenaKhaustova: Done

kedro-airflow/kedro_airflow/grouping.py
ElenaKhaustova: Done

kedro-airflow/kedro_airflow/plugin.py
ElenaKhaustova: Yes, exactly

ElenaKhaustova:
> Left a couple of small comment. Is this tested manually?
> 
> The only concern for me is implementing a new `dfs` here. I recalled I have to implement a bfs for the `SoftFailRunner`, at the end I end up using some `Pipeline` API, which actually is doing the search already. I just want to make sure this is considered already, and do we see a need to extend on the Pipeline API side instead.
> 
> example of bfs: https://github.com/noklam/kedro-softfail-runner/blob/970036ea8d5c969d02ed9150bfd4a2dc4baf967a/kedro_softfail_runner/core.py#L92

Thank you for the comments, I've addressed them.

Not sure I fully understand the concern about using `dfs` vs `bfs`. We cannot rely just on `Pipeline` API since we need to adjust the graph by adding new edges to group connected MemoryDatasets into one node. Thus we need an updated adjacency matrix and one traversal to find connected components. There's no difference whether it's a `dfs` or `bfs`. The only issue which might happen with `dfs` is if the number of nodes is more the the default size of the recursion stack which is around 1000.

kedro-airflow/kedro_airflow/plugin.py
ElenaKhaustova: We tested manually, yes.

Theoretically, we can randomise `n` test cases and run each test `k` times to ensure we get the same results within runs. But given that there's no randomness, the results will be the same, so I don't think we need it.

noklam: The randomness cannot be simulated with n use cases, this is another stories, but I agree we don't need it here.

ElenaKhaustova: Randomness - no, but you can make a stress test for determinism, which will not guarantee it with 100% chance, but for large n and k, it will be close to it. Anyway, creating an adjacency list is deterministic, and `dfs` is deterministic; except for them, we have other loops which rely on `Pipeline` API, so there's no place from where randomness can come. So, I agree on not doing anything with it. 🙂

noklam:
> Not sure I fully understand the concern about using dfs vs bfs. We cannot rely just on Pipeline API since we need to adjust the graph by adding new edges to group connected MemoryDatasets into one node. Thus we need an updated adjacency matrix and make one traversal to find connected components. There's no difference whether it's a dfs or bfs. The only issue which might happen with dfs is if the number of nodes is more than the default size of the recursion stack which is around 1000.

@ElenaKhaustova I haven't dived deep into this enough to understand it fully. I will approve this now as I don't want to block the PR as it works perfectly fine. 

https://github.com/kedro-org/kedro/discussions/3758 It's another topic but a way to adjust the graph. The API is obviously bad but just want to show that there may be potential to open this up a little bit.

ElenaKhaustova:
> > Not sure I fully understand the concern about using dfs vs bfs. We cannot rely just on Pipeline API since we need to adjust the graph by adding new edges to group connected MemoryDatasets into one node. Thus we need an updated adjacency matrix and make one traversal to find connected components. There's no difference whether it's a dfs or bfs. The only issue which might happen with dfs is if the number of nodes is more than the default size of the recursion stack which is around 1000.
> 
> @ElenaKhaustova I haven't dived deep into this enough to understand it fully. I will approve this now as I don't want to block the PR as it works perfectly fine.
> 
> [kedro-org/kedro#3758](https://github.com/kedro-org/kedro/discussions/3758) It's another topic but a way to adjust the graph. The API is obviously bad but just want to show that there may be potential to open this up a little bit.

To be honest, I don't think the API is bad it's just out of the scope of `pipeline` API.

PR #681, opened 2024-05-17
feat: Exit before_command_run if telemetry consent is passed through kedro run

lrcouto:
## Description

Works together with https://github.com/kedro-org/kedro/pull/3876.

Modifies the `before_command_run` hook in `kedro-telemetry` to exit if the user has passed the telemetry consent through the flag on `kedro run`.

## Checklist


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -107,6 +107,10 @@ def before_command_run(
             if not project_metadata:  # in package mode
                 return
 
+            for arg in command_args:
```
merelcht, line 110: Doesn't this mean though that using the `--telemetry` flag is the same as opting out? If we return at this point, no data will be sent even if the user provided "yes" to the flag.

lrcouto: Hmm I imagined the person  would only use the flag once and then on subsequent runs that would be handled, but you're right. I'm going to look over this one again.

DimedS:
Why, then, did we create other PRs for the Kedro CLI? Perhaps that task can be handled within that hook for all commands?

lrcouto:
> Why, then, did we create other PRs for the Kedro CLI? Perhaps that task can be handled within that hook for all commands?

We'd still need at the very least to have a PR that introduces the "--telemetry" flag to the `run` function to people to use in the CLI, so it needs to be changed regardless.

DimedS:
> > Why, then, did we create other PRs for the Kedro CLI? Perhaps that task can be handled within that hook for all commands?
> 
> We'd still need at the very least to have a PR that introduces the "--telemetry" flag to the `run` function to people to use in the CLI, so it needs to be changed regardless.

In that case, I believe it would be better to put all the logic for validating the answer and writing to the file into that hook (since it was already there previously). We can then add just a few lines to the Kedro CLI to initialise that command. This approach avoids duplication and ensures the same logic is used for all commands. Do you know why that hook doesn't work on `kedro new` command?

DimedS:
> What do you think @lrcouto @merelcht @DimedS
I agree with your point, but for me, it's more about naming. We can change it to `set_telemetry=on/off` and keep the concept as it is. I believe it's useful to provide this option to users. Additionally, I think the core issue is that we need to run the hook before the `kedro-new` command. This way, we can ask users if they want to participate in telemetry when they are creating a project, rather than during the first operation in the project. This approach avoids potential confusion, especially if that first operation occurs in a production environment.

lrcouto:
> What do you think @lrcouto @merelcht @DimedS

In a vacuum, a do agree with this notion that you should be able to use the command to decide if telemetry will be collected on a run by run basis, but the idea of this specific task was to introduce a CLI option that reproduces the behavior of the prompt the user gets when running the CLI on a project for the first time, and that writes on the `.telemetry` file. So the idea would be using the flag only if you do want to write on the file and change consent. This is also the behavior you see on the flag in the `new` function.

If we were to do that, I'd change the name of the flag so it doesn't become confusing, since we have another one with the same name that does change the `.telemetry` file.

Changing the prompt to consent or not to telemetry to project creation instead of on the first CLI run would need to change when the plugin first becomes available, and it's a bit out of the scope of this specific task. But we could discuss that and maybe open another issue.

lrcouto:
Also, if we do want to rethink how this is going to work, should we revert https://github.com/kedro-org/kedro/pull/3876? With only that one merged and not this one, it will allow the user to set consent through a flag but the prompt will override it.

DimedS:
I agree with @ankatiyar proposals that setting telemetry on `kedro new` should apply to the entire project (we should also consider asking users about telemetry at that step, rather than just allowing them to specify `--telemetry`). I agree that from one side it will be clearer if using the `--telemetry` option with other commands applies only to that particular run, but from other side - the same flag will behave differently. However, I believe we need to gather more opinions on this matter.

I also concur with @lrcouto that we should finalise the feature design before proceeding with the coding. The PR, which has already been merged, is not currently properly functioning. It seems that we will need the next sprint to agree on a new solution and implement it. Therefore, I support reverting the previous PR and rethinking the approach to avoid blocking the current release.

noklam:
Just left a comment here before everyone jump in, I asked @astrojuanlu to clarify the requirements here. I think both solutions are correct but the main question here is what the desired user journey.

merelcht:
Thanks all for your thoughts on this. I've looked at the original tickets and comments and I agree that the feature ask actually wasn't clear here. I interpreted it the same way as @lrcouto did, thinking that `kedro new --telemetry` and `kedro run --telemetry` should behave the same. The main reason we introduced this feature is for automated processes (test in CI), so I didn't actually think about the setting of telemetry for a specific run. But @ankatiyar brings some very thoughtful points.

I still think that the way @lrcouto built it is my ideal approach, but would be good to get a product view from @astrojuanlu

lrcouto:
PR for reversion here in case we want it: https://github.com/kedro-org/kedro/pull/3888

We could go either way, proceed as it is or revert and go back to the drawing table.

astrojuanlu:
To recap on how we got here:

The original user problem was described in #1640:

> When any kedro command is executed for the first time or in a clean environment (which is often the case in CI/CD) the telemetry prompt gets run. The user then has to answer Yes or No to running telemetry. If no user is involved, e.g. in an automated CI/CD workflow a hack needs to be put in place to programatically add a .telemetry file.

Then in #2867 a solution was proposed, but it fixated in `kedro new`.

And in the middle of the implementation #3701, we realised that adding that flag to `kedro new` didn't actually solve the original problem https://github.com/kedro-org/kedro/pull/3701#pullrequestreview-1945846072 because it's `kedro run` the command that is most likely to run on an unattended environment and therefore get stuck.

So in https://github.com/kedro-org/kedro/pull/3701#issuecomment-2015339202 we agreed to add "a similar flag" to `kedro run`.

But indeed, we never fully clarified what said flag was supposed to do.

I do agree with @ankatiyar that it's weird for `kedro run --telemetry` to set the telemetry consent for the whole project, rather than for the specific run. Essentially what she described in https://github.com/kedro-org/kedro-plugins/pull/681#issuecomment-2127020217. Responding to a couple of things:

>  (This will require adding `--telemetry` to all project subcommands)

I suggested that we take this to `kedro-telemetry` but it was not clear how feasible it was, plus we wanted to be practical and cover just `kedro new` and `kedro run`.

> OR maybe we can add a CLI command to `kedro-telemetry` to turn change consent to false `kedro telemetry off` or somthing?

This is an idea we hadn't considered yet and maybe worth giving a second thought. At least it decouples `kedro` commands from `kedro-telemetry` stuff, which is really neat and clean.

As a user, going to my CI runs and changing `kedro run` to `kedro run --skip-telemetry` or doing `kedro telemetry off && kedro run` looks mostly the same (?).

noklam:
>As a user, going to my CI runs and changing kedro run to kedro run --skip-telemetry or doing kedro telemetry off && kedro run looks mostly the same (?).

Conceptually this de-coupling is cleaner, but I don't know if it's really useful. This would work equally well:
```
pip uninstall kedro-telemetry && kedro run
```

lrcouto:
From my understanding, the main idea of this flag would be avoiding an automated run to stop because of the consent prompt, in a CI situation or similar. Our own e2e tests, for example, are manually creating the `.telemetry` file to prevent that from happening while they run.

astrojuanlu:
After discussing this a bit more internally, we decided to revert this PR.

I'll give a proper rationale in https://github.com/kedro-org/kedro/issues/1640 but essentially, the way we will solve the problem of "telemetry prompt blocking unattended environments" is actually removing the prompt, rather than adding a `--telemetry` flag to more `kedro` CLI commands.

Thanks everybody for the thoughtful discussion here, the requirements weren't clear and @lrcouto you did the right thing 🙏🏼 Let's close this one and go with https://github.com/kedro-org/kedro/pull/3888.

PR #689, opened 2024-05-22
Update badges for kedro-datasets

noklam:
## Checklist

### DIFF
--- a/kedro-datasets/README.md
+++ b/kedro-datasets/README.md
@@ -3,7 +3,7 @@
 <!-- Note that the contents of this file are also used in the documentation, see docs/source/index.md -->
 
 [![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
-[![Python Version](https://img.shields.io/badge/python-3.9%20%7C%203.10%20%7C%203.11-blue.svg)](https://pypi.org/project/kedro-datasets/)
+[![Python Version](https://img.shields.io/badge/python-3.9%20%7C%203.10%20%7C%203.11%20%7C%203.12-blue.svg)](https://pypi.org/project/kedro-datasets/)
 [![PyPI Version](https://badge.fury.io/py/kedro-datasets.svg)](https://pypi.org/project/kedro-datasets/)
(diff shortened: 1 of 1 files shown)


astrojuanlu (approved):
🚀

ankatiyar (approved):
good catch!

merelcht (changes requested):
I think we should wait a sec with dropping 3.9 support, because I think Viz relies on `kedro-datasets` @rashidakanchwala. I'd like us to revisit this "imbalance" in support of Python versions across plugins and Kedro, because it has led to complications when we did that transcoding change.

rashidakanchwala:
The last time this happened, we made significant code changes on Kedro-viz/Kedro-datasets in order to decouple it. And because kedro-datasets dropped support for python 3.8 whilst kedro-viz still supported it; we had to drop support for 3.8 because there were breaking changes and it didn't work for 3.8; the other alternative was to keep the old code bits to support python 3.8 users. 

I think moving forward, we wouldn't be impacted as much -- only the new features in kedro-datasets wouldn't be available to python 3.9 users.

deepyaman:
> I think we should wait a sec with dropping 3.9 support, because I think Viz relies on `kedro-datasets` @rashidakanchwala. I'd like us to revisit this "imbalance" in support of Python versions across plugins and Kedro, because it has led to complications when we did that transcoding change.

This shouldn't cause any issues, because pip won't let you install a library that isn't supported on the earlier version, and will simply fall back to an older version that does support the Python version.

That said, there's also no real point in dropping support without making code changes that are 3.9-incompatible.

astrojuanlu:
Just to reduce the risk, we could make a kedro-viz release _before_ dropping Python 3.9 support here.

noklam:
@astrojuanlu @merelcht Let's keep this a separate discussion as there are different opinions here. I agree with @deepyaman that there is no rush to drop support when it's perfectly compatible.

This PR will be limited to the badge update only.

merelcht (approved):
💯 Agree to discuss the dropping of 3.9 for datasets separately!

ankatiyar (approved):

PR #696, opened 2024-05-23
feat(datasets): Add option to async load and save in PartitionedDatasets

puneeter:
## Description
- This PR provides the user to load and save `PartitionedDataset` asynchronously for partitions provided.
- PartitionedDatasets already provide a way to do lazy loading, which solves for memory complexity. With this PR the time complexity is also reduced if the user wants to save/load these partitions in parallel with the help of `use_async` argument.

## Development notes
- Additional `use_async` argument to `PartitionedDataset` constructor is used to control the async load/save.
- Based on this `argument`, `_save` and `_load` methods call different private functions.
- Leveraged existing tests for `PartitionedDataset` by parameterizing value for `use_async` using `@pytest.mark.parametrize("use_async", [True, False])`

## Checklist


puneeter:
> Hi @puneeter, can you please provide a description and any relevant development notes on the PR? This will make it easier for the team to review.

I updated the description. Please let me know if it needs any refactoring.

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
```diff
@@ -300,7 +311,32 @@ def _load(self) -> dict[str, Callable[[], Any]]:
...
+        async def load_partition(partition: str) -> None:
+            kwargs = deepcopy(self._dataset_config)
+            kwargs[self._filepath_arg] = self._join_protocol(partition)
+            dataset = self._dataset_type(**kwargs)  # type: ignore
+            partition_id = self._path_to_partition(partition)
+            partitions[partition_id] = dataset.load
+
+        await asyncio.gather(
+            *[load_partition(partition) for partition in self._list_partitions()]
+        )
```
astrojuanlu, line 326: If I understand correctly, there's no actual I/O being performed here, right? Only the `partitions` dictionary is being populated.

I don't see the need of using async helpers and `asyncio.gather` here.

If anything, as a user I'd expect to have the `async` loaders available in my node function so that I can `await` them (provided that my node is asynchronous), use `asyncio.gather` myself, or use an `asyncio.TaskGroup`.

```yaml
my_partitioned_dataset:
  type: partitions.PartitionedDataset
  path: s3://my-bucket-name/path/to/folder
  ...
  use_async: True
```

```python
def concat_partitions(partitioned_input: dict[str, Awaitable]) -> pd.DataFrame:
    tasks = []
    async with asyncio.TaskGroup() as tg:
        for partition_key, partition_load_func in sorted(partitioned_input.items()):
            tasks.append(tg.create_task(partition_load_func()))

    result = pd.DataFrame()
    result = pd.concat([result] + [tasks.result() for task in tasks], ignore_index=True, sort=True)
```

(not that I find this a particularly friendly DX, but it's more or less a continuation of our current approach https://docs.kedro.org/en/stable/data/partitioned_and_incremental_datasets.html#partitioned-dataset-load)

What am I missing?

puneeter: That makes sense. I am open to both the options. Let me know if you want to revert the load method to the original definition. Happy to also update the documentation once we are aligned with the changes made

noklam: Having fine control of async in node is ok IMO. Is there a way of avoiding coupling node logic to it?

IIRC the original question was using the async option of Runner, and we found that the partitioned dataset only do async on the whole dataset level.

ElenaKhaustova: Is there any reason why we prefer making it at the dataset level rather than runner? It seems like having the common approach at the above layer is needed anyway to make it efficient.

astrojuanlu: For now I think this is to achieve consistency with synchronous `PartitionedDatasets`, not sure what you have in mind for runners but maybe we should discuss that separately? Unless you still see issues with the proposed approach

puneeter:
Would need team's help to point to the right documentation to be changed because of this change.

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
```diff
@@ -302,6 +307,12 @@ def load(self) -> dict[str, Callable[[], Any]]:
         return partitions
 
     def save(self, data: dict[str, Any]) -> None:
+        if self._use_async:
+            asyncio.run(self._async_save(data))
```
astrojuanlu, line 311: If I understand correctly, `asyncio.run` creates a new event loop, so if there's already an event loop running (for example, in a Jupyter notebook), calling this will raise an error.

This is essentially the [red/blue function problem](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/)... Most of Kedro is synchronous anyway AFAIK, but I think this might set an API expectation that could be difficult to satisfy cleanly.

@merelcht @ElenaKhaustova do you have more thoughts?

merelcht: > If I understand correctly, asyncio.run creates a new event loop, so if there's already an event loop running (for example, in a Jupyter notebook), calling this will raise an error.

That is indeed correct. Maybe it's alright though to say async saving doesn't work in interactive envs? 

> but I think this might set an API expectation that could be difficult to satisfy cleanly

@astrojuanlu can you elaborate what you mean with this?

merelcht:
@puneeter I see all the tests have been modified to take the `use_async` argument, but is there a way to also check that the async functionality is working?

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
astrojuanlu: Sorry for the delay. What I meant is that use_async=True as it stands in this PR won't work (I think) in certain cases, and that fixing it might be hard

astrojuanlu:
My comments on `asyncio.run` above were premonitory, because we ended up finding a actual example of such breakage in a separate PR https://github.com/kedro-org/kedro/issues/4611

I think this had good intentions but it might actually be difficult to do. @puneeter do you mind if we turn this PR into an issue and we tackle it at some other time?

Issue #672, opened 2024-05-08
kedro-airflow: Allow passing `conf_source` to the created DAG


ankatiyar:
## Description
Add `--conf-source` flag to `kedro airflow create` which will pass this on to the Jinja template used to generate the DAGs `airflow_dag_template.j2`

## Context
To run a Kedro project as a DAG on airflow currently -
- You should package and install your kedro project in the executor env
- Copy the config from the Kedro project to the Airflow project folder https://docs.kedro.org/en/stable/deployment/airflow.html#create-prepare-and-package-example-kedro-project
- If something changes in the conf, one must also remember to copy over the changes

Additionally,
- `KedroSession.create()` now allows passing `conf_source` path
- `kedro package` creates a `.whl` file for the project pipelines etc and a `.tar.gz` file for the `conf/` folder of the project which can be saved elsewhere

## Possible Implementation

* Add `--conf-source` to the CLI command which passes this info the Jinja template `airflow_dag_template.j2`

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

astrojuanlu:
@DimedS: "This change would be quite useful to deploy Kedro to some cloud Airflow because it's sometimes impossible to put configuration in the working directory" (paraphrasing)

PR #712, opened 2024-06-03
feat(airflow): Add `--conf-source` to DAG created with `kedro-airflow`

ankatiyar:
## Description

Resolve https://github.com/kedro-org/kedro-plugins/issues/672

## Checklist

### DIFF
--- a/kedro-airflow/tests/test_plugin.py
+++ b/kedro-airflow/tests/test_plugin.py
@@ -376,3 +376,18 @@ def test_create_airflow_all_and_pipeline(cli_runner, metadata):
         "Error: Invalid value: The `--all` and `--pipeline` option are mutually exclusive."
         in result.stdout
     )
+
+
+def test_create_airflow_conf_source(cli_runner, metadata):
+    command = ["airflow", "create", "--conf-source", "conf/"]
+    result = cli_runner.invoke(commands, command, obj=metadata)
+    assert result.exit_code == 0
+    dag_file = metadata.project_path / "airflow_dags" / "fake_project_dag.py"
+
+    assert dag_file.exists()
+
+    expected_airflow_dag = f'conf_source = "{metadata.project_path}/conf"'
+    with dag_file.open(encoding="utf-8") as f:
+        dag_code = [line.strip() for line in f.read().splitlines()]
+        assert expected_airflow_dag in dag_code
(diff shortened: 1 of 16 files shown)


ankatiyar:
Ready!

kedro-airflow/tests/test_plugin.py
```diff
@@ -376,3 +376,19 @@ def test_create_airflow_all_and_pipeline(cli_runner, metadata):
...
+    result = cli_runner.invoke(commands, command, obj=metadata)
+    assert result.exit_code == 0
+    dag_file = metadata.project_path / "airflow_dags" / "fake_project_dag.py"
+
+    assert dag_file.exists()
+
+    expected_airflow_dag = f'conf_source = "{Path(metadata.project_path / "conf").resolve()}" or Path.cwd() / "conf"'
+    with dag_file.open(encoding="utf-8") as f:
+        dag_code = [line.strip() for line in f.read().splitlines()]
+        print(dag_code)
```
merelcht, line 392: left-over print

DimedS:
Great work, @ankatiyar ! The PR looks good, but I have one concern, I was following [deployment manual for cloud services](https://docs.kedro.org/en/stable/deployment/airflow.html) and tried to run 
`kedro airflow create --conf-source="plugins/conf-new_kedro_project.tar.gz"`
I received an error:
`Error: Invalid value for '--conf-source': Path 'plugins/conf-new_kedro_project.tar.gz' does not exist.`
I believe that's because I don't have that file in my current folder, but it will be on cloud container after it will be unpacked in plugins folder, so I think maybe we don't need that strict check on DAG creation step? 

Also it will be great to update deployment docs according to that new feature.

ankatiyar:
@DimedS you're right, I turned of the check to see if the path exists.

ankatiyar:
@DimedS Just wanted to check why resolved paths wouldn't work?

DimedS:
> @DimedS Just wanted to check why resolved paths wouldn't work?

I think typically I'm creating a DAG in my local computer and it will be one absolute path and when I will run that DAG in some container in the cloud environment, absolute path to the same config will be different. So it's ok to use relative path, but not absolute.

DimedS (approved):
Great job, @ankatiyar !

Issue #726, opened 2024-06-14
kedro-telemetry: integrate kedro-telemetry as a core dependency in kedro


DimedS:
## Description
Part of #715:

To implement opt-out telemetry as per the main issue, we need to make changes to how the `kedro-telemetry` plugin is included in kedro projects.

1. Remove the `kedro-telemetry` dependency from the requirements.txt file in Kedro project template and all starters.
2. Modify the `pyproject.toml` file of kedro to include `kedro-telemetry` in the `dependencies` section.

Issue #730, opened 2024-06-14
`kedro-telemetry`: Spike to reduce redundant telemetry events 


DimedS:
## Description

Currently, when running a `kedro new` command with telemetry enabled, we send mostly the same data using `_send_heap_event()` 3! times. This seems excessive, especially as we move to opt-out telemetry with #715, which will definitely increase traffic.

The reason for this redundancy is that we send two identical events after each kedro command in the `before_command_run` hook:
```
            _send_heap_event(
                event_name=f"Command run: {main_command}",
                identity=user_uuid,
                properties=cli_properties,
            )

            # send generic event too, so it's easier in data processing
            generic_properties = deepcopy(cli_properties)
            generic_properties["main_command"] = main_command
            _send_heap_event(
                event_name="CLI command",
                identity=user_uuid,
                properties=generic_properties,
            )
```
I don't know the exact reasons for this duplication, but it seems like the last one is sufficient since it fully contains the first one.

Additionally, if `after_catalog_created` is triggered, we send one more event with data that was already included in the previous two events, along with additional project properties like the number of nodes and pipelines.

So proposal is to send only one piece of data per command:

1. If `after_catalog_created` will be triggered after the command, send all information for that command in that hook.
2. If not, send all information about that command in the `before_command_run` hook.

We need to gather requirements about what exactly is needed on the `HEAP`/ snowflake(?) side. 

Additionally, after revising, please address the following:

1. Update the notification accordingly, as mentioned [here](https://github.com/kedro-org/kedro-plugins/pull/760#discussion_r1672421191).
2. Review the try-except blocks to ensure they are not excessive, as mentioned [here](https://github.com/kedro-org/kedro-plugins/pull/760#pullrequestreview-2168743351).

noklam:
I have some idea, which may depends on 
- https://github.com/kedro-org/kedro-plugins/pull/707

There are 2 event sent originally.
1. run command specific
2. generic event

The 3 was added in `after_catalog_created` for project statistic and this will be triggered for both CLI/ non-CLI run.
1. was created mainly because it's easier to analyse the result directly in HEAP. Processing 2 requires some extra data cleaning and is harder to do it on HEAP. As mentioned, 2. is the superset so we can always retrieve those info from Snowflake, but we need to understand what's the impact of removing 1. event.

Is the performance critical? Especially if we can send the event only at the end, I don't think there will be a strong performance hit (I haven't seen any complains about this before). 

[...]

Issue #728, opened 2024-06-14
`kedro-telemetry`: Introduce environment variables for disabling telemetry and revise disabling methods


DimedS:
## Description
Part of https://github.com/kedro-org/kedro-plugins/issues/715:

Introduce two new environment variables for disabling telemetry:
- `DO_NOT_TRACK`
- `KEDRO_DISABLE_TELEMETRY`
When either of these variables is set to True or 1, telemetry should not be used. This will have the same effect as setting `consent: false` in the `.telemetry` file, but without the need to modify the file. We will only check the consent.

## Context
Detailed explanation provided by @astrojuanlu in [this comment](https://github.com/kedro-org/kedro-plugins/issues/715#issuecomment-2161181479).

When all the opt-out tasks for #715 are completed these should be the ways to disable telemetry:
1. Environment Variables:
Set the environment variables `DO_NOT_TRACK` or `KEDRO_DISABLE_TELEMETRY` to 1. Setting either of these variables will fully disable telemetry for all projects in that environment.
2. Telemetry File:
Create a `.telemetry` file in the root folder of the Kedro project and write there `consent: false`. This disables telemetry for any kedro command executed within that project's folder.
3. CLI Option when creating a new project:
When creating a new project, users can use the command `kedro new --telemetry=yes/no`. This will create a `.telemetry` file with `consent: true/false` accordingly. However, telemetry data about the execution of the kedro new command will still be sent after #729 implementation.
4. `pip uninstall kedro-telemetry`
5. CLI telemetry Disable for each command execution:
Previously, we attempted to implement a CLI consent disable for `kedro run` command (see [Issue 2867](https://github.com/kedro-org/kedro/issues/2867)), but it was reverted due to unclear initial design. Considering the shift to an opt-out model, I believe @ankatiyar's [proposal](https://github.com/kedro-org/kedro-plugins/pull/681#issuecomment-2127020217) to provide consent for each execution of certain Kedro commands is still relevant. 

**Let's discuss this proposal**:
- Implement a `--no-telemetry` flag for `kedro run` and possibly other commands: `kedro run --no-telemetry`.
- This flag will disable telemetry only for that execution of that particular command.
- This can be beneficial for users who want to run their projects in CI/CD processes without needing to set environment variables.

astrojuanlu:
4. `pip uninstall kedro-telemetry`

[...]

Issue #727, opened 2024-06-14
`kedro-telemetry`: Implement telemetry message notification


DimedS:
## Description
Part of https://github.com/kedro-org/kedro-plugins/issues/715:

Implement a new telemetry message to inform users about anonymous usage data collection:

```
Kedro is sending anonymous usage data with the sole purpose of improving the product. No personal data or IP addresses are stored on our side. 
If you want to opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, or create a  `.telemetry` file in the current working directory with the contents `consent: false`.
Read more at https://docs.kedro.org/en/latest/configuration/telemetry.html
```

Display that message after:
1. Kedro installation with `pip install kedro` command.
2. Each kedro command: once per kedro command execution when the `kedro-telemetry` plugin was used to send telemetry (consent wasn't disabled). 

QUESTION PLEASE COMMENT: Should we message something If consent was disabled
Currently, the message displayed when consent is disabled is:
```
Kedro-Telemetry is installed, but you have opted out of
sharing usage analytics so none will be collected.
```

astrojuanlu:
I think it will be difficult or outright impossible to display it after `pip install`. Precompiled wheels are just `unzip && cp`, no chance to give any message.

astrojuanlu:
Oh I see how my phrasing led you into investigating this. Yeah, enough with Kedro commands.

DimedS:
> Oh I see how my phrasing led you into investigating this. Yeah, enough with Kedro commands.

I reread your phrase and understand that it seems you mean installing Kedro and then executing `kedro run` and then printing. I interpreted it slightly differently)) However, it would actually be great to print it after running `pip install kedro`, before the user executes any command, but it looks like there really is no way to do that.

PR #744, opened 2024-07-01
feat(telemetry): Remove consent confirmation prompt for `kedro-telemetry`

ElenaKhaustova:
## Description
Partly solves https://github.com/kedro-org/kedro-plugins/issues/726

## Development notes

- `_confirm_consent` was removed
- now we use consent from the `.telemetry` file if it exists and has a valid format otherwise, we consider consent is `True`

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -338,9 +338,6 @@ def _send_heap_event(
                 resp.status_code,
                 resp.reason,
             )
-
-        print(resp.status_code, resp.reason)
-        print(data)
     except requests.exceptions.RequestException as exc:
         logger.warning(
             "Failed to send data to Heap. Exception of type '%s' was raised.",
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -347,50 +347,23 @@ def _send_heap_event(
 
 
 def _check_for_telemetry_consent(project_path: Path) -> bool:
+    """
+    Use a telemetry consent from ".telemetry" file if it exists and has a valid format.
+    Telemetry is considered as opt-in otherwise.
+    """
     telemetry_file_path = project_path / ".telemetry"
-    if not telemetry_file_path.exists():
-        return _confirm_consent(telemetry_file_path)
-    with open(telemetry_file_path, encoding="utf-8") as telemetry_file:
-        telemetry = yaml.safe_load(telemetry_file)
-        if _is_valid_syntax(telemetry):
-            return telemetry["consent"]
-        return _confirm_consent(telemetry_file_path)
+    if telemetry_file_path.exists():
+        with open(telemetry_file_path, encoding="utf-8") as telemetry_file:
+            telemetry = yaml.safe_load(telemetry_file)
+            if _is_valid_syntax(telemetry):
+                return telemetry["consent"]
+    return True
 
 
 def _is_valid_syntax(telemetry: Any) -> bool:
     return isinstance(telemetry, dict) and isinstance(
         telemetry.get("consent", None), bool
     )
-
-
-def _confirm_consent(telemetry_file_path: Path) -> bool:
-    try:
-        with telemetry_file_path.open("w") as telemetry_file:
-            confirm_msg = (
-                "As an open-source project, we collect usage analytics. \n"
-                "We cannot see nor store information contained in "
-                "a Kedro project. \nYou can find out more by reading our "
-                "privacy notice: \n"
-                "https://github.com/kedro-org/kedro-plugins/tree/main/kedro-telemetry#"
-                "privacy-notice \n"
-                "Do you opt into usage analytics? "
-            )
-            if click.confirm(confirm_msg):
-                yaml.dump({"consent": True}, telemetry_file)
-                click.secho("You have opted into product usage analytics.", fg="green")
-                return True
-            click.secho(
-                "You have opted out of product usage analytics, so none will be collected.",
-                fg="green",
-            )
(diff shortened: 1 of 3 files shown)


astrojuanlu:
A few things:

- I contend that we should merge this one before proceeding with https://github.com/kedro-org/kedro/pull/3976
- Maybe we could try to address #728 in this same PR
- I gave this a try with a previously created project, and got this:

```
$ kedro run
[07/04/24 08:19:51] INFO     Using `conf/logging.yml` as logging configuration. You can change this by setting the KEDRO_LOGGING_CONFIG environment __init__.py:249
                             variable accordingly.                                                                                                                 
                    ERROR    Failed to retrieve UUID: badly formed hexadecimal UUID string                                                             plugin.py:78
[07/04/24 08:19:52] WARNING  Failed to send data to Heap. Response code returned: 400, Response reason: Bad Request                                   plugin.py:336
                    WARNING  Failed to send data to Heap. Response code returned: 400, Response reason: Bad Request                                   plugin.py:336
```

which is probably not related to this PR, but maybe it's too scary (specially given that it's printed twice, maybe because of #730?). I think there are 2 things here:

- The "Failed to retrieve UUID" message introduced in #596 could maybe be "Could not find or create project UUID" and maybe set it as a warning (although I'm hesitating whether the user needs to see it at all)
- The HTTP 400 to Heap was unexpected, maybe there's something wrong when the Project UUID is empty? https://github.com/kedro-org/kedro-plugins/blob/92bf6ebaa249c9e144fc6fee18a83b76973c13f2/kedro-telemetry/kedro_telemetry/plugin.py#L80

And finally, the "Kedro is sending anonymous usage data" message I proposed in https://github.com/kedro-org/kedro-plugins/issues/715 is not being shown.

ElenaKhaustova:
> A few things:
> 
> * I contend that we should merge this one before proceeding with [Make `kedro-telemetry` a core dependency kedro#3976](https://github.com/kedro-org/kedro/pull/3976)
> * Maybe we could try to address [`kedro-telemetry`: Introduce environment variables for disabling telemetry and revise disabling methods #728](https://github.com/kedro-org/kedro-plugins/issues/728) in this same PR
> * I gave this a try with a previously created project, and got this:
> 
> ```
> $ kedro run
> [07/04/24 08:19:51] INFO     Using `conf/logging.yml` as logging configuration. You can change this by setting the KEDRO_LOGGING_CONFIG environment __init__.py:249
>                              variable accordingly.                                                                                                                 
>                     ERROR    Failed to retrieve UUID: badly formed hexadecimal UUID string                                                             plugin.py:78
> [07/04/24 08:19:52] WARNING  Failed to send data to Heap. Response code returned: 400, Response reason: Bad Request                                   plugin.py:336
>                     WARNING  Failed to send data to Heap. Response code returned: 400, Response reason: Bad Request                                   plugin.py:336
> ```
> 
> which is probably not related to this PR, but maybe it's too scary (specially given that it's printed twice, maybe because of #730?). I think there are 2 things here:
> 
> * The "Failed to retrieve UUID" message introduced in [feat(telemetry): Unique User IDs in kedro-telemetry - merge only for kedro-telemetry release 0.4.0 #596](https://github.com/kedro-org/kedro-plugins/pull/596) could maybe be "Could not find or create project UUID" and maybe set it as a warning (although I'm hesitating whether the user needs to see it at all)
> * The HTTP 400 to Heap was unexpected, maybe there's something wrong when the Project UUID is empty? https://github.com/kedro-org/kedro-plugins/blob/92bf6ebaa249c9e144fc6fee18a83b76973c13f2/kedro-telemetry/kedro_telemetry/plugin.py#L80
> 
> And finally, the "Kedro is sending anonymous usage data" message I proposed in #715 is not being shown.

@astrojuanlu, thank you for the review!

1. I believe there is not much difference if we merge this PR before #3976 or after as it only removes the consent confirmation prompt and works fine with the latest `kedro==0.19.6` version and changes introduced in the PR #3976
2. I think we can follow @DimedS split and address #727 ("Kedro is sending anonymous usage data" will be added with this PR) and #728 separately on top of this PR and, once they're ready, merge them together one by one.
3. I was not able to reproduce the errors you mentioned. I've tested the current change on `spaceflights-pandas` with `kedro==0.19.6` and changes from here #3976. Both work well. I double checked the response status when sending to the heap - it looks good - `200 OK`. I also checked that project ID is creating well and is part of the query. But anyway, the current changes cannot not affect the project ID logic. Could you please check if the error is not appearing for you without current changes?

DimedS (approved):
Thank you, @ElenaKhaustova. I tested all the PRs together, and everything works well. I also checked the telemetry info on Heap, and it's functioning correctly.

I believe we can merge the current PR and [this one](https://github.com/kedro-org/kedro/pull/3976), then continue with [message notification](https://github.com/kedro-org/kedro-plugins/issues/727) and [new disabling methods](https://github.com/kedro-org/kedro-plugins/issues/728), followed by other telemetry tickets.

I agree on merging the [Starters PR](https://github.com/kedro-org/kedro-starters/pull/224) last, just before the release.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -347,14 +346,17 @@ def _send_heap_event(
 
 
 def _check_for_telemetry_consent(project_path: Path) -> bool:
+    """
+    Use a telemetry consent from ".telemetry" file if it exists and has a valid format.
```
merelcht, line 350:
```suggestion
    Use telemetry consent from ".telemetry" file if it exists and has a valid format.
```

merelcht (approved):
I didn't test this, but from a code perspective this looks good to me. I agree that the environment variables should be done separately. We just need to make sure we don't do any releases until the full opt-out flow is implemented in all parts.

PR #764, opened 2024-07-11
fix(telemetry): Masking test

ElenaKhaustova:
## Description
Fix https://github.com/kedro-org/kedro-plugins/issues/763

## Checklist

### DIFF
--- a/kedro-telemetry/tests/test_masking.py
+++ b/kedro-telemetry/tests/test_masking.py
@@ -20,6 +20,7 @@ REPO_NAME = "cli_tools_dummy_project"
 PACKAGE_NAME = "cli_tools_dummy_package"
 DEFAULT_KEDRO_COMMANDS = [
     "catalog",
+    "info",
     "ipython",
     "jupyter",
     "micropkg",


kedro-telemetry/tests/test_masking.py
```diff
@@ -20,6 +20,7 @@
 PACKAGE_NAME = "cli_tools_dummy_package"
 DEFAULT_KEDRO_COMMANDS = [
     "catalog",
+    "info",
```
ElenaKhaustova, line 23: Any idea why `"info"` was not in `DEFAULT_KEDRO_COMMANDS` before?

DimedS (approved):
Thank you, @ElenaKhaustova

kedro-telemetry/tests/test_masking.py
ElenaKhaustova: I suspect it’s connected to these changes [Lazy load click subcommands PR](https://github.com/kedro-org/kedro/pull/3883)

merelcht: I think info is part of the global commands.

merelcht: Hmm actually I'm not 100% sure, but it looks like the `info` command is part of a different separate CLI group compared to the others. @ankatiyar probably knows best 😄

ElenaKhaustova: I see, but what confuses me is that `new` and `starter` are also global but were in the `DEFAULT_KEDRO_COMMANDS` but not `info`.

ankatiyar (approved):
Thanks @ElenaKhaustova!

kedro-telemetry/tests/test_masking.py
ankatiyar: `info` is a separate group because the group has the help text that shows up on top when you do `kedro --help`. I didn't change the setup to lazy for that just in case something broke. The cli group is called `cli` and the module itself is also called `cli`., I suspect the test setup was a bit weird with some mocking and importing so `info` wasn't getting discovered in tests but I'm also not a 100% 

For example, see the `tests/framework/cli/test_cli_hooks.py` in https://github.com/kedro-org/kedro/pull/3883 😕

Issue #709, opened 2024-05-30
kedro-telemetry: Improve performance


ankatiyar:
## Description
Follow up on https://github.com/kedro-org/kedro/issues/1476

From profiling various Kedro CLI commands, it turns out that a good chunk of the time is spent sending information to Heap in the `before_command_run` hook. This issue is to improve things on `kedro-telemetry` side.

![image](https://github.com/kedro-org/kedro-plugins/assets/110245118/83ebcb40-eb87-4639-bad4-07f34994d17e)

## Context
https://github.com/kedro-org/kedro/issues/1476#issuecomment-2135588917
https://github.com/kedro-org/kedro/issues/1476#issuecomment-2139515747

## Possible Implementation
- Move to `after_command_run` hook, but this does not get executed if an exception occurs during the `kedro run` etc
- Try asynchronous sending of Heap data but this might be tricky since it's still an open request on `pluggy` https://github.com/pytest-dev/pluggy/issues/320

### Also to consider
There's some project hooks executed by `kedro-telemetry` too - we could figure out how to improve performance by not having to send this data twice

ankatiyar:
- `kedro-plugin` side solution already in #766 
- Kedro side solution https://github.com/kedro-org/kedro/pull/4014

PR #707, opened 2024-05-30
feat(telemetry): Switch to after_command_run hook

ankatiyar:
## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #766, opened 2024-07-12
feat(telemetry): Combine Telemetry hook to send heap event once

### LINKED ISSUE #730: `kedro-telemetry`: Spike to reduce redundant telemetry events

noklam:
## Description

Fix #730 

## Development notes

The test isn't quite working yet, as some test is passing when run individually but fail when I run in a batch. The change is quite big already so I want to get some review:

- `KedroTelemetryCLIHooks` and `KedroTelemetryProjectHooks` are combined to avoid cross hook communication and make it possible to send event ONCE only by storing some intermediate states (handle entrypoint for both `kedro` command or just `session.run()`
- Added Integration test as I think it's important to make sure the event is only sent once, mocking wouldn't work so I need a real project here. For the effort I just create a minimal kedro project. I don't think we should spend too much effort as we moved `kedro-telemetry` back to `kedro` then we can share the same starter project. It's a temporary structure as I don't think adding it to the current `test_plugin.py` fits as the test is more high level.
- I created a fake telemetry ID in case something isn't working we would still capture the fake telemetry from ourselves. 

```toml

[tool.kedro_telemetry]
project_id = "KEDRO_TELEMETRY_TEST"
```

- As discussed, the almost duplicate CLI command event is removed. https://github.com/kedro-org/kedro-plugins/issues/730#issuecomment-2223050576
There are [more detailed conversation internally ](https://github.com/user-attachments/assets/32e19571-f70d-472c-9568-0017008b6778) for this approach.

## Checklist


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -149,28 +149,37 @@ def _generate_new_uuid(full_path: str) -> str:
         return ""
 
 
-class KedroTelemetryCLIHooks:
+class KedroTelemetryHook:
     """Hook to send CLI command data to Heap"""
 
+    def __init__(self):
+        self.consent = None
+        self._sent = False
```
merelcht, line 157: Why is this private and the rest public?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -186,40 +195,47 @@ def before_command_run(
...
                 "Something went wrong in hook implementation to send command run data to Heap. "
                 "Exception: %s",
                 exc,
             )
 
-
-class KedroTelemetryProjectHooks:
-    """Hook to send project statistics data to Heap"""
+    @cli_hook_impl
+    def after_command_run(self):
```
merelcht, line 211: Is this already for the performance issues? #707

kedro-telemetry/pyproject.toml
```diff
@@ -43,10 +43,10 @@ test = [
 ]
 
 [project.entry-points."kedro.cli_hooks"]
-kedro-telemetry = "kedro_telemetry.plugin:cli_hooks"
+kedro-telemetry = "kedro_telemetry.plugin:telemetry_hook"
 
 [project.entry-points."kedro.hooks"]
-kedro-telemetry = "kedro_telemetry.plugin:project_hooks"
+kedro-telemetry = "kedro_telemetry.plugin:telemetry_hook"
```
merelcht, line 49: It looks weird to me that these are now exactly the same.. Can you explain a bit more why it's necessary to combine the hooks?

noklam: The conversation can be found here: https://kedro-org.slack.com/archives/C054U0LJLAC/p1720714959292789

In short. Kedro has two hooks manager, one for cli, the other for the "usual" one. If the hook is set in `settings.py` instead of registered through entrypoint, you will find that only the `@hook_impl` method is invoked but not the `@cli_hook_impl`.

I tried to argue it may make sense to combine these hooks into one. In theory it's nice to have the hook manager separated, but when it comes to actual usage, you need to implement hooks for both and starting to get diverge behavior unless you know that they are handled by two different hook manager(internal details). Or, alternatively, we  keep one `@hook_impl` but each hook manager handle a subset of hookspec (i.e. before_command_run).

^ This is a discussion for longer-term 

You can see that they are handled separately in Kedro

For regular hook entrypoint / settings.py: https://github.com/kedro-org/kedro/blob/1aea4a3d6218ba50c0ae89945bf2b38d751fbfca/kedro/framework/session/session.py#L117-L120

For cli hook: https://github.com/kedro-org/kedro/blob/1aea4a3d6218ba50c0ae89945bf2b38d751fbfca/kedro/framework/cli/hooks/manager.py#L16

noklam: cli_hook: https://github.com/kedro-org/kedro/blob/1aea4a3d6218ba50c0ae89945bf2b38d751fbfca/kedro/framework/cli/hooks/manager.py#L37

hook: https://github.com/kedro-org/kedro/blob/1aea4a3d6218ba50c0ae89945bf2b38d751fbfca/kedro/framework/hooks/manager.py#L68

This is the direct reason why it has to be defined twice in `pyproject.toml`, essentially the same object is being passed to two different entrypoint, and then the corresponding hook manager execute it accordingly.

noklam: I tried to argue it may make sense to combine these hooks into one. In theory it's nice to have the hook manager separated, but when it comes to actual usage, you need to implement hooks for both and starting to get diverge behavior unless you know that they are handled by two different hook manager(internal details). Or, alternatively, we  keep one `@hook_impl` but each hook manager handle a subset of hookspec (i.e. before_command_run).

^ This is a discussion for longer-term 

I would like to simplify this at least in the "front end" where user doesn't need to know the different between these two entrypoints. IMO it's redundant as well, as we can already different CLI specific entrypoint via `@cli_hook_impl` vs `@hook_impl`. Or we can even just tell the different from the name of the method.

kedro-telemetry/kedro_telemetry/plugin.py
noklam: They should be all internal.

kedro-telemetry/kedro_telemetry/plugin.py
noklam: There are 2 performance gains here:
1. Reduce event 
2. Delay the event during execution to `after_command_run` (side-effect).

The 2 is a co-incident. This is actually carefully designed because I need a unified way to ensure event only sent once for both entrypoints (session.run, kedro run and all other kedro run command).

Requirements:
1. The hook need to be executed in the corresponding entrypoint 
  - `after_command_run` won't be triggered for `session.run`, so it is only related to the `kedro run` flow.
2. Avoid duplicate computation, if the info is collected already, don't collect it again.
3. Send event once and only

Scenario 1 - ` kedro run`
1. before_command_run collect basic info
2. after_catalog_created collect more project statistic (avoid collection basic info twice, i.e. project name)
5. after_command_run send the heap event

Scenario 2 - session.run
1. before_command_run is not triggered
2. after_context_created and after_catalog_created are run, in this case the `after_catalog_created` hook will collect both basic info and project statistic. Event are sent immediately (Maybe can even delay this to after_pipeline_run? See also scenario 4)

Scenario 3 - kedro info / kedro ipython
1. before_command_run collect basic info
2. after_command_run send the actual event

Scenario 4 - kedro catalog create (or other catalog related hook)
Same as scenario 2 because event are sent eagerly after catalog created.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -231,18 +247,24 @@ def after_catalog_created(self, catalog):
...
-        )
-        _send_heap_event(
-            event_name="Kedro Project Statistics",
-            identity=user_uuid,
-            properties=project_statistics_properties,
+        enriched_properties = _format_project_statistics_data(
+            event_properties, catalog, default_pipeline, pipelines
         )
+        self.event_properties = enriched_properties
+        if not self._sent:
```
DimedS, line 261: I believe we don't need that check here. The logic should be that `enriched_properties` should be sent in any case. It is our responsibility to ensure the event is not sent again in the `after_command_run()` hook later.

noklam: Yeah I think you are right, since we only have two hooks `after_catalog_created` and `after_command_run` that will send event, and this is the earliest one so in reality there shouldn't be case this is set `True`.

kedro-telemetry/pyproject.toml
merelcht: I read the Slack conversation beforehand, but I was confused by the implementation because it seemed to do the opposite of what was said in Slack? 

> @idanov said: "it doesn't really sit right - single responsibility principle, meaning CLI hooks and Runtime events are better suited to be separate (at least to my eyes) "

ankatiyar:
Generally, the approach makes sense but I think it would be good to wait to switch to `after_command_run` till @lrcouto's analysis on the Kedro side is complete (re #709) - i.e. if the hook is executed even if there's an exception during the command execution. Otherwise we miss out on data about commands run with failures.

DimedS:
> Generally, the approach makes sense but I think it would be good to wait to switch to `after_command_run` till @lrcouto's analysis on the Kedro side is complete (re #709) - i.e. if the hook is executed even if there's an exception during the command execution. Otherwise we miss out on data about commands run with failures.

My opinion is that it would be better to proceed with current approach, but first completely remove the `after_command_run()` hook for clarity. The logic will be clear: if a step includes catalog creation, enhanced data will be sent in the `after_catalog_creation()` hook; otherwise, a simple set of telemetry data will be sent at the last step.

I think it's acceptable to lose information about failed runs for now because we are not currently analysing it properly. I believe it makes sense to implement a special flow for failed runs later, such as collecting error types and analysing them for some insights.

noklam:
@ankatiyar 
The after_command_run change is slightly tricky.

There may be conflicts between the two requirements:
1. Send event even if command fails
2. Make sure event is not sent at the beginning and only sent it after the metadata is enriched

noklam:
I can't approve my own PR, but it looks good to me now :)

DimedS:
As agreed, we have merged `KedroTelemetryCLIHooks` and `KedroTelemetryProjectHooks` into `KedroTelemetryHook`. I refactored the code to improve clarity and reduce repetition by consolidating relevant parts into the new class. Additionally, I updated the tests to reflect this merge.

The current logic is as follows:

1. We collect information about the CLI command and basic project details in the `before_command_run` hook.
2. If the `after_catalog_created` hook is triggered, we add information about nodes and pipelines and then send the event.
3. If step 2 does not occur, we send the event in the `after_command_run` hook.

I attempted to fix the integration tests but was unsuccessful. After discussing with @noklam, I have moved the integration tests to a separate issue and PR ([#771](https://github.com/kedro-org/kedro-plugins/pull/771)).

Please re-review the PR.

kedro-telemetry/kedro_telemetry/plugin.py
ankatiyar: @DimedS does this still need to be updated?

ankatiyar (approved):
LGTM! I think this partially resolves #709 too then!

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: thanks, fixed to private!

Issue #755, opened 2024-07-08
`kedro-datasets`: Investigate `spark.SparkHiveDataset` flaky unit tests 


ankatiyar:
## Description
The tests for `spark.SparkHiveDataset` have been failing intermittently during nightly builds and checks for PRs. Usually re-running them solves the problem but we should investigate the root cause.

## Context
Eg for a failed run - https://github.com/kedro-org/kedro-plugins/actions/runs/9807280152/job/27080744636
Seems to be related to the `SparkSession` setup.

felipemonroy:
Hi @ankatiyar , I have opened a PR with a possible solution. Could you check it please?. I am not sure on how to re-run the actions to test it again.

felipemonroy:
I am not sure how to fix it. I assume some tests use an unintended spark session (like the ones in `test_spark_hive_dataset` using one without hive support). Do we have the option of running `tests/spark` without xdist as the ones in `tests/tensorflow/test_tensorflow_model_dataset.py`?

ankatiyar:
@felipemonroy we can investigate this at our end, if in the meantime your tests are failing on other PRs with this error, tag someone from the team (or me) we can re-run the tests for the PR. Locally, you can run `make tests-no-spark` I think to run all the other unit tests

PR #773, opened 2024-07-20
fix(datasets): Fix spark tests

felipemonroy:
## Description
Spark tests are failing randomly. This could be due that spark sessions are not properly closed, so some tests are using the one that do not correspond. See #755 

## Development notes
I replaced all SparkSession.builder.getOrCreate() calls for the fixture spark session, so we can control which spark session is used. Additionally, in the tear down of the spark session fixtures I added `spark._instantiatedContext = None`, which allegedly ensure that another spark session is created after (see [here](https://stackoverflow.com/a/41512072)).

I am not 100% sure that this is going to fix the issue. Maybe we can rerun the actions a few time just to be sure.

## Checklist

### DIFF
--- a/kedro-datasets/tests/spark/conftest.py
+++ b/kedro-datasets/tests/spark/conftest.py
@@ -27,7 +27,7 @@ def _setup_spark_session():
     ).getOrCreate()
 
 
-@pytest.fixture(scope="module", autouse=True)
+@pytest.fixture(scope="module")
 def spark_session(tmp_path_factory):
     # When running these spark tests with pytest-xdist, we need to make sure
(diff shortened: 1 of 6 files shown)


ankatiyar:
Hey @felipemonroy, first of all, thank you so much for taking this on! 🙏🏾 
I re-ran the unit tests and it failed with the same errors for the spark tests https://github.com/kedro-org/kedro-plugins/actions/runs/10023089947/job/27738930808
If you need any help, let us know

astrojuanlu:
Should we add `.enableHiveSupport()`, like it's done in another fixture?

PR #775, opened 2024-07-22
feat(telemetry): Run telemetry outside of kedro projects

### LINKED ISSUE #729: `kedro-telemetry`: Spike how to enable telemetry for "kedro new" and other commands executed outside of kedro project folder
## Description
Enable telemetry for `kedro new` and other commands executed outside of kedro project folder (e.g. before project creation).

## Context
Currently, `kedro-telemetry` doesn't work with the `kedro new` command. I believe this is due to the following reasons:
1. To participate in telemetry, the `kedro-telemetry` plugin needs to be installed. However, it is only installed after creating the project with the `kedro new` command and running `pip install -r requirements.txt`. But even if the plugin is pre-installed, `kedro new` will not be covered by telemetry.
2. It wasn't possible to provide `consent: false` before the project's folder was created.
3. There were issues in a packaged mode with writing a file in a read-only environment.

With the shift to opt-out consent in #715, and the introduction of new ways to disable telemetry in #728, it would be reasonable to send telemetry in the pre-command hook of any command, including `kedro new`. With opt-out consent, we will stop asking for consent and will avoid problems with writing an answer.

**Important:**
However, we should continue not sending telemetry in packaged mode for the first iteration of the opt-out flow. This is so we can determine more easily what the impact is of changing to an opt-out model. Immediately collecting data for packaged projects would muddle the effect.

So we need to change how we determine the packaged mode. Currently, we use:
```
            if not project_metadata:  # in package mode
                return
```
and it works for package mode but also disables it for `kedro new` commands.

DimedS:
## Description
Closes #729,

Telemetry will be executed if a CLI command is run outside of a Kedro project (including the `kedro new` command). In this case, the same information will be sent, except for the project-related data.

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -271,7 +271,7 @@ def _is_known_ci_env(known_ci_env_var_keys: set[str]):
     return any(os.getenv(key) for key in known_ci_env_var_keys)
 
 
-def _get_project_properties(user_uuid: str, project_path: Path) -> dict:
+def _get_project_properties(user_uuid: str, project_path: Path | None) -> dict:
     if project_path:
         pyproject_path = project_path / PYPROJECT_CONFIG_NAME
         project_id = _get_or_create_project_id(pyproject_path)
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -168,16 +168,15 @@ class KedroTelemetryHook:
     ):
         """Hook implementation to send command run data to Heap"""
 
-        if not project_metadata:  # in package mode
-            return
+        project_path = project_metadata.project_path if project_metadata else None
 
-        self._consent = _check_for_telemetry_consent(project_metadata.project_path)
+        self._consent = _check_for_telemetry_consent(project_path)
         if not self._consent:
             self._opt_out_notification()
             return
 
         # get KedroCLI and its structure from actual project root
-        cli = KedroCLI(project_path=project_metadata.project_path)
+        cli = KedroCLI(project_path=project_path if project_path else Path.cwd())
         cli_struct = _get_cli_structure(cli_obj=cli, get_help=False)
         masked_command_args = _mask_kedro_cli(
             cli_struct=cli_struct, command_args=command_args
@@ -185,9 +184,7 @@ class KedroTelemetryHook:
 
         self._user_uuid = _get_or_create_uuid()
 
-        event_properties = _get_project_properties(
-            self._user_uuid, project_metadata.project_path / PYPROJECT_CONFIG_NAME
-        )
+        event_properties = _get_project_properties(self._user_uuid, project_path)
         event_properties["command"] = (
             f"kedro {' '.join(masked_command_args)}" if masked_command_args else "kedro"
         )
@@ -224,7 +221,7 @@ class KedroTelemetryHook:
 
         if not self._event_properties:
             self._event_properties = _get_project_properties(
-                self._user_uuid, self._project_path / PYPROJECT_CONFIG_NAME
+                self._user_uuid, self._project_path
             )
 
         project_properties = _format_project_statistics_data(
@@ -274,12 +271,16 @@ def _is_known_ci_env(known_ci_env_var_keys: set[str]):
     return any(os.getenv(key) for key in known_ci_env_var_keys)
 
 
-def _get_project_properties(user_uuid: str, pyproject_path: Path) -> dict:
-    project_id = _get_or_create_project_id(pyproject_path)
-    package_name = PACKAGE_NAME or UNDEFINED_PACKAGE_NAME
-    hashed_project_id = (
-        _hash(f"{project_id}{package_name}") if project_id is not None else None
-    )
+def _get_project_properties(user_uuid: str, project_path: Path) -> dict:
+    if project_path:
+        pyproject_path = project_path / PYPROJECT_CONFIG_NAME
+        project_id = _get_or_create_project_id(pyproject_path)
+        package_name = PACKAGE_NAME or UNDEFINED_PACKAGE_NAME
+        hashed_project_id = (
+            _hash(f"{project_id}{package_name}") if project_id is not None else None
+        )
+    else:
+        hashed_project_id = None
 
     properties = {
         "username": user_uuid,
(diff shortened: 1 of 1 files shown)


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -168,26 +168,23 @@ def before_command_run(
     ):
         """Hook implementation to send command run data to Heap"""
 
-        if not project_metadata:  # in package mode
```
noklam, line 171: Does that mean we also track packaged mode now?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -168,26 +168,23 @@ def before_command_run(
...
 
-        self._consent = _check_for_telemetry_consent(project_metadata.project_path)
+        self._consent = _check_for_telemetry_consent(project_path)
         if not self._consent:
             self._opt_out_notification()
             return
 
         # get KedroCLI and its structure from actual project root
-        cli = KedroCLI(project_path=project_metadata.project_path)
+        cli = KedroCLI(project_path=project_path if project_path else Path.cwd())
```
noklam, line 179: I guess my question is, does that `project_path` here actually matters? Where is it used as the `command_args` are coming from hook and `KedroCLI` which was created in `kedro` framework.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -274,12 +271,16 @@ def _is_known_ci_env(known_ci_env_var_keys: set[str]):
     return any(os.getenv(key) for key in known_ci_env_var_keys)
 
 
-def _get_project_properties(user_uuid: str, pyproject_path: Path) -> dict:
-    project_id = _get_or_create_project_id(pyproject_path)
-    package_name = PACKAGE_NAME or UNDEFINED_PACKAGE_NAME
-    hashed_project_id = (
-        _hash(f"{project_id}{package_name}") if project_id is not None else None
-    )
+def _get_project_properties(user_uuid: str, project_path: Path) -> dict:
```
noklam, line 274: type hint need to be fixed

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: I believe that line isn't related to package mode:
https://github.com/kedro-org/kedro-plugins/issues/729#issuecomment-2220467918

As I understand it, in package mode, execution will occur without the CLI. Am I correct?

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: I am also trying to understand why KedroCLI is dependent on a specific Kedro project. I believe the CLI structure should be the same for all projects. Do you think we can just use Path.cwd() here?

kedro-telemetry/kedro_telemetry/plugin.py
merelcht: A packaged project with use the  `__main__.py` as entrypoint. It will then look for the run method to execute. 
I think you're right this line here in telemetry actually isn't related to package mode.. It seems to only be reached if you're not in a kedro project.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -274,12 +271,16 @@ def _is_known_ci_env(known_ci_env_var_keys: set[str]):
...
+    if project_path:
+        pyproject_path = project_path / PYPROJECT_CONFIG_NAME
```
merelcht, line 276: Instead of overwriting the supplied `project_path` here I'd just create a new variable to make it easier to follow what happens and debug if something is wrong.

DimedS: I believe it's already a new variable?

kedro-telemetry/kedro_telemetry/plugin.py
noklam: I don't think the `project_path` matters but we should add a comment to explain this. I have questions about the `_recurse_cli` used by the function, does it takes care of the LazyCommand we added lately?

kedro-telemetry/kedro_telemetry/plugin.py
noklam: Packaged mode will execute the CLI through the entrypoint. I believe the commend about `package mode` because it was assumed that it is the only situation that `project_metadata` will be missing. This is not exactly true now because it would be either:
1. packaged mode
2. kedro command outside of project (i.e. kedro new)

I believe bot situations should be handled.

DimedS: Thank you, @noklam. Could you please provide more details about packaged mode? From what I see in the code, if I package a Kedro project, it will have an entry point `project_name.__main__:main`. When I execute `project_name` in the terminal, the `kedro/framework/cli/project.py:run()` function will be executed directly, without calling the `before_command_run()` hook. This is because that hook is only called in the `KedroCLI()` class during Kedro CLI execution. I'm trying to understand what we need to handle in the telemetry hooks code concerning packaging mode.

kedro-telemetry/kedro_telemetry/plugin.py
DimedS: > I have questions about the `_recurse_cli` used by the function, does it takes care of the LazyCommand we added lately?

I checked the [Lazy Loading PR](https://github.com/kedro-org/kedro/pull/3883), looks like `_recurse_cli()` func should work as previously. Also I tested current PR manually with latest kedro main branch on fresh virtual env, everything works well from the 1st command.

ankatiyar: I think Nok is right, the recursion through the CLI definitely structure triggers the loading of lazy the CLI command groups -
It uses `get_command()` :
https://github.com/kedro-org/kedro-plugins/blob/63a0c91f16b2149d083892788615e51b9a11edf1/kedro-telemetry/kedro_telemetry/masking.py#L44
Which if you look at the implementation of the `LazyGroup`, triggers the lazy load: 
https://github.com/kedro-org/kedro/blob/8bbbfb6c7257005bef841143690308f05d4b829b/kedro/framework/cli/utils.py#L504-L509

I believe we need to find a way to do the masking without loading relying on KedroCLI to not defeat the purpose of introducing lazy loading

noklam (approved):
Approved with notes. https://github.com/kedro-org/kedro-plugins/pull/775#discussion_r1698756126, upon some investigation, I think there are issues to follow up:
1. `project_path` is not arbitary as it affects the result of recursive call
2. The recursive call undo the effort of LazyCommand as it eagerly import all the commands.

I will open a new issue and I suggest we fix both problem together (1. may not needed if we end up throwing away the recursive call)

DimedS:
Thanks for your review, @ankatiyar and @noklam. I agree with you; `recurse_cli()` in telemetry is currently undermining the efforts made by Lazy Load, and we should prioritise fixing that. However, this issue is somewhat loosely connected to the current PR. Therefore, it's better to proceed with this PR before the release to fully complete the opt-out telemetry modernisation.

PR #786, opened 2024-07-26
fix(datasets): mask credentials in ibis dataset when printed

noklam:
## Description

As the `_describe` method will be used more often to print dataset (i.e. notebook). We want to make sure credentials are not printed to the terminal. 

From my understanding the `connection` is unique to individual backend and they don't share specific arguments. The only thing that is common is `backend` so I only expose this in the `describe` method.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -186,7 +186,9 @@ class TableDataset(AbstractDataset[ir.Table, ir.Table]):
             "filepath": self._filepath,
             "file_format": self._file_format,
             "table_name": self._table_name,
-            "backend": self._connection_config.get("backend") if self._connection_config else None,
+            "backend": self._connection_config.get("backend")
+            if self._connection_config
+            else None,
             "load_args": self._load_args,
             "save_args": self._save_args,
             "materialized": self._materialized,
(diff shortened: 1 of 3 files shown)


kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -186,7 +186,7 @@ def _describe(self) -> dict[str, Any]:
             "filepath": self._filepath,
             "file_format": self._file_format,
             "table_name": self._table_name,
-            "connection_config": self._connection_config,
+            "connection_config": self._connection_config.get("backend"),
```
ElenaKhaustova, line 189: From the class constructor `self._connection_config` can be none
```suggestion
            "connection_config": self._connection_config.get("backend") if self._connection_config else None
```

ElenaKhaustova: Shall we also rename `"connection_config"` to `"backend"` or it can be treated as a breaking change?

noklam: I have changed it to `backend` as I think this is small enough to be ignored as a breaking change.

ElenaKhaustova (approved):
Thank you, @noklam, LGTM!

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -186,7 +186,7 @@ def _describe(self) -> dict[str, Any]:
             "filepath": self._filepath,
             "file_format": self._file_format,
             "table_name": self._table_name,
-            "connection_config": self._connection_config,
+            "backend": self._connection_config.get("backend") if self._connection_config else None,
```
deepyaman, line 189: Makes sense, since it avoids constructing the connection! Initially, I was wondering why not just use `self.connection`, but this is better. :)

deepyaman (approved):

Issue #785, opened 2024-07-25
`kedro-telemetry`: Environment var do not override the consent from the `.telemetry` if not command run case


ElenaKhaustova:
## Description
Relates to https://github.com/kedro-org/kedro-plugins/pull/762 and https://github.com/kedro-org/kedro/issues/4024

## Context
We call `_check_for_telemetry_consent` before each command run. However, in the rest of the cases, we call `_check_for_telemetry_consent` after the context is created but we also check if `self._consent` was not set: https://github.com/kedro-org/kedro-plugins/blob/24cd01c23dc36145ff7893f5497df86960a0f68f/kedro-telemetry/kedro_telemetry/plugin.py#L209

It makes it impossible to override the consent from the `.telemetry` by setting the environment variable (`"DO_NOT_TRACK"` or `"KEDRO_DISABLE_TELEMETRY"`) for example in interactive environments.

## Steps to Reproduce
1. create `.telemetry` with `consent: true`
2. `kedro ipython`
3. consent in `.telemetry` is set to true, telemetry is sent (as expected}
4. set `os.environ["DO_NOT_TRACK"] = "true"`
5. `%reload_ext kedro.ipython`
6. consent in `.telemetry` is set to true, `os.environ["DO_NOT_TRACK"]` is true, telemetry is still sent (not expected)

The same issue if changing `.telemetry` in interactive environment:
1. create `.telemetry` with `consent: true`
2. `kedro ipython`
3. consent in `.telemetry` is set to true, telemetry is sent (as expected}
4. update `.telemetry` with `consent: false`
5. `%reload_ext kedro.ipython`
6. consent in `.telemetry` is set to false, telemetry is still sent (not expected)

lrcouto:
On https://github.com/kedro-org/kedro-plugins/pull/791 I tried to change the after_context_created hook to always set `self._consent` insead of only when it's None. It stops further unwanted calls every time we reload IPython with the DO_NOT_TRACK variable set. However, it still sends info to heap after the IPython shell is exited.

If you edit the .telemetry file itself inside IPython, it still sends info to heap once, because when the command was first called consent was true. I think this is a weird behavior. I'm looking into it further.

Issue #781, opened 2024-07-23
`kedro-telemetry` causes error in read-only environments


astrojuanlu:
## Description
Details still TBC, but there are hints that `kedro-telemetry>=0.3.1` can produce an error when running on a read-only environment.

## Context
https://linen-slack.kedro.org/t/22727587/hi-everyone-i-am-new-to-kedro-and-i-am-wondering-how-we-can-#66d97df5-5f59-4c93-a9f1-f68b5cef416f

The user was trying to run a pipeline on Google App Engine like this:

```python
from kedro.framework.session import KedroSession
from kedro.framework.startup import bootstrap_project
from pathlib import Path
bootstrap_project(Path.cwd())
with KedroSession.create(save_on_close=False) as session:
    session.run()
```

(Following this link: https://docs.kedro.org/en/stable/kedro_project_setup/session.html)

cc @oscarvilla

ElenaKhaustova:
Reached out to the user to get some details.

oscarvilla:
For completeness, I'm running the pipeline just with ```kedro run```.

And when I remove telemetry from the requirements.txt passed into the App Engine builder/deployer, the read-only error disappears.

ElenaKhaustova:
Thank you, @oscarvilla!

Some additional context:

1. The user sees a `log.error` message that we output, but the error itself is caught, so it does not affect the execution 
2. That's the expected behaviour for read-only environments; here is how we handle it: https://github.com/kedro-org/kedro-plugins/blob/24cd01c23dc36145ff7893f5497df86960a0f68f/kedro-telemetry/kedro_telemetry/plugin.py#L112
3. In this case, `project_id` is considered as not set

It doesn't look like any immediate fixes are required, but we can consider the following:

1. Change the log lever from error to warning so it doesn't look scary to users.
2. Show it only once for read-only environments now we log it every time we collect the telemetry
3. Do not show it to users at all but log it at telemetry, so we know it's a read-only environment

astrojuanlu:
Agreed with all the points @ElenaKhaustova !

merelcht:
Also flagged by a user: https://github.com/kedro-org/kedro/issues/4066

PR #790, opened 2024-07-29
fix(telemetry): Update README.md for opt-out model

DimedS:
## Description
Update the kedro-telemetry README.md following the transition to an opt-out model for telemetry.

## Checklist

### DIFF
--- a/kedro-telemetry/README.md
+++ b/kedro-telemetry/README.md
@@ -29,30 +29,31 @@ Kedro-Telemetry uses [`pluggy`](https://pypi.org/project/pluggy/) hooks and [`re
 Data collection for telemetry is enabled by default. To withdraw consent, you have a few options:
 
 1. **Set Environment Variables**:
-   Set the environment variables `DO_NOT_TRACK` or `KEDRO_DISABLE_TELEMETRY` to any value. The presence of these environment variables will disable telemetry for all Kedro projects in that environment and will override any consent specified in the `.telemetry` file of the specific project.
+   Set the environment variables `DO_NOT_TRACK` or `KEDRO_DISABLE_TELEMETRY` to any value. The presence of any of these environment variables will disable telemetry for all Kedro projects in that environment and will override any consent specified in the `.telemetry` file of the specific project.
 
 2. **CLI Option When Creating a New Project**:
    When creating a new project, you can use the command:
 
    ```console
-   kedro new --telemetry=yes/no
+   kedro new --telemetry=no
    ```
-   This will create a new project with a `.telemetry` file in its root folder, containing `consent: true/false` accordingly. This file will be used when executing Kedro commands within that project folder. Note that telemetry data about the execution of the `kedro new` command will still be sent if telemetry has not been disabled using environment variables.
+   This will create a new project with a `.telemetry` file in its root folder, containing `consent: false`. This file will be used when executing Kedro commands within that project folder. Note that telemetry data about the execution of the `kedro new` command will still be sent if telemetry has not been disabled using environment variables.
 
    >*Note:* The `.telemetry` file should not be committed to `git` or packaged in deployment. In `kedro>=0.17.4` the file is git-ignored.
 
-3. **Modify or Create the .telemetry File**:
+3. **Modify or Create the `.telemetry` file manually**:
    If the `.telemetry` file exists in the root folder of your Kedro project, set the `consent` variable to `false`. If the file does not exist, create it with the following content:
      ```yaml
      consent: false
      ```
 
-4. **Uninstall the Plugin**:
+4. **Uninstall the plugin**:
    Remove the `kedro-telemetry` plugin:
 
    ```console
    pip uninstall kedro-telemetry
    ```
+   >*Note:* This is a last resort option, as it will break the dependencies of Kedro (for example, `pip check` will report issues).
 
 ## What happens when I withdraw consent?
 
--- a/kedro-telemetry/README.md
+++ b/kedro-telemetry/README.md
@@ -57,11 +57,7 @@ Data collection for telemetry is enabled by default. To withdraw consent, you ha
 
 ## What happens when I withdraw consent?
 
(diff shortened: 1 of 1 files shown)


kedro-telemetry/README.md
```diff
@@ -20,42 +20,39 @@ for further information on the intent of the data collection and what data is co
...
 
-```yaml
-consent: true
-```
+   >*Note:* The `.telemetry` file should not be committed to `git` or packaged in deployment. In `kedro>=0.17.4` the file is git-ignored.
 
->*Note:* The `.telemetry` file should not be committed to `git` or packaged in deployment. In `kedro>=0.17.4` the file is git-ignored.
-
-## How do I withdraw consent?
+3. **Modify or Create the `.telemetry` File**:
```
ElenaKhaustova, line 40:
```suggestion
3. **Modify or Create the .telemetry File**:
```

kedro-telemetry/README.md
```diff
@@ -20,42 +20,39 @@ for further information on the intent of the data collection and what data is co
 For technical information on how the telemetry collection works, you can browse
 [the source code of `kedro-telemetry`](https://github.com/kedro-org/kedro-plugins/tree/main/kedro-telemetry).
 
-## How do I consent to the use of Kedro-Telemetry?
+## How do I withdraw consent?
```
ElenaKhaustova, line 23: I would suggest changing the order of paragraphs and moving **How is the data collected** above the current one, so the withdrawal paragraph is at the bottom.

kedro-telemetry/README.md
```diff
@@ -20,42 +20,39 @@ for further information on the intent of the data collection and what data is co
...
-pip uninstall kedro-telemetry
-```
+   ```console
+   pip uninstall kedro-telemetry
+   ```
 
 ## What happens when I withdraw consent?
 
-Data will only be collected if consent is given. Otherwise, if consent was explicitly denied or withdrawn, the message below will be printed out on every Kedro CLI invocation. If you explicitly deny consent from the beginning, no data will be collected. If you withdraw consent later, the processing of data will be stopped from that moment on.
+Data collection for telemetry is enabled by default. If consent was explicitly denied or withdrawn, the message below will be printed out on every Kedro CLI invocation. If you explicitly deny consent from the beginning, no data will be collected. If you withdraw consent later, the processing of data will be stopped from that moment on.
```
ElenaKhaustova, line 55: > Data collection for telemetry is enabled by default. 

Maybe we should move it to the `How do I withdraw consent?`. Otherwise, it's a bit confusing that we explain how to withdraw consent at first.

kedro-telemetry/README.md
```diff
@@ -20,42 +20,39 @@ for further information on the intent of the data collection and what data is co
...
-## How do I consent to the use of Kedro-Telemetry?
+## How do I withdraw consent?
 
-Kedro-Telemetry is a Python plugin. To install it:
+To withdraw consent, you have a few options:
 
-```console
-pip install kedro-telemetry
-```
+1. **Set Environment Variables**:
```
ElenaKhaustova, line 27: Shall we also add some explanation on what happens if several options are applied at the same time, how they redefine each other?

Relates to: https://github.com/kedro-org/kedro-plugins/issues/785

kedro-telemetry/README.md
ElenaKhaustova, line 55: > the message below will be printed out on every Kedro CLI invocation

This is outside the scope of this PR, but should we print the message only once at the first invocation? Otherwise, it might be annoying for the user.

Relates to: https://github.com/kedro-org/kedro-plugins/issues/781

lrcouto: I agree with Elena here. It would make more sense to move it to the "How do I withdraw consent?" section.

kedro-telemetry/README.md
DimedS: Good catch with that bug. I added a description about the priority of the environment variables and overriding the `.telemetry` file. I believe we will fix the bug.

kedro-telemetry/README.md
DimedS: Thanks, agree with you, done.

kedro-telemetry/README.md
DimedS: Done

kedro-telemetry/README.md
DimedS: I tried to get a clear answer about whether we should print something when telemetry is disabled in the issue: [#727](https://github.com/kedro-org/kedro-plugins/issues/727). However, I haven't received a response yet, so I left it as it is. I think we can create a new issue to discuss this further.

astrojuanlu (commented):
Added some comments in the other PR. I see the text is very similar, maybe we should focus on one of them first.

kedro-telemetry/README.md
ElenaKhaustova: Yep, let's wait until the fix and then revisit this paragraph to make sure they align.

kedro-telemetry/README.md
astrojuanlu: I agree it can be annoying to see that every time. But it feels like too much work to print it "just the first time" (per project? per user?). What about not printing anything if telemetry is disabled?

DimedS: That makes sense. I agree that it's better to remove the notification in this case. If a person has disabled telemetry, they are likely aware that it is no longer active. Additionally, we will stop sending successful telemetry notifications to alleviate any concerns that telemetry is still being sent.

DimedS: Done in https://github.com/kedro-org/kedro-plugins/pull/792, docs updated accordingly in current PR.

ElenaKhaustova (approved):
Thank you, @DimedS, LGTM!

Let's merge after we agree on https://github.com/kedro-org/kedro-plugins/issues/785

astrojuanlu (approved):
Let's go!

Issue #196, opened 2023-04-27
Support parquet and feather datasets from GeoPandas


Calychas:
## Description
Two datasets for feather and parquet formats based on GeoPandas, which will make loading and saving GeoDataFrames with geometries easier.

## Context
When operating on geospatial data GeoPandas is of great use, however at the moment kedro-datasets plugin only supports [GeoJsonDataset](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py). It is possible to load and save parquet and feather files with geometries using standard pandas' datasets, but the geometry data needs special treatment afterwards (e.g. parsing WKB and creating GeoDataFrame manually).

## Possible Implementation
Take an implementation of existing geopandas.GeoJsonDataset, then create geopandas.ParquetDataSet and geopandas.FeatherDataSet. I have an already working implementation privately and I can try to add it here

noklam:
@Calychas Awesome, would you be able to create a PR?

Calychas:
@noklam Yes, I will try to do that sometime in the following 2 weeks

SajidAlamQB:
Awesome, thank you!

noklam:
@Calychas Hey! pinning to see if you still have some bandwidth to add support for this :)

Calychas:
Hey @noklam! Sorry for the delay, just came back to this task

Calychas:
Which dataset would you consider a gold standard for implementation to base the new datasets on? When I look over some of the datasets sometimes I see some minor discrepancies - e.g. using BytesIO or not

noklam:
@Calychas CDVDataSet is probably the most common one.

PR #812, opened 2024-08-21
feat(datasets): Replace geopandas.GeoJSONDataset with geopandas.GenericDataset

harm-matthias-harms:
## Description

Geopandas supports reading parquet and feather files since version 0.8.0. These offer performance improvements over classical file types, such as `.geojson` or `.shp.zip` (supported by `geopandas.GeoJSONDataset`). We have used a private implementation for some time, but want to contribute them to `kedro-datasets`. This PR closes  https://github.com/kedro-org/kedro-plugins/issues/196.

## Development notes

Extended the implementation of `geopandas.GeoJSONDataset` with an optional `file_format`parameter which is needed to use custom `read_*` and `to_*` methods. Replaced the `geopandas.GeoJSONDataset` with `geopandas.GenericDataset`.

## Checklist


harm-matthias-harms:
If the implementation is okay, I can add the feather dataset in the same PR or open a new one.

harm-matthias-harms:
Some pysparks tests are failing. But I think they are unrelated to my changes.

astrojuanlu (commented):
Thanks for this contribution @harm-matthias-harms ! One question: do you think we could have some sort of `geopandas.GenericDataset`, like we do for pandas and Polars? (Just so that we avoid proliferation of datasets)

harm-matthias-harms:
> Thanks for this contribution @harm-matthias-harms ! One question: do you think we could have some sort of `geopandas.GenericDataset`, like we do for pandas and Polars? (Just so that we avoid proliferation of datasets)

To be honest, that would be my preferred solution as well. The current approach feels like a lot of duplicated boilerplate. I really like the Polars dataset and will try to modify the GeoJSON dataset tomorrow to make it work more generally.

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
```diff
@@ -18,6 +19,13 @@
     get_protocol_and_path,
 )
 
+from kedro_datasets.pandas.generic_dataset import (
+    NON_FILE_SYSTEM_TARGETS as PANDAS_NON_FILE_SYSTEM_TARGETS,
+)
+
+# Unofficially GeoPandas also supports pandas io-methods, such as read_csv.
+NON_FILE_SYSTEM_TARGETS = ["postgis"] + PANDAS_NON_FILE_SYSTEM_TARGETS
+
```
harm-matthias-harms, line 28: I don't like this too much, but I don't know if there are better ways.

Officially just documented `geopandas read_*` methods should be supported, but `geopandas.read_csv` also works because geopandas extends pandas...

This tries to rescue the panda's non-file system targets but also adds a dependency to another dataset...

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
```diff
@@ -50,12 +58,13 @@ class GeoJSONDataset(
     """
 
     DEFAULT_LOAD_ARGS: dict[str, Any] = {}
-    DEFAULT_SAVE_ARGS = {"driver": "GeoJSON"}
+    DEFAULT_SAVE_ARGS: dict[str, Any] = {}
```
harm-matthias-harms, line 61: I don't know why this was required, because it also works with shapefiles out of the box. It looks like `fiona` finds the driver automatically by the file name.

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
```diff
@@ -50,12 +58,13 @@ class GeoJSONDataset(
...
 
     def __init__(  # noqa: PLR0913
         self,
         *,
         filepath: str,
+        file_format: str = "file",
```
harm-matthias-harms, line 67: I think it's fine to set a default value, especially because this ensures backward compatibility

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
astrojuanlu: > It looks like fiona finds the driver automatically by the file name.

Most likely!

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
astrojuanlu: What happens if one reads a non-geo file with GeoPandas? Does it return a normal pandas DataFrame?

astrojuanlu (commented):
Thanks a lot @harm-matthias-harms! This looks like a step in the right direction.

My worry is that `GeoJSONDataset` is no longer GeoJSON-exclusive. So I'm wondering if we should more or less take the code you created here, but put it in a `geopandas.GenericGeoDataset`, and leave `geopandas.GeoJSONDataset` alone.

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
harm-matthias-harms: I tried to write a test for this, somehow it didn't worked anymore. Nevermind, I cleaned up the code...

harm-matthias-harms:
> My worry is that GeoJSONDataset is no longer GeoJSON-exclusive. So I'm wondering if we should more or less take the code you created here, but put it in a geopandas.GenericGeoDataset, and leave geopandas.GeoJSONDataset alone.

@astrojuanlu I see two options:
1. GeoJSON is also not GeoJSON exclusive today because I can read many more formats than geojson with `read_file`.
2. We create a `GenericDataset` (this fixes the naming and offers the same functionality and more ) and deprecate the `GeoJSONDataset` to remove it in some future major versions. Then we don't have to maintain the same dataset twice in the future.

Please tell me how you would like to proceed here.

astrojuanlu:
> GeoJSON is also not GeoJSON exclusive today because I can read many more formats than geojson with read_file.

Hah, true... I see this wasn't discussed in https://github.com/kedro-org/kedro/pull/190 originally.

I'm actually in favour of having a new dataset with a more generic name and deprecate the old one, but wondering if the churn is worth the advantages. Any thoughts @merelcht @noklam ?

merelcht:
> I'm actually in favour of having a new dataset with a more generic name and deprecate the old one, but wondering if the churn is worth the advantages. Any thoughts @merelcht @noklam ?

We can just remove the old one and add a new one with a generic name in a breaking release. I don't think it's worth the effort of doing a non-breaking release with a deprecation warning first and then remove the old one in the next breaking release just for this one dataset.

astrojuanlu:
Let's proceed that way then 👍🏼 @harm-matthias-harms are you willing to update the PR accordingly? Namely:

> just remove the old one and add a new one with a generic name in a breaking release.

harm-matthias-harms:
@astrojuanlu I updated the PR. Since I merged main the tests have been failing. It looks like a fsspec problem. I don't know much about ffspec, maybe you have a good idea of how to fix this.

Last successful run: https://github.com/kedro-org/kedro-plugins/actions/runs/10593320929/job/29354563096?pr=812
First run that failed: https://github.com/kedro-org/kedro-plugins/actions/runs/10719091250/job/29722424013?pr=812

kedro-datasets/kedro_datasets/geopandas/geojson_dataset.py
harm-matthias-harms: We don't need this anymore because of backward compatibility, but it's more or less the generic method of geopandas, and it saves some boilerplate in most cases.

astrojuanlu:
I do not know.

Does geopandas handle remote filepaths? Like, if we do `geopandas.read_{self._file_format}("s3://...")` directly without

https://github.com/harm-matthias-harms/kedro-plugins/blob/9e88c5ee0284865b1256a5667bd6f7b691ddb3f7/kedro-datasets/kedro_datasets/geopandas/generic_dataset.py#L155-L156

will it work?

harm-matthias-harms:
@astrojuanlu This was also a thought I had, but that currently makes more tests fail, and it may be better to have all datasets similar. I don't know why the CI fails.

`pytest kedro_datasets/geopandas/generic_dataset.py --doctest-modules --doctest-continue-on-failure` and `pytest tests/geopandas/test_generic_dataset.py` both pass locally with python 3.12. It looks like I have to investigate the CI a bit more. I hope I find some time at the end of the week for that.

astrojuanlu:
Thanks for your patience! Let us know how it goes

merelcht:
> @astrojuanlu This was also a thought I had, but that currently makes more tests fail, and it may be better to have all datasets similar. I don't know why the CI fails.
> 
> `pytest kedro_datasets/geopandas/generic_dataset.py --doctest-modules --doctest-continue-on-failure` and `pytest tests/geopandas/test_generic_dataset.py` both pass locally with python 3.12. It looks like I have to investigate the CI a bit more. I hope I find some time at the end of the week for that.

It has to do with the docstring part and seems to affect python 3.10 specifically. I'm able to replicate the error locally, but haven't found what's wrong yet.

harm-matthias-harms:
@astrojuanlu I figured out the problem... I also bumped geopandas to v1. They use `pyogrio` as the new default over `fiona`. `pyogrio` does not yet support an alternate file system like `fsspec` (https://github.com/geopandas/pyogrio/issues/430). There are workarounds using buffers, but the simplest way would be to change the used engine to `fiona` until `pyogrio` supports `fsspec`.

kedro-datasets/kedro_datasets/geopandas/generic_dataset.py
```diff
@@ -114,28 +129,57 @@ def __init__(  # noqa: PLR0913
...
+        }
 
-        _fs_open_args_save.setdefault("mode", "wb")
-        self._fs_open_args_load = _fs_open_args_load
-        self._fs_open_args_save = _fs_open_args_save
+    def _ensure_file_system_target(self) -> None:
+        # Fail fast if provided a known non-filesystem target
+        if self._file_format in NON_FILE_SYSTEM_TARGETS:
+            raise DatasetError(
+                f"Cannot create a dataset of file_format '{self._file_format}' as it "
```
ElenaKhaustova, line 148: This validation is happening at the `load` and `save` time, won't the error text be confusing at least for `save`?

kedro-datasets/kedro_datasets/geopandas/__init__.py
```diff
@@ -1,12 +1,12 @@
-"""``GeoJSONDataset`` is an ``AbstractVersionedDataset`` to save and load GeoJSON files."""
+"""``AbstractDataset`` implementations that produce geopandas GeoDataFrames."""
```
ElenaKhaustova, line 1:
```suggestion
"""``GenericDataset`` is an ``AbstractVersionedDataset`` to save and load GeoDataFrames."""
```

ElenaKhaustova (commented):
Thank you for the contribution, @harm-matthias-harms!

Left a few nit comments, but overall it looks good!

kedro-datasets/pyproject.toml
```diff
@@ -215,8 +215,9 @@ test = [
     "deltalake>=0.10.0",
     "dill~=0.3.1",
     "filelock>=3.4.0, <4.0",
+    "fiona >=1.8, <2.0",
     "gcsfs>=2023.1, <2023.3",
-    "geopandas>=0.6.0, <1.0",
+    "geopandas>=0.8.0, <2.0",
```
deepyaman, line 220: I'll let others who were already looking into this review, but thank your for this! ❤️

merelcht (approved):
LGTM! Thanks @harm-matthias-harms ⭐

ElenaKhaustova (approved):
Thank you for addressing the comments!

ElenaKhaustova:
@astrojuanlu, do you wanna add anything before we merge?

PR #824, opened 2024-09-04
fix(telemetry): Avoid eager loading the whole KedroCLI for masking

### LINKED ISSUE #794: `kedro-telemetry`: remove recursive logic to avoid eagerly load commands
# Context
image

I found out telemetry is loading all CLI commands which undermine the effort done in https://github.com/kedro-org/kedro/pull/3883.

# Impact
CLI is slow because telemetry always load all the commands.

# Description
At first glance, the recursive call does not seems to be necessary as it collects the entire CLI but they are not consume by telemetry later. Maybe all we need is the related command group. 1. is a small comment but we should check telemetry would work consistently for package mode as well.

> 1. `project_path` is not arbitary as it affects the result of recursive call
> 2. The recursive call undo the effort of LazyCommand as it eagerly import all the commands.
_Originally posted by @noklam in https://github.com/kedro-org/kedro-plugins/pull/775#pullrequestreview-2210638106_

ankatiyar:
## Description

Resolve #794 

## Development notes

* Instead of recursively loading **all the commands** from `KedroCLI` (which are lazy now) to generate a `dict` CLI structure, only load the structure for the command that is called to perform masking.
* I've also updated the test to actually check masking with real Kedro commands since `_mask_kedro_cli` needs the actual `KedroCLI` object.
* Removed the `_recursive_items()` function and associated unit tests since it wasn't being used anywhere. 

## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/masking.py
+++ b/kedro-telemetry/kedro_telemetry/masking.py
@@ -81,16 +81,17 @@ def _get_cli_structure(
     return output
 
 
-def _mask_kedro_cli(
-    cli_struct: dict[str | None, Any], command_args: list[str]
-) -> list[str]:
+def _mask_kedro_cli(KedroCLI, command_args: list[str]) -> list[str]:
     """Takes a dynamic vocabulary (based on `KedroCLI`) and returns
     a masked CLI input"""
     output = []
+    cli_struct = _get_cli_structure(
+        KedroCLI.get_command(ctx=None, cmd_name=command_args[0])
+    )
 
     # Preserve the initial part of the command until parameters sections begin
     arg_index = 0
-    current_CLI = cli_struct.get("kedro", {})
+    current_CLI = cli_struct
     while (
         arg_index < len(command_args)
         and not command_args[arg_index].startswith("-")
--- a/kedro-telemetry/kedro_telemetry/masking.py
+++ b/kedro-telemetry/kedro_telemetry/masking.py
@@ -76,7 +76,8 @@ def _get_cli_structure(
     `click.Context` and return a `dict`.
     """
     output: dict[str | None, Any] = {}
-    _recurse_cli(cli_obj, None, output, get_help)
+    with click.Context(cli_obj) as ctx:  # type: ignore
+        _recurse_cli(cli_obj, ctx, output, get_help)
     return output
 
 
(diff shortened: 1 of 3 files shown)


kedro-telemetry/kedro_telemetry/masking.py
```diff
@@ -81,16 +81,19 @@ def _get_cli_structure(
     return output
 
 
-def _mask_kedro_cli(
-    cli_struct: dict[str | None, Any], command_args: list[str]
-) -> list[str]:
+def _mask_kedro_cli(KedroCLI, command_args: list[str]) -> list[str]:
```
noklam, line 84: Is this a variable name or a actual class?

ankatiyar: variable name, updated it

merelcht (commented):
The implementation looks good! 👍 

Any way to test it end to end?

ankatiyar:
> The implementation looks good! 👍
> 
> Any way to test it end to end?

Hmm, nothing straight-forward. You can print the result of `_get_cli_structure()` to see that the whole CLI dict is not loaded and the result of `_mask_kedro_cli()` if the masking is happening as expected.

ElenaKhaustova:
> > The implementation looks good! 👍
> > Any way to test it end to end?
> 
> Hmm, nothing straight-forward. You can print the result of `_get_cli_structure()` to see that the whole CLI dict is not loaded and the result of `_mask_kedro_cli()` if the masking is happening as expected.

Tested locally and compared outputs from  `_get_cli_structure()` and  `_mask_kedro_cli()` for the current branch and `main` branch  - works as expected 🚀

ElenaKhaustova (approved):
LGTM!

kedro-telemetry/kedro_telemetry/masking.py
```diff
@@ -81,16 +81,19 @@ def _get_cli_structure(
...
     """Takes a dynamic vocabulary (based on `KedroCLI`) and returns
     a masked CLI input"""
     output = []
-
-    # Preserve the initial part of the command until parameters sections begin
     arg_index = 0
-    current_CLI = cli_struct.get("kedro", {})
+    cmd = command_args[0] if command_args else ""
+    if cmd in {"--help", "--version", "-h", "-v", ""}:
+        return command_args
```
DimedS, line 91: Perhaps we only need to return `cmd` instead of the entire `command_args`, to avoid capturing any unnecessary or irrelevant input that might follow.

DimedS (approved):
Works well—tested with HeapAnalytics. Thanks, @ankatiyar, great job! I left one minor comment.

kedro-telemetry/kedro_telemetry/masking.py
ankatiyar: If we return `cmd` here, the masked command would be `kedro ****` even when it's `kedro --help` etc. This sort of masking also takes place when it's a wrong command `kedro typo` is masked as `kedro ****`. Maybe it's worth having that distinction? WDYT?

DimedS: I agree, it makes sense

PR #829, opened 2024-09-10
feat(datasets)!: expose `load` and `save` publicly

### LINKED ISSUE #793: Latest kedro `load` and `save` breaks `mypy` on datasets
## Description

> kedro_datasets/pandas/json_dataset.py:210: error: "AbstractDataset[Any, Any]" has no attribute "_load"; maybe "load"?  [attr-defined]

https://github.com/kedro-org/kedro-plugins/actions/runs/10181415594/job/28161510649?pr=786

The root cause is now dataset cannot find the `_load` method as it's looking for signature of `AbstractDataset`, while the `_load` is implemented in `JSONDataset`

## Potential Fix
Use generic type instead of returing `AbstractDataset` in https://github.com/kedro-org/kedro/blob/8bbbfb6c7257005bef841143690308f05d4b829b/kedro/io/core.py#L403-L407

deepyaman:
## Description
Take advantage of the ability to define public `load` and `save` methods.

Also resolves #793 

## Development notes
This is not compatible with Kedro before 0.19.7.

If Kedro-Datasets required `kedro>=0.18` or something before this, I'd be more concerned about bumping the lower bound; right now, it's just a minor bump.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -21,6 +21,8 @@
 * Fixed incorrect `pandas` optional dependency
 
 ## Breaking Changes
+* Exposed `load` and `save` publicly for each dataset. This requires Kedro version 0.19.7 or higher.
+
 ## Community contributions
 Many thanks to the following Kedroids for contributing PRs to this release:
 * [Brandon Meek](https://github.com/bpmeek)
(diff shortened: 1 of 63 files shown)


noklam (approved):

ElenaKhaustova (approved):

ankatiyar:
Would also request to double check if this change needs to be applied to any new datasets that were merged since the PR was open!

deepyaman:
> Would also request to double check if this change needs to be applied to any new datasets that were merged since the PR was open!

I will set up my dev env, look to update those new datasets, and then merge this. Hopefully get it done this weekend!

Issue #815, opened 2024-08-23
`kedro-telemetry` fix failing to send telemetry when user UUID not found


astrojuanlu:
## Description
If the user UUID is malformed for some reason, telemetry data should still be send, but instead with an "anonymous" or empty UUID.

## Context
How has this bug affected you? What were you trying to accomplish?

## Steps to Reproduce
```
$ cat (python -c "from appdirs import user_config_dir as u; print(u('kedro'))")/telemetry.toml
[telemetry]
uuid = ""
```

## Expected Result
Nothing

## Actual Result

```
$ uv run kedro registry list
[08/23/24 18:13:54] INFO     Using                                                                        __init__.py:249
                             '/private/tmp/kedro-telemetry-test-starter/.venv/lib/python3.12/site-package                
                             s/kedro/framework/project/rich_logging.yml' as logging configuration.                       
[08/23/24 18:13:56] ERROR    Failed to retrieve UUID: badly formed hexadecimal UUID string                   plugin.py:81
- __default__

                    INFO     Kedro is sending anonymous usage data with the sole purpose of improving the   plugin.py:233
                             product. No personal data or IP addresses are stored on our side. If you want               
                             to opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment                 
                             variables, or create a `.telemetry` file in the current working directory with              
                             the contents `consent: false`. Read more at                                                 
                             https://docs.kedro.org/en/stable/configuration/telemetry.html                               
[08/23/24 18:13:57] WARNING  Failed to send data to Heap. Response code returned: 400, Response reason: Bad plugin.py:336
                             Request
```

astrojuanlu:
This issue is about finding out why the Heap request fails. Several things not specifically related to it:

- Having the `INFO` and `WARNING` messages after the error make it more difficult to spot.
  - Maybe the `INFO` should be on top?
  - And, if the request to Heap ever fails, I'm not sure the user cares. Maybe we should make it `DEBUG`
- It was surprising to read "is there a `pyproject.toml` in your cwd?" message when there was indeed one. Maybe we could tweak the hint so it says "is there a `pyproject.toml` with a `[tool.kedro]` table there?"

PR #830, opened 2024-09-10
fix(telemetry): Set default identity value

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro-plugins/issues/815

## Development notes
The `identity` parameter is mandatory when sending data to the heap. Not passing it or setting it to `""` or `None` causes the heap server to return `response code 400`.

The problem was that we used `identity=self._user_uuid`. If we failed to read/generate `user_uuid`, it was treated as an empty string, so passing it to `identity` caused `response code 400` when sending data to the server.

To handle the above case `MISSING_USER_IDENTITY` constant was added, which is used as a default value for the `identity` in case `user_uuid` is `""` or `None`.

## Checklist


noklam (approved):

ankatiyar (approved):
Tested locally!

DimedS (approved):
thank you, @ElenaKhaustova !

PR #838, opened 2024-09-15
Update Lazy Polars to support streaming and in-built parquet support

pascalwhoop:
## Description
Currently playing with Polars and noticing that we don't support streaming which is quite nice for larger than memory datasets. 

*Note* I'm just opening this as a trigger for discussion, I realize this is not yet done and cannot promise I will continue on this since this was just me playing around on the weekend. But hoping I get to play more and bring this into our pipeline (and thus here) :) 

## Development notes
Updated to the latest features of polars which includes streaming=True and leaving the option to the user to do partitioned writing using rust based parquet support

## Checklist


astrojuanlu:
https://github.com/pola-rs/polars/issues/20947 I'd say let's close this until the new streaming engine in Polars is ready

PR #842, opened 2024-09-20
feat(datasets): create separate `ibis.FileDataset`

### LINKED ISSUE #828: Add write support for `ibis.TableDataset` to files
## Description
Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."
Linen not found, so I linked Slack instead: https://kedro-org.slack.com/archives/C03RKPCLYGY/p1724362310850549, requested by user.

Currently `ibis.TableDataset` didn't support write to file but only to tables.

deepyaman:
## Description
Resolves #828 

## Development notes
So far, I copied `ibis.TableDataset`, removing code paths for reading database tables, and adding support for file export.

Just wanted to put this out there for early feedback, but what else should be done?

- [x] Deprecate file I/O in `TableDataset`
- [x] Support versioning (I don't think there's any reason why this wouldn't work out of the box for `FileDataset`)
- [x] Docs stuff (add `FileDataset` to toctree, etc.)

**Update**: Versioning is actually not a trivial subject, because backends don't implement a consistent interface for checking whether a file exists. I plan to do this with PyArrow filesystem, but I will do that in a follow-up PR (to limit complexity added here); this PR handles local versioning.

## Checklist


kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+        if load_args is not None:
+            self._load_args.update(load_args)
+
+        self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+    @property
+    def connection(self) -> BaseBackend:
+        def hashable(value):
```
datajoely, line 135: I think this section needs comments because it's super clever and complicated, but someone else maintaining the class won't be able to understand what's going on without a lot of reverse engineering.

deepyaman: Added a docstring!

deepyaman:
One of the downsides of having a separate `ibis.FileDataset` is that the `connection` cache isn't shared, so you won't reuse an existing connection across `ibis.FileDataset` and `ibis.TableDataset`. Maybe this could be resolved by having the `_connections` class variable tied to a mixin these both inherit from, or, perhaps more simply, taking a `Backend` object and using a custom resolver to pass that.

I can explore this in a future PR; this is probably good enough for now.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+        self._load_args = deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+
+        self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+    @property
+    def connection(self) -> BaseBackend:
```
noklam, line 133: is this exactly the same as table_dataset?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+
+    DEFAULT_LOAD_ARGS: ClassVar[dict[str, Any]] = {}
+    DEFAULT_SAVE_ARGS: ClassVar[dict[str, Any]] = {}
+
+    _connections: ClassVar[dict[tuple[tuple[str, str]], BaseBackend]] = {}
+
+    def __init__(  # noqa: PLR0913
+        self,
+        filepath: str,
+        file_format: str,
```
noklam, line 69: How to find out what options are available? Would it be bad if we default to something like parquet?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+        ``FileDataset`` connects to the Ibis backend object constructed
+        from the connection configuration. The `backend` key provided in
+        the config can be any of the `supported backends <https://ibis-\
+        project.org/install>`_. The remaining dictionary entries will be
+        passed as arguments to the underlying ``connect()`` method (e.g.
+        `ibis.duckdb.connect() <https://ibis-project.org/backends/duckdb\
+        #ibis.duckdb.connect>`_).
+
+        The read method corresponding to the given ``file_format`` (e.g.
+        `read_csv() <https://ibis-project.org/backends/\
```
datajoely, line 89: This is the right URL (I think) - wonder why it wasn't being picked up by CI
https://ibis-project.org/backends/duckdb#ibis.backends.duckdb.Backend.read_csv

datajoely, line 89: This is the right URL (I think) - wonder why it wasn't being picked up by CI
https://ibis-project.org/backends/duckdb#ibis.backends.duckdb.Backend.read_csv

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+        link to the underlying file exists past ``FileDataset.load()``.
+
+        Args:
+            filepath: Path to a file to register as a table. Most useful
+                for loading data into your data warehouse (for testing).
+                On save, the backend exports data to the specified path.
+            file_format: String specifying the file format for the file.
+            table_name: The name to use for the created table (on load).
+            connection: Configuration for connecting to an Ibis backend.
+            load_args: Additional arguments passed to the Ibis backend's
```
datajoely, line 101: So there is an `fsspec` elephant in the room. Does Kedro want to provide fsspec to Ibis or do we delegate this to each individual Ibis backend?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,184 @@
...
+    """``FileDataset`` loads/saves data from/to a specified file format.
+
+    Example usage for the
+    `YAML API <https://docs.kedro.org/en/stable/data/data_catalog_yaml_examples.html>`_:
+
+    .. code-block:: yaml
+
+        cars:
+          type: ibis.TableDataset
+          filepath: data/01_raw/company/cars.csv
```
datajoely, line 25: can we do a `gcs` or `s3` example as well please, and use polars to `read_delta` just to show the power of Ibis

datajoely (changes requested):
This is 99% there, I'd prefer some extra bits of clarity so that @deepyaman 's bus factor isn't as high when it comes to maintainability

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: > is this exactly the same as table_dataset?

@noklam Yes; see https://github.com/kedro-org/kedro-plugins/pull/842#issuecomment-2394011470

deepyaman: > I would like some docstrings here because it's not clear what purpose it serves to me, would make much more maintainable in the future

@datajoely Sounds good, I will add. That said, this is pretty much identical to the pattern for `engine` property in something like `pandas.SQLTableDataset`.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: > How to find out what options are available?

It would be the `to_*` methods in the docs for that backend.

> Would it be bad if we default to something like parquet?

Nah, this seems reasonable to do.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Ideally, this should be delegated to the individual backends, because some of them have fast paths for certain filesystems. The correct fallback is probably to use the PyArrow filesystem implementations.

I briefly wrote about this in the "Development notes" section, as it relates to versioning; perhaps this is something that can address in a follow-up.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Sorry, what's the difference? 😅

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Done!

kedro-datasets/kedro_datasets/ibis/file_dataset.py
datajoely: I thought the linebreak was causing issues - maybe I'm wrong

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -0,0 +1,187 @@
...
+class FileDataset(AbstractVersionedDataset[ir.Table, ir.Table]):
+    """``FileDataset`` loads/saves data from/to a specified file format.
+
+    Example usage for the
+    `YAML API <https://docs.kedro.org/en/stable/data/data_catalog_yaml_examples.html>`_:
+
+    .. code-block:: yaml
+
+        cars:
+          type: ibis.TableDataset
```
deepyaman, line 24:
```suggestion
          type: ibis.FileDataset
```

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Done!

datajoely (approved):
🔥

noklam (approved):
Thank you, it's a nice addition to the ibis datasets

Issue #839, opened 2024-09-17
Error saving `TensorFlowModelDataset` to Azure Blob Storage


gitgud5000:
## Description
I get the following error when trying to save a `TensorFlowModelDataset` to Azure Blob Storage. **The issue occurs only in Azure Blob Storage, not in the local filesystem:**

```vbnet
DatasetError: Cannot save versioned dataset 'model_prestamos.keras' to 
'azureml/modelling/reprecios/data/06_models/run_2024-09-09' because a file with the same name already exists in the
directory. This is likely because versioning was enabled on a dataset already saved previously. Either remove 
'model_prestamos.keras' from the directory or manually convert it into a versioned dataset by placing it in a 
versioned directory (e.g. with default versioning format 
'azureml/modelling/reprecios/data/06_models/run_2024-09-09/model_prestamos.keras/YYYY-MM-DDThh.mm.ss.sssZ/model_pre
stamos.keras').
```
The error indicates that versioning is enabled despite the `versioned: False` setting in the dataset catalog. The file is never created or exists at any point in the Azure Blob Storage.

### Dataset Catalog Definition

```yaml
_azure_base_path: 'abfs://azureml/modelling/reprecios/data'
"{namespace}.model":
  type: ${_datasets.tensormodel}
  credentials: azure_credentials
  versioned: False
  filepath: ${_azure_base_path}/06_models/run_${runtime_params:run_date}/model_{namespace}.keras
  save_args:
    overwrite: True 
``` 

## Steps to Reproduce
**Edit:** Included code below to reproduce. (credit to @merelcht, @astrojuanlu)
```python
from kedro_datasets.tensorflow import TensorFlowModelDataset
import tensorflow as tf
import numpy as np

credentials = {
    'account_name': '',
    'account_key': ''
}

data_set_local = TensorFlowModelDataset(filepath="data/tensorflow_model.keras")
data_set = TensorFlowModelDataset(filepath="abfs://azureml/modelling/data/tensorflow_model.keras",
                                  credentials=credentials)

inputs = tf.keras.Input(shape=(3,))
x = tf.keras.layers.Dense(4, activation=tf.nn.relu)(inputs)
outputs = tf.keras.layers.Dense(5, activation=tf.nn.softmax)(x)
model = tf.keras.Model(inputs=inputs, outputs=outputs)

input_data = np.array([[0.5, 0.3, 0.2]])
predictions = model.predict(input_data)

# This works
data_set_local.save(model)

[...]

PR #844, opened 2024-09-22
fix(datasets): Use `put()` and `get()` instead of `copy` in `TensorFlowModelDataset`'s `_save` and `_load` methods.

gitgud5000:
## Description

This PR resolves the issue in [#839](https://github.com/kedro-org/kedro-plugins/issues/839), where saving `TensorFlowModelDataset` to Azure Blob Storage fails due to incorrect use of `.copy()` from the `fsspec.filesystem` interface.

## Development notes

The issue occurs in the `_save` and `_load` methods on lines 147 and 172, where `.copy()` is used. According to the [fsspec documentation](https://filesystem-spec.readthedocs.io/en/latest/copying.html), `.copy()` is intended for remote-to-remote transfers. Since this involves copying from a local filesystem to remote storage, `.put()` should be used for saving and `.get()` for loading.

The code has been updated to use `.put()` and `.get()` accordingly, replacing the use of `.copy()`. 

Both methods work for local-to-local and local-to-remote(& vice versa) transfers based on testing.
## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -14,13 +14,14 @@
 
 ## Bug fixes and other changes
 * Refactored all datasets to set `fs_args` defaults in the same way as `load_args` and `save_args` and not have hardcoded values in the save methods.
+* Fixed bug related to loading/saving models from/to remote storage using `TensorFlowModelDataset`.
 
 ## Breaking Changes
 ## Community contributions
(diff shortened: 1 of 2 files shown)


merelcht (approved):
Thanks for the investigation and contribution of the fix @gitgud5000 ! This looks good to me 👍 

Can you add a comment in the release notes and add your name to the contributors list for the upcoming release as well? 🙂

PR #845, opened 2024-09-23
ci(airflow): Replace type hints with CatalogProtocol

ankatiyar:
## Description

Fix https://github.com/kedro-org/kedro-plugins/issues/840

## Development notes

Replace `DataCatalog` with `CatalogProtocol`

## Checklist

### DIFF
--- a/kedro-airflow/kedro_airflow/grouping.py
+++ b/kedro-airflow/kedro_airflow/grouping.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from kedro.io import DataCatalog
+from kedro.io import CatalogProtocol
 from kedro.pipeline.node import Node
 from kedro.pipeline.pipeline import Pipeline
 
@@ -11,7 +11,7 @@ def _is_memory_dataset(catalog, dataset_name: str) -> bool:
     return False
 
 
-def get_memory_datasets(catalog: DataCatalog, pipeline: Pipeline) -> set[str]:
+def get_memory_datasets(catalog: CatalogProtocol, pipeline: Pipeline) -> set[str]:
     """Gather all datasets in the pipeline that are of type MemoryDataset, excluding 'parameters'."""
     return {
         dataset_name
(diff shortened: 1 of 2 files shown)


ElenaKhaustova (commented):
@ankatiyar should we also specify kedro version when release? Cause `CatalogProtocol` will be available only in `0.19.9`.

ankatiyar:
> @ankatiyar should we also specify kedro version when release? Cause `CatalogProtocol` will be available only in `0.19.9`.

@ElenaKhaustova Yeah, you're right, let me see if I can make this work with both older and newer versions otherwise we might need to pin airflow plugin to newer versions of Kedro

ankatiyar:
@merelcht @ElenaKhaustova I've made some changes and checked with released Kedro 0.19.8 version as well as the latest `main` version of Kedro that the tests pass and airflow plugin works. The `mypy` check doesn't for older versions complaining that `CatalogProtocol` doesn't exist in `kedro.io` but if I add an ignore statement it complains on CI because we run these tests with the `main` version so I've left it as it is. 

I've also added ignore statements to the `bandit` failures in the experimental dataset (see #850) Will leave that ticket open for the actual solution, but this will unblock the CI.

Would you mind taking another look?

merelcht (approved):
LGTM! 👍

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -21,7 +30,7 @@ def get_memory_datasets(catalog: DataCatalog, pipeline: Pipeline) -> set[str]:
 
 
 def create_adjacency_list(
-    catalog: DataCatalog, pipeline: Pipeline
+    catalog: CatalogProtocol[Any] | DataCatalog, pipeline: Pipeline
```
ElenaKhaustova, line 33: qq - why do we need `CatalogProtocol[Any]` but not `CatalogProtocol`?

ElenaKhaustova (approved):
LGTM, thank you!

Left one non-blocking comment.

kedro-airflow/kedro_airflow/grouping.py
ankatiyar: The original `mypy` error said something like it's getting arguments of the type `CatalogProtocol[Any]` but seems like removing this works just fine with `CatalogProtocol` so i've reverted that

Issue #695, opened 2024-05-23
Review the Supported Python Version Policy  of kedro-datasets


noklam:
@astrojuanlu @merelcht Let's keep this a separate discussion as there are different opinions here. I agree with @deepyaman that there is no rush to drop support when it's perfectly compatible.

This PR will be limited to the badge update only.

_Originally posted by @noklam in https://github.com/kedro-org/kedro-plugins/issues/689#issuecomment-2126820944_

deepyaman:
> Pro:
> 
> * We should drop it as the policy states
> * One less version to support, strictly less burden to maintain.
> 
> Con:
> 
> * Is it necessary to drop 3.9 support when it's compatible and not causing any issue?

Following a Python version support policy (like NEP-29) helps us in a few situations:

* If we get issues on a 3.9 build, we can not investigate, and just shut it off
* If somebody wants to use features or libraries that aren't supported on 3.9, they can do so without concern
* Any 3.9 compatibility code can be removed

It's less important to explicitly prevent people from installing on 3.9, unless you introduce features that won't work on 3.9. I usually use `pyupgrade` and bump the target Python version to make use of newer features, at which point you should prevent installation.

astrojuanlu:
I'm of the opinion that the current policy is fine, and in fact I was about to open an issue about dropping 3.9.

merelcht:
As clarified in https://github.com/kedro-org/kedro-viz/issues/2086#issuecomment-2343330747, `kedro-datasets` can continue with the NEP-29 policy and shouldn't affect `kedro-viz`. So we can close this issue and proceed with #818.

PR #870, opened 2024-10-07
chore: remove support for python 3.9

### LINKED ISSUE #818: Drop Python 3.9 on `kedro-datasets`
## Description
Following NEP 29, let's drop Python 3.9 on `kedro-datasets`. NumPy has already done so. https://numpy.org/doc/stable/release/2.1.0-notes.html

In #695 there was a proposal to review the current policy, but nothing happened. It comes from https://github.com/kedro-org/kedro-plugins/pull/689, where we expressed some reservations about the potential impact on Viz. Some months have passed though, so hopefully it's safe to do now.

felixscherz:
## Description
Hi, this is removing support for python 3.9 and updates the python syntax accordingly. Intended to resolve #818. 

## Development notes
- removed `typing.Union` in favor of `|` operator
- removed `typing.Callable` in favor of `collections.abc.Callable`
- removed python 3.9 from unit test matrix in CI

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,5 +1,6 @@
 # Upcoming Release
 ## Major features and improvements
+* Removed support for Python 3.9
 * Added the following new **experimental** datasets:
 
 | Type                                | Description                                               | Location                                |
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -32,6 +32,7 @@ Many thanks to the following Kedroids for contributing PRs to this release:
 * [janickspirig](https://github.com/janickspirig)
 * [Galen Seilis](https://github.com/galenseilis)
 * [Mariusz Wojakowski](https://github.com/mariusz89016)
+* [Felix Scherz](https://github.com/felixscherz)
 
 
 # Release 4.1.0
--- a/.github/workflows/kedro-datasets.yml
+++ b/.github/workflows/kedro-datasets.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        python-version: [ "3.9", "3.10", "3.11", "3.12" ]
+        python-version: [ "3.10", "3.11", "3.12" ]
     uses: ./.github/workflows/unit-tests.yml
     with:
       plugin: kedro-datasets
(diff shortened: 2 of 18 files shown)


deepyaman:
I think can also simplify https://github.com/felixscherz/kedro-plugins/blob/chore/remove-python-3.9/kedro-datasets/pyproject.toml#L230-L231; not sure if there are other such places.

On a separate note (and this is in no way a reflection on this PR, which resolves the issue very well), I've started thinking more about https://github.com/kedro-org/kedro-plugins/issues/695, especially

> Is it necessary to drop 3.9 support when it's compatible and not causing any issue?

While I was a big proponent of NEP-29, [some experience on another project](https://github.com/ibis-project/ibis/issues/10064) makes me wonder if we should hold off a bit, especially given almost nothing of value is gained through these changes (the typing changes are nice, but don't have any real impact...). I wouldn't be opposed to holding off a major version, unless somebody can point to more bad things we're doing right now to support 3.9.

felixscherz:
Thanks for having a look, good catch with the environment markers! I'll remove the one you referenced which seems to be the only one that deals with python 3.9.
I would agree with you that removing support is not actually necessary as long as there are no major workarounds required and python 3.9 is still officially supported.

kedro-datasets/RELEASE.md
```diff
@@ -1,5 +1,6 @@
 # Upcoming Release
 ## Major features and improvements
+* Removed support for Python 3.9
```
merelcht, line 3: Don't forget to add your name to the contributor list 🙂

merelcht (approved):
Thanks for the contribution @felixscherz ! I'm still in favour of adhering to NEP-29, which is what we agreed on recently in backlog grooming as well. 

@astrojuanlu I'm guessing you also still want to go ahead with this?

kedro-datasets/RELEASE.md
felixscherz: Will do:)

astrojuanlu (approved):
Agreed to continue adhering to NEP 29 indeed! Thanks @felixscherz for this PR 🙌🏼 

Approving to signal my acceptance although naturally the CI will need to be green

felixscherz:
Thanks for the review! I'm having a look at the CI failures, it seems like they are failing on the `main` branch as well.

kevin1kevin1k:
@felixscherz Hey I am the author of https://github.com/kedro-org/kedro/pull/4212.
I noticed that under `.github/workflows` you modified `kedro-datasets.yml`.
According to the error patterns I guess the CI errors will be gone after modifying `kedro-{airflow, docker, telemetry}.yml` as well.

felixscherz:
@kevin1kevin1k thank you! I opened an issue to deal with the removal of python 3.8: #872, we need to remove it from github actions but also make updates to the python syntax and update the `pyrpoject.toml` files, I'm working on those PRs at the moment:)

PR #896, opened 2024-10-17
Add GH action to check for TSC votes on core dataset changes

merelcht:
## Description
Add an automated check to get 1/2 TSC approvals on core dataset changes.

## Development notes
This solution isn't perfect, because it will also trigger on any changes to existing datasets and we only need an official votes for new datasets being contributed to the core datasets. 

I suggest we don't make this a mandatory check, but it will serve as an automatic reminder to check if we need a TSC vote.

## Checklist

### DIFF
--- a/.github/workflows/check-tsc-vote.yml
+++ b/.github/workflows/check-tsc-vote.yml
@@ -1,0 +1,13 @@
+name: Required Reviews
+on:
+  pull_request: {}
+  pull_request_review: {}
+jobs:
+  required-reviews:
+    name: Required Reviews
+    runs-on: ubuntu-latest
+    steps:
+      - name: required-reviewers
+        uses: theoremlp/required-reviews@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


ankatiyar:
This will need to be added to the `merge-gatekeeper.yml` under ignored tests so it doesn't block the merging on other PRs

.github/workflows/check-tsc-vote.yml
```diff
@@ -0,0 +1,13 @@
+name: Required Reviews
+on:
+  pull_request: {}
+  pull_request_review: {}
```
ankatiyar, line 4: I think you can also restrict it to changes in `kedro-datasets/kedro_datasets` - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request

ankatiyar (approved):
Small comment but looks good otherwise!

noklam (approved):

.github/workflows/check-tsc-vote.yml
merelcht: Done!

Issue #748, opened 2024-07-03
Make the APIDataset.save stores the response received 


npfp:
## Description

When running a `POST` or `PUT` request with the `APIDataset`, the response is currently lost while it would be useful to store it.

## Context
We rely a lot on the APIDataset to fetch but also to save data to external API. Keeping tracks or the answer is then really important to us.

## Possible Implementation

We built a custom `APIDataset` that takes a `filepath` argument. If this argument is not None, a `TextDataset(filepath=filepath)` is created and is called in the `_execute_save_request`:

```
self.local_dataset.save(response.text)
```

## Possible Alternatives
Not found any other.

I would then like to make a PR with this proposed change but before making the actual PR, I wanted to double check with you that this feature would be of interest for the community.

merelcht:
It makes total sense to me to save the response. I wouldn't save it as an other type of dataset though (e.g. `TextDataset` mentioned in the description), but rather directly save it to a file format that makes sense. The main reason for that is that to me it feels odd to have one type of dataset be the return type of another dataset.

npfp:
@merelcht I see, regarding the use of `TextDataset` it was really to reduce the maintenance/test burden by relying on a maintained dataset while keeping the interface simple.

When you say 
> directly save it to a file format that makes sense
do you mean use directly the `open` context manager/`write` operation?

MinuraPunchihewa:
@merelcht I am happy to work on this unless @npfp is already in the process.

merelcht:
Sure go for it @MinuraPunchihewa. I see I never got back to the question:

> do you mean use directly the open context manager/write operation?

Yes! My point was mainly that I wouldn't import another kedro dataset like `TextDataset` and return that, but indeed directly use the `open` context manager and `write` operation.

npfp:
For your information, I implemented a custom APIDataset with a nested dataset to store the response.

I didn't go for the open context manager because I wanted to have flexibility of where I would store the responses: locally, or in s3, in json or in raw text, in a versioned way or not, etc.

Also I decided to implement the `load` method when using `POST`, `PUT` mode: this method would just load the response that has been saved, so it can be processed downstream in the pipeline.

I'm aware this design is arguable.

What I could do is to open a PR as the code is pretty much ready and we discuss this there? I could do the PR by the beginning of next week. If it doesn't fit to the kedro spirit then we can just close it.

@merelcht @MinuraPunchihewa What do you think?

[...]

PR #905, opened 2024-10-23
feat(api_dataset): add wrapped dataset to store the response from a POST/PUT request

npfp:
Proof of concept to start discussions about design on how to store and use the `response` object.

## Description
To pursue discussions on how we could store the `response` object from a `POST` or `PUT` api request. 

Relates to https://github.com/kedro-org/kedro-plugins/issues/748

## Development notes
- Only poc code
- Tests to be implemented

## Checklist


kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -222,13 +234,52 @@ def _execute_save_request(self, json_data: Any) -> requests.Response:
...
 
     def _exists(self) -> bool:
         with sessions.Session() as session:
             response = self._execute_request(session)
         return response.ok
+
+    @property
+    def _nested_dataset_type(
+        self,
+    ) -> Type[JSONDataset | PickleDataset | MemoryDataset]:
```
npfp, line 261: There might be a better way to do this.

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -222,13 +234,52 @@ def _execute_save_request(self, json_data: Any) -> requests.Response:
...
+                response: requests.Response = self._execute_save_request(json_data=data)
+
+            if self._wrapped_dataset is None:
+                return response
+            if self._extension == "json":
+                self.wrapped_dataset.save(response.json()) #TODO(npfp): expose json loads arguments
+            elif self._extension == "text":
+                self.wrapped_dataset.save(response.text)
+            elif self._extension:
+                self.wrapped_dataset.save(response)
```
npfp, line 248: Might be better to check on the `self.wrapped_dataset` type:
```suggestion
            elif isinstance(self.wrapped_dataset, JSONDataset):
                self.wrapped_dataset.save(response.json()) #TODO(npfp): expose json loads arguments
            elif  isinstance(self.wrapped_dataset, TextDataset)
                self.wrapped_dataset.save(response.text)
            elif self._extension:
                self.wrapped_dataset.save(response)
```

npfp:
I don't think it's a breaking change of the API dataset: the wrapped dataset args are optionals.

If not provided, the response is not saved as it's already the case.

merelcht:
It's been a while since I looked at the issue this PR is for, but if I understand correctly the problem with APIDataset is that it only returns the response, but doesn't actually save it? And this PR aims to add a wrapper so a user can indicate what format to save the response in? 

On the original issue I said I didn't like the idea of importing other datasets, which I still agree with, but at the same time it does make sense now I see it in code 😜 I'll have to think about this a bit more to review it properly.

npfp:
> It's been a while since I looked at the issue this PR is for, but if I understand correctly the problem with APIDataset is that it only returns the response, but doesn't actually save it? And this PR aims to add a wrapper so a user can indicate what format to save the response in?
> 
> On the original issue I said I didn't like the idea of importing other datasets, which I still agree with, but at the same time it does make sense now I see it in code 😜 I'll have to think about this a bit more to review it properly.

Yes that's right:
 
 * some APIs would return a response that might be important to store somewhere,
 * yes exactly.
 
 Good to know, thx! (I'm a bit biased because we almost always use wrapped dataset for our custom datasets: we're lazy and like relying on a maintained classe :-) )

merelcht (commented):
On second review I think this solution is a pretty good solution. I have left some comments and asked for a second opinion. If we all agree on approach you can polish this and we can do the proper review process 🙂

ElenaKhaustova:
The idea of having such a dataset makes sense to me. Implementation-wise, I would expect to see something similar to https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-6.0.0/_modules/kedro_datasets/partitions/partitioned_dataset.html#PartitionedDataset, so logic of requesting data is separated from saving it via dataset and the extension is derived from dataset set rather than vice a versa.

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -222,13 +236,52 @@ def _execute_save_request(self, json_data: Any) -> requests.Response:
...
+            )
+
+    @property
+    def wrapped_dataset(
+        self,
+    ) -> JSONDataset | PickleDataset | MemoryDataset | None:
+        """The wrapped dataset where response data is stored."""
+        if self._wrapped_dataset is None and self._extension is not None:
+            self._wrapped_dataset = self._nested_dataset_type(
+                **self._wrapped_dataset_args
```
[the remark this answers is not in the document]
npfp: I'm not sure if I understood your question, but as `_nested_dataset_type` returns a dataset class, 

```
self._nested_dataset_type(
                 **self._wrapped_dataset_args
                 )
```
will create the instance of the wrapped dataset:


Also I should change the name of the property `_nested_dataset_type` to `_wrapped_dataset_type` to be consistent...

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -222,13 +236,52 @@ def _execute_save_request(self, json_data: Any) -> requests.Response:
...
         if self._request_args["method"] in ["PUT", "POST"]:
             if isinstance(data, list):
-                return self._execute_save_with_chunks(json_data=data)
-
-            return self._execute_save_request(json_data=data)
+                response: requests.Response = self._execute_save_with_chunks(json_data=data)
+            else:
+                response: requests.Response = self._execute_save_request(json_data=data)
+
+            if self._wrapped_dataset is None:
```
npfp, line 243:
```suggestion
            if self.wrapped_dataset is None:
```

ankatiyar:
Hi team and @npfp, this PR has been open for a while. I'll add the related issue #748 to the backlog and maybe we can pick this PR up again after discussion?

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -185,6 +260,11 @@ def load(self) -> requests.Response:
         if self._request_args["method"] == "GET":
             with sessions.Session() as session:
                 return self._execute_request(session)
+        elif (
```
ElenaKhaustova: This is surprising behaviour for a Kedro dataset:

- load() no longer means “fetch from the API”
- Instead, it means “load whatever was last persisted locally”

We should add `get_response()` (or similar) to retrieve the result instead

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -92,8 +117,9 @@ def __init__(  # noqa: PLR0913
         method: str = "GET",
         load_args: dict[str, Any] | None = None,
         save_args: dict[str, Any] | None = None,
-        credentials: tuple[str, str] | list[str] | AuthBase | None = None,
+        credentials: dict[str, Any] | None = None,
```
ElenaKhaustova: Is there any specific reason to change this? It also contradicts the docstring now.

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -111,18 +137,31 @@ def __init__(  # noqa: PLR0913
...
+                - `JSONDataset`: Stores `response.json()` (parsed JSON data)
+                - `TextDataset`: Stores `response.text` (response body as string)
+                - Other datasets (e.g., `PickleDataset`, `MemoryDataset`): Stores the
+                  full `requests.Response` object
+
+                Can be specified as:
+
+                - A string type identifier: `"json.JSONDataset"`
+                - A dict with `"type"` key: `{"type": "json.JSONDataset", "filepath": "..."}`
+                - A dataset class (advanced usage)
```
ElenaKhaustova, line 165: Do we have tests for this use case?

ElenaKhaustova (commented):
Thank you, @lrcouto, it looks cleaner now.

Made a first pass and left a few comments.

kedro-datasets/kedro_datasets/api/api_dataset.py
rashidakanchwala: +1 or  maybe `get_last_response()` just so it's clear

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -181,12 +257,24 @@ def _execute_request(self, session: Session) -> requests.Response:
...
+        if self._request_args["method"] != "GET":
+            raise DatasetError(
+                "Only GET method is supported for load()."
+                "Use save() to send data or get_last_response() to retrieve "
+                "a persisted response."
+            )
 
-        raise DatasetError("Only GET method is supported for load")
+        with sessions.Session() as session:
+            return self._execute_request(session)
```
ElenaKhaustova, line 289: Relates to another line below, but good to fix

```python
n_chunks = len(json_data) // chunk_size + 1
```

If len(json_data) is an exact multiple of chunk_size, this generates an extra empty chunk.

We should replace with

```python
import math
n_chunks = math.ceil(len(json_data) / chunk_size)
```

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -181,12 +257,24 @@ def _execute_request(self, session: Session) -> requests.Response:
 
         return response
 
+    def get_last_response(self) -> requests.Response:
+        if self._response_dataset is None:
+            raise DatasetError(
+                "No response_dataset configured; cannot retrieve persisted response."
+            )
+
+        return self._response_dataset.load()  # type: ignore[return-value]
```
ElenaKhaustova, line 278: load() could return anything depending on dataset type. We should annotate the return type as `Any` or the expected response type (JSON/dict, str, or requests.Response).

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -62,10 +65,33 @@ class APIDataset(AbstractDataset[None, requests.Response]):
         ... )
         >>> dataset.save(example_table)
 
+        To store API responses, use the `response_dataset` parameter to automatically
```
ElenaKhaustova: Right now, the docstring mentions response dataset and shows examples, but it doesn’t fully explain how the feature behaves.

Can we extend the docstring a little bit, adding some details about how it works? For example:

```python
"""
APIDataset can automatically persist the output of POST/PUT requests via the
`response_dataset` parameter. This is useful for auditing, debugging, or 
reusing API responses downstream in a pipeline.

Behavior based on `response_dataset` type:

- JSONDataset: stores `response.json()` (parsed JSON)
- TextDataset: stores `response.text` (raw text)
- Other datasets (e.g., PickleDataset, MemoryDataset): stores the full `requests.Response` object

You can retrieve the persisted response using:

>>> response = dataset.get_last_response()
"""
```

kedro-datasets/RELEASE.md
```diff
@@ -3,6 +3,7 @@
 ## Bug fixes and other changes
 
 - Fixed `ibis.TableDataset` `exists` method to account for `database` (i.e. the collection of tables, or schema).
+- `api.APIDataset` now stored the response received from a `PUT` or `POST` request via the `response_dataset` parameter.
```
ElenaKhaustova:
```suggestion
- `api.APIDataset` now stores the response received from a `PUT` or `POST` request via the `response_dataset` parameter.
```

PR #971, opened 2024-12-20
feat(datasets): Implement `spark.GBQQueryDataset` for reading data from BigQuery as a spark dataframe using SQL query

abhi8893:
## Description

- The current implementation of `spark.SparkDataset` has rudimentary support for reading data from bigquery using a SQL query.
- Adding this functionality in `spark.SparkDataset` may not comply with `kedro_datasets` design principles as it requires making `filepath` as an optional argument. 
- Further, similar to `pandas.GBQQueryDataset`, the `spark.GBQQueryDataset` is also a `read-only` dataset, hence it's a more suited implementation to maintain the overall design of datasets.

## Development notes

To test the dataset, the following is the manual way:

1. Create a GCP Project (`project_id`)
2. Create a test dataset inside BigQuery as `<project)_id>.<test_dataset>`
3. 2. Create a test materialization dataset inside BigQuery as `<project)_id>.<test_mat_dataset>`
4. Create a test table inside the test dataset as `<project)_id>.<test_dataset>.<test_table>`
5. Create a service account with following permissions:
   - <TODO>
6. Download service account credentials json key

```python
>>> from kedro_datasets.spark import GBQQueryDataset
>>> import pyspark.sql as sql
>>>
>>> # Define your SQL query
>>> sql = "SELECT * FROM `<project)_id>.<test_dataset>.<test_table>`"
>>>
>>> # Initialize dataset
>>> dataset = GBQQueryDataset(
...     sql=sql,
...     materialization_dataset="your_dataset",
...     materialization_project="your_project",  # optional
...     credentials=dict(file="/path/to/your/credentials.json"),
... )
>>>
>>> # Load data
>>> df = dataset.load()
>>>
>>> # Example output
>>> df.show()
```

## Checklist


datajoely:
So this is a great first start - the credentials resolution looks complicated, but also well thought out.

I think we'd need to see some tests for this to go in as is, you can take some inspiration from the [pandas equivalent](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/tests/pandas/test_gbq_dataset.py)

That also being said we could look at contributing this to the experimental part of `[kedro-datasets-experimental](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/CONTRIBUTING.md#experimental-datasets)` which has a lower threshold for admission.

astrojuanlu (commented):
Thank you @abhi8893 , this is very promising! We will look at this soon

abhi8893:
Thanks @datajoely , @astrojuanlu 

###  Credentials Handling

Yes, the credentials handling is a bit different than the rest of the datasets. Also, I do think that `spark.SparkDataset` does not support setting GCP service account credentials, rather always relies on setting credentials for the parsed `filesystem` (`fsspec`). 

Earlier, I have been reading bigquery tables using `spark.SparkDataset` using the following methods of auth:

1. Work on a GCP resource which already has the required authorization via assigned service account
2. OR, download service account json key, and set env var `export GOOGLE_APPLICATION_CREDENTIALS=`/path/to/credentials.json`
3. OR, set the following as spark conf

```yaml
spark.hadoop.google.cloud.auth.service.account.enable: true
spark.hadoop.google.cloud.auth.service.account.json.keyfile: env/gcs_bq_credentials.json
```

With above dataset, I wanted to allow passing credentials directly to the dataset. But it seems, we may have to standardize it a little bit for all other kedro datasets for this GCP case.

### Implementing tests

And let me take a look at how tests can be implemented for this. Initial thoughts: Since this doesn't involve a bigquery client, hence the method of mocking (as in `pandas.GBQQueryDataset` may not be relevant here). 


### Moving to experimental

For moving this to experimental, let me know and I'll lift and shift this to `kedro_datasets_experimental` namespace :)

abhi8893:
Added some tests. 

Related to credentials, one thing that isn't implemented here is reading the SQL query from a text file, which maybe placed anywhere, hence reading the SQL text file itself may require credentials. This is already implemented in `pandas.GBQQueryDataset`.  I haven't implemented providing `filepath` argument here in `spark.GBQQueryDataset`.

So this can necessisate providing 2 types of credentials? Or maybe just one could suffice, and use it authenticate with BigQuery as well as any storage backend?

abhi8893:
So, now I have implemented passing separate credentials for `bigquery` and `filesystem` (since `sql` file maybe placed on any filesystem which may require specific auth). 

I feel the credentials handling is a bit unstandardized across datasets, however I have done what suits this case best. 

Found a few kedro discussions on credentials:

- https://github.com/kedro-org/kedro/wiki/Credentials-in-kedro
- https://github.com/kedro-org/kedro/issues/1646

Part of what makes it tricky is:

- Making sure Python API and YAML API are compatible, and the interface doesn't burdern user with overly specific schema
- Authentication and Authorization is a BIG topic with alot of nuances. I think it's best for kedro to leave the heavylifting of credentials management to third party client libraries, and have as less custom code for this. 
   - Anything more custom should be implemented by the user (but kedro may have guidance on it as it matures more)
- The _where do we keep it_, and _how do we pass it_ dilemma. Often times it isn't clear as to where the credentials should be managed in kedro (`local`, or as environment variable etc) 

Happy to contribute to standardizing credentials handling in kedro as we move forward. However I do realise, this would require quite a bit of research (and user surveys) to perfect.

merelcht:
Hi @abhi8893 , thank you for this contribution! I haven't used BigQuery myself, but generally the code looks fine to me. There's still some failing builds, which will need to pass before we can put this up for a vote. Otherwise, like @datajoely suggested, this could be contributed as experimental, meaning the build checks are less strict.

abhi8893:
@merelcht Finally got around to picking this up back again. Now the test coverage is 100% and all tests pass. If the kedro team takes a vote to put this to experimental, then will shift it there 🙁

merelcht (approved):
I left a couple more minor comments, but otherwise this looks good 👍 

Approving with one caveat: I haven't been able to test this on GCP, but the code is easy to follow and written cleanly.

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+                Alternatively, you can pass the credentials in load_args as follows:
+
+                When passing as `base64`:
+                `load_args={"credentials": "your_credentials"}`
+
+                When passing as a `file`:
+                `load_args={"credentialsFile": "/path/to/your/credentials.json"}`
+
+                When passing as a json object:
+                NOT SUPPORTED
```
ankatiyar, line 110: It says here that JSON credentials are not supported but the code below does accept `json`, is this meant to be updated?

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+            self._sql = None  # type: ignore
+
+        self._materialization_dataset = materialization_dataset
+        self._materialization_project = materialization_project
+        self._load_args = load_args or {}
+        self._bq_credentials = bq_credentials or {}
+
+        self._metadata = metadata
+
+    def _get_spark_bq_credentials(self) -> dict[str, str]:
```
ankatiyar, line 155: Sorry I'm not super familiar with Google Big query but I'm wondering why the credential handling is different from the `pandas.GBQTableDataset`? Would the same way not work?

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
```diff
@@ -0,0 +1,239 @@
+"""``AbstractDataset`` implementation to access Spark dataframes using
+``pyspark``.
```
ravi-kumar-pilla, line 2:
```suggestion
BigQuery Spark connector .
```

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+        ... )
+        >>>
+        >>> # Load data
+        >>> df = dataset.load()
+        >>>
+        >>> # Example output
+        >>> df.show()
+    """
+
+    _VALID_CREDENTIALS_KEYS = {"base64", "file", "json"}
```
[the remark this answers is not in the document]
AbhishekBhatiaQB: Thanks for pointing this out. Refactored to use this in validation / error messages when incorrect keys are passed.

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
AbhishekBhatiaQB: Fixed this! :)

AbhishekBhatiaQB: Made the doc more descriptive. Technically `credentialsFile` is the json "file" which is accepted. I use `json` terminology here to describe directly passing a python dictionary, which is only supported if it's parsed out and `base64encoded`. This is a similar use to other libraries which intermix `json` and `dict` (such as python `requests` library).

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
AbhishekBhatiaQB: The primary difference is that `pandas.GBQTableDataset` use `google.cloud.BigQuery.Client` to handle authentication + connection, while spark has its own specialized BigQuery spark connector. Hence requiring different authentication methods.

kedro-datasets/kedro_datasets/spark/spark_gbq_dataset.py
```diff
@@ -0,0 +1,239 @@
...
+            raise DatasetError(
+                "'sql' and 'filepath' arguments cannot both be empty."
+                "Please provide a sql query or path to a sql query file."
+            )
+
+        if sql:
+            self._sql = sql
+            self._filepath = None
+        else:
+            # TODO: Add protocol specific handling cases for different filesystems.
```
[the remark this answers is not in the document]
AbhishekBhatiaQB: Thanks, removed the TODO. This is already being done and the implementation is very similar to how `pandas.GBQQueryDataset` handles it to let user specify the `sql` as a file, rather than dumping the fully query, which can be very useful to maintain sql files separately in version control, rather than crowding the catalog.

ankatiyar (approved):
Thanks for the explanations @abhi8893! The code looks good to me!

ravi-kumar-pilla (approved):
Thank you @abhi8893

merelcht:
This PR has gone through review, thanks to @ankatiyar and @ravi-kumar-pilla, so now calling for a TSC vote to have this dataset added to the core datasets: @kedro-org/kedro-developers

SajidAlamQB (approved):

kedro-datasets/kedro_datasets/spark/__init__.py
```diff
@@ -10,6 +10,7 @@
 SparkHiveDataset: Any
 SparkJDBCDataset: Any
 SparkStreamingDataset: Any
+GBQQueryDataset: Any
```
deepyaman, line 13: Nit: Can you move this above `SparkDataset` to maintain the existing alphabetical order?

kedro-datasets/kedro_datasets/spark/__init__.py
```diff
@@ -19,5 +20,6 @@
         "spark_hive_dataset": ["SparkHiveDataset"],
         "spark_jdbc_dataset": ["SparkJDBCDataset"],
         "spark_streaming_dataset": ["SparkStreamingDataset"],
+        "spark_gbq_dataset": ["GBQQueryDataset"],
```
deepyaman, line 23: To be consistent, this should be

```suggestion
        "gbq_dataset": ["GBQQueryDataset"],
```

(see `spark.deltatable_dataset`, `pandas.gbq_dataset`)

(and move it right after `deltatable_dataset`, while you're at it)

deepyaman (approved):
Minor change request for consistency, but approving for inclusion in core `kedro-datasets`.

DimedS (approved):

rashidakanchwala (approved):

noklam (commented):
My main question is, what specific is this dataset related to BigQuery? From the implementation there is no bigquery related import.

From my understanding, `SparkDataset` today support `bigquery` as an engine but it doesn't use the `sql` interface. So is this more like a generic Spark SQL Dataset?

abhi8893:
> My main question is, what specific is this dataset related to BigQuery? From the implementation there is no bigquery related import.
> 
> From my understanding, `SparkDataset` today support `bigquery` as an engine but it doesn't use the `sql` interface. So is this more like a generic Spark SQL Dataset?

Hey @noklam , thanks for the review. 

This dataset is still very specific to Bigquery, and is not a generic spark SQL dataset. Reading from other SQL warehouses like Azure Synapse, AWS Redshift, Snowflake will require different mechanisms for reading + authorization. 

`spark.SparkDataset` has a rudimentary support to read data from `bigquery` using a sql query as follows:

```yaml
my_spark_bq_dataset:
  type: spark.SparkDataset
  filepath: None # This is a required argument hence needs to be provided
  load_args:
    query: SELECT * FROM `my_project.my_dataset.my_table`
```

Prior to spark bigquery connector version `0.43.0`, `materializationDataset` also was needed to be set, which `spark.SparkDataset` can't handle natively but `spark.GBQQueryDataset` can. 

In summary, these are the pros for `spark.GBQQueryDataset` vs `spark.SparkDataset`:

1. ✅ Support for providing `sql` query directly as well as through a file (`filepath) placed on any filesystem
2. ✅ Separate handling of filesystem and bigquery credentials
3. ✅ Native bigquery connector support

If this is accepted, it would also make sense to create `spark.GBQTableDataset` in line with the design of `pandas.GBQQueryDataset` and `pandas.GBQTableDataset`

lmk your thoughts on this! 🙁

noklam:
@abhi8893

Thanks for the explanation. I am sorry for the late reply as I missed the GitHub notification.

This makes sense to me and strong enough argument to make gbq standalone for now. We may need to think about this harder in the future as I don't want to end up in the situation that we have two dataset per connectors.

Thank you again for your contribution!

noklam (approved):

astrojuanlu:
Hi everyone! Looks like this PR has several approvals, what do we need to get it over the finish line?

merelcht:
> Hi everyone! Looks like this PR has several approvals, what do we need to get it over the finish line?

@astrojuanlu this is addition to the core datasets, so needs 1/2 TSC approval.

datajoely (approved):

astrojuanlu (approved):

merelcht:
Merging as we have 10/20 approvals.

Issue #759, opened 2023-08-21
Error when saving `TensorFlowModelDataset` as partition


anabelchuinard:
## Description
Can't save TensorFlowModelDataset objects as partition.

## Context
I am dealing with a project where I have to train several models concurrently. I started writing my code using PartitionedDataset where each partition corresponds to the data relative to one training. When I am trying to save the resulting tensorflow  models as a partition, I get an error. I wonder is this has to do with the fact that those inherit from the AbstractVersionedDataset instead of the AbstractDataset. And if yes, I am interested to know if there is any workaround for batch saving those.

This is the instance of my catalog corresponding to the models I want to save:
```
tensorflow_models:
  type: PartitionedDataset
  path: data/derived/ML/models
  filename_suffix: ".hdf5"
  dataset:
    type: kedro.extras.datasets.tensorflow.TensorFlowModelDataset
```
Note: Saving one model (not as partition) works.

## Steps to Reproduce
1. Generate a bunch of trained models 
2. Try to save them in a partition as TensorFlowModelDataset objects

## Expected Result
Should save one .hdf5 file per partition with the name of the file being the associate dictionary key.

## Actual Result
Getting this error:
```
DatasetError: Failed while saving data to data set PartitionedDataset(dataset_config={}, dataset_type=TensorFlowModelDataset,
path=...).
The first argument to `Layer.call` must always be passed.
```
## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): kedro, version 0.18.12
* Python version used (`python -V`): 3.9.16
* Operating system and version: Mac M2

noklam:
> Suggested fix
> Make PartitionedDataset accept only lambda functions for lazy saving and ignore other callable objects - https://github.com/kedro-org/kedro-plugins/pull/978

To me this seems to be a niche case, and changing PartitionedDataset to only accept lambda is a bigger breaking change. Any useful callable will likely be more complicated than a simple lambda. Maybe we can disable lazy loading/saving (default enable) when specified?

DimedS:
Thanks for the investigation and PR, @ElenaKhaustova! I agree with @noklam that relying solely on lambda functions for lazy saving doesn't seem like a generic solution. While it is a breaking change, it's hard to determine how much it will impact users. In my opinion, it would be better to avoid treating all Callables as participants in lazy saving by default. However, this would also be a breaking change. As a simpler alternative, we could provide an option to disable lazy saving, as you suggested.

[...]

PR #978, opened 2025-01-07
fix(datasets): Add parameter to enable/disable lazy saving for `PartitionedDataset`

ElenaKhaustova:
## Description
Fixes https://github.com/kedro-org/kedro-plugins/issues/759

Cause of the issue: https://github.com/kedro-org/kedro-plugins/issues/759#issuecomment-2575562106

Dataset lazy saving docs: https://docs.kedro.org/en/stable/data/partitioned_and_incremental_datasets.html#partitioned-dataset-lazy-saving

## Development notes

1. Added a parameter (`save_lazily`) to enable/disable lazy saving for `PartitionedDataset`. The parameter is enabled by default but users still need to wrap objects with callable as it was required before to apply lazy saving, so the change is not breaking.
2. We didn't add a similar argument for save function as was suggested initially since it contradicts the definition of `AbstractDataset.save`
3. Added callable object saving and loading test
4. Following docs update PR: https://github.com/kedro-org/kedro/pull/4402/

### How to test without TensorFlow
The following code will execute without errors and save two objects if `save_lazily` is set to `False`. The code will fail and `TestSaveCallable.__calll__` will be called if `save_lazily` is set to `True` or not set.

```python
from kedro_datasets.partitions import PartitionedDataset


class TestSaveCallable:
    def __init__(self, a: int):
        print(f"--- Initialized {a} ---")

    def __call__(self, *args, **kwargs):
        print("--- Called ---")


class TestSaveNotCallable:
    def __init__(self, a: int):
        print(f"--- Initialized {a} ---")


def main():
    partitioned_dataset = PartitionedDataset(
        path="data/01_raw/pickle",
        dataset="kedro_datasets.pickle.PickleDataset",
        filename_suffix=".pkl",
        save_lazily=False
    )

    save_dict = {
        "object_1": TestSaveNotCallable(1),
        "object_2": TestSaveCallable(2),
    }

    partitioned_dataset.save(save_dict)


if __name__ == "__main__":
    main()
```

## Checklist


noklam (commented):
I think callable data (that is not intend to be lazy saved) is rare and is not worth to break PartitionedDataset for it. Left the comment in the original issue since I saw two PRs are associated.

https://github.com/kedro-org/kedro-plugins/issues/759#issuecomment-2575783454

merelcht (approved):
Nice and clean solution 👌

ankatiyar (approved):

SajidAlamQB (approved):

tynandebold (approved):

noklam (approved):

rashidakanchwala (approved):

lrcouto (approved):

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
```diff
@@ -152,6 +154,7 @@ def __init__(  # noqa: PLR0913
         load_args: dict[str, Any] | None = None,
         fs_args: dict[str, Any] | None = None,
         overwrite: bool = False,
+        save_lazily: bool = True,
```
Galileo-Galilei, line 157: Why does  `save_lazily` is not a key inside the `save_args` arguments? It seems inconsistent with other datasets.

Galileo-Galilei (commented):
I really like the solution, but a bit confused about the naming choice  which I find inconsistent with other datasets. What's the rationale of not using `save_args` and add save_lazily a key of the dictionary?

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
datajoely: We typically use `save_args` when delegating to an underlying implementation (which we do our best not to alter) e.g. `pd.read_csv(**save_args)`. Since partitioned dataset is a wrapper, the `save_args` would still pertain to the `pd.read_csv(**save_args)` this part if it was a partitioned CSV, the `save_lazily` arg is a Kedro thing not a Pnadas one.

ElenaKhaustova: Another point is that we do not expect any other save arguments since the underlying dataset handles them, so we added `save_lazily` as is rather than creating nesting.

Galileo-Galilei (approved):

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
Galileo-Galilei: Thanks!

Galileo-Galilei: Thanks, makes sense!

Issue #853, opened 2024-09-27
Polars SQL datasets


AntonNikishin:
## Description
It would be great to have Polars implementations of SQLQueryDataset and SQLTableDataset, similar to the Pandas versions: [pandas.SQLTableDataset](https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-4.1.0/api/kedro_datasets.pandas.SQLTableDataset.html) and [pandas.SQLQueryDataset](https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-4.1.0/api/kedro_datasets.pandas.SQLQueryDataset.html).

## Context
Sometimes users would like to read / write polars DataFrames directly from SQL databases.

## Possible Implementation
The datasets will have similar implementation to Pandas versions, but will use polars built-in functions [read_database](https://docs.pola.rs/api/python/stable/reference/api/polars.read_database.html) and [write_database](https://docs.pola.rs/api/python/stable/reference/api/polars.DataFrame.write_database.html).

P.S. I'm happy to work on that ☺️

noklam:
Would the ibis dataset already support polar as a backend?

deepyaman:
> Would the ibis dataset already support polar as a backend?

It does, but:

1. I'm guessing the `read_database` would need to be implemented in Ibis.
2. If a user just wants to use Polars syntax in their nodes,  I guess it's a fair ask.

It's a separate question whether Polars is the best way to manipulate data in a database (definite downside is pulling it into memit. For manipulation, rather than pushing down compute), but a user may still want to do it.

deepyaman:
I would recommend to create `polars.DatabaseDataset` instead of mirroring the pandas datasets, because:

1. Polars provides symmetrical read and write methods.
2. SQL is less explicit,  because Polars SQL is also a thing.

noklam:
Make sense, @AntonNikishin is this something you would like to work on?

deepyaman:
> @noklam I am happy to work on this if it is still open.

@MinuraPunchihewa Pretty sure this is still up for grabs! I'll assign you.

PR #990, opened 2025-01-14
feat(datasets): Added the Experimental PolarsDatabaseDataset

MinuraPunchihewa:
## Description
This PR adds the `PolarsDatabaseDataset` to support interactions with databases using Polars.

Fixes https://github.com/kedro-org/kedro-plugins/issues/853

## Development notes
Quite a bit of code has been copied over from `SQLQueryDataset` in this implementation.

These changes have been tested,

1. Manually, by running the code locally to load and save Polars DataFrames from and to SQLite files.
~~2. Via the existing and newly added unit tests.~~

## Checklist


MinuraPunchihewa:
Hey @noklam, @deepyaman,
I was able to come up with this implementation for the `PolarsDatabaseDataset` by extending `SQLQueryDataset` and it seems to work quite well (at least `load()` does).

Should we implement `save()` as well? This would require a table name to be provided as parameter.

Or do you have different thoughts on how this dataset ought to be implemented?

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+
+import polars as pl
+
+from kedro_datasets.pandas.sql_dataset import SQLQueryDataset, get_filepath_str
+
+
+class PolarsDatabaseDataset(SQLQueryDataset):
+
+    def __init__(  # noqa: PLR0913
+        self,
```
noklam, line 13:
```suggestion
        self,
        *,
```

noklam: To make this keywords-only

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+            query = load_args.pop("sql")
+
+        return pl.read_database(
+            query=query,
+            connection=self._connection_str,
+            **load_args
+        )
+
+    def save(self, data: None) -> NoReturn:
+        pass
```
noklam, line 48: maybe `NotImplemented` instead?

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+            load_path = get_filepath_str(PurePosixPath(self._filepath), self._protocol)
+            with self._fs.open(load_path, mode="r") as fs_file:
+                query = fs_file.read()
+        else:
+            query = load_args.pop("sql")
+
+        return pl.read_database(
+            query=query,
+            connection=self._connection_str,
+            **load_args
```
noklam, line 44: Can you add some docs or checking to reflect this logic
1. If filepath exist use it
2. Otherwise use sql
3. If both are defined, maybe error out or at least log which one is being used?

Ideally I would re-order the argument as well, since the dataset put `sql` as the first argument but actually `filepath` have higher priority which feels counter-intuitive.

noklam (commented):
It'd be great if there is at least one example on how to use this, maybe with a sqlite database to avoid the setup.

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
MinuraPunchihewa: `save()` has also been implemented now.

MinuraPunchihewa:
Hey @noklam and @deepyaman,
I've refactored this by removing the dependency on `pandas.SQLQueryDataset` as @deepyaman suggested. I copied over quite a bit of code from there though since these work quite similarly. I think the duplication can definitely be removed to a large extent by moving some of the common functions to `_utils`. Should I handle this in a different PR since `pandas.SQLQueryDataset` is already a fixed (non-experimental) dataset?
Let me know what you think and I will add the finishing touches to this PR (tests and so forth).

DimedS (approved):
Thanks, @MinuraPunchihewa ! Looks great to me!

astrojuanlu (commented):
Please add to https://github.com/MinuraPunchihewa/kedro-plugins/blob/feature/polars_database_dataset/kedro-datasets/docs/pages/api/kedro_datasets_experimental/index.md so the documentation appears in https://kedro--990.org.readthedocs.build/projects/kedro-datasets/en/990/pages/api/kedro_datasets/ 🙏🏼

astrojuanlu:
I'm interested in this dataset 😄 will try to bring it back to life

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -0,0 +1,298 @@
...
+    >>> import sqlite3
+    >>>
+    >>> from kedro_datasets_experimental.polars import PolarsDatabaseDataset
+    >>>
+    >>> data = pl.DataFrame({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
+    >>> sql = "SELECT * FROM table_a"
+    >>> tmp_path = Path.cwd() / "tmp"
+    >>> tmp_path.mkdir(parents=True, exist_ok=True)
+    >>> credentials = {"con": f"sqlite:///{tmp_path / 'test.db'}"}
+    >>> dataset = SQLQueryDataset(sql=sql, credentials=credentials, table_name="table_a")
```
ElenaKhaustova: Should it be `PolarsDatabaseDataset`?

ElenaKhaustova (approved):
Left a minor comment, the rest LGTM 👍

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
astrojuanlu: Oops yes

merelcht: Fixed

merelcht:
> I'll address @ElenaKhaustova 's comment ☝🏼 but how do I fix the detect-secrets thing?

I've addressed the comment and fixed the detect-secrets check as well.

PR #1031, opened 2025-03-10
feat(airflow): Add grouping by namespace option to kedro-airflow

lrcouto:
## Description

Adds the option to group nodes by namespace to the `kedro-airflow` plugin.

The CLI option `--group-in-memory` is altered to `--group-by`, which can receive the values `memory` or `namespace`. Functionality for grouping by memory should not be altered.

## Checklist


kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -218,6 +226,8 @@ def create(  # noqa: PLR0913, PLR0912
             # topological sort order obtained from pipeline.nodes, see group_memory_nodes()
             # implementation
             nodes, dependencies = group_memory_nodes(context.catalog, pipeline)
+        elif group_by_namespace:
+            nodes, dependencies = group_by_namespace(pipeline)
```
merelcht, line 230: Should this call the method that @ankatiyar added to the framework?

lrcouto: it's calling `Pipeline.grouped_nodes_by_namespace`, I based it on Ankita's prototype

merelcht: Ahh I see now. I guess I didn't expect there still be so much code needed to process the result from `pipeline.grouped_nodes_by_namespace`

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -113,3 +113,37 @@ def dfs(cur_node_name: str, component: int) -> None:
...
+            node_to_namespace[node] = ns
+
+    for ns, nodes in nodes_by_namespace.items():
+        dependent_namespaces = set()
+        for node in nodes:
+            for parent in pipeline.node_dependencies.get(node, []):
+                parent_ns = node_to_namespace.get(parent, "__default__")
+                if parent_ns != ns:
+                    dependent_namespaces.add(parent_ns)
+        dependencies_by_namespace[ns] = list(dependent_namespaces)
```
merelcht, line 147: Why do you need to do this? `pipeline.grouped_nodes_by_namespace` also returns the dependencies for each namespace right? https://github.com/kedro-org/kedro/blob/main/kedro/pipeline/pipeline.py#L398

kedro-airflow/kedro_airflow/plugin.py
ankatiyar, line 230: I get this error: `TypeError: 'bool' object is not callable`
I think it's because the `group_by_namespace` flag (which is boolean) and the function `group_by_namespace()` are called the same thing. Maybe it can be called `group_namespace_nodes()` similar to `group_memory_nodes()` it's also not been imported above

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -113,3 +113,37 @@ def dfs(cur_node_name: str, component: int) -> None:
...
+    """
+    nodes_by_namespace: dict[str, list[Node]] = {}
+    dependencies_by_namespace: dict[str, list[str]] = {}
+
+    grouped_nodes = pipeline.grouped_nodes_by_namespace
+
+    node_to_namespace = {}
+
+    for group_name, group_info in grouped_nodes.items():
+        ns = group_info["name"] if group_info["type"] == "namespace" else "__default__"
```
ankatiyar, line 133: I don't think you need to set a default. Ideally it should be -
- group the namespace 
- individual nodes per task if not namespaced

Which is how the `grouped_by_namespace` property `dict` also returns to. I think the issue with the default would be that it could create a somewhat circular dependencies with the execution. For example when pipeline1 is not namespaced i.e. is default, pipeline2 is namespaced and pipeline3 is not namespaced, the execution order would be impossible to determine

ankatiyar:
Side note, I think `--group-by-namespace` and `--group-in-memory` should not be separate flags by they could be `--group-by=memory/namespace`

kedro-airflow/kedro_airflow/plugin.py
lrcouto: Ah good call, I was trying to solve that.

kedro-airflow/kedro_airflow/grouping.py
lrcouto: That's a question that I had yeah, it didn't follow the behavior of the results that come from the grouped_nodes_by_namespace function but it made sense when I was writing it.

ankatiyar:
The dependencies order here is reversed for some reason in the created DAG file :
Screenshot 2025-03-18 at 15 35 06

lrcouto:
> The dependencies order here is reversed for some reason in the created DAG file : Screenshot 2025-03-18 at 15 35 06

I've noticed it last night, looking into it. Any idea why this might be happening?

lrcouto:
Debugging the function, there's some variation on the order in which dependencies within the dict that comes from `Pipeline.grouped_nodes_by_namespace`. For example:

```
defaultdict(<class 'dict'>, {'f1': {'name': 'f1', 'type': 'node', 'nodes': [Node(identity_one_to_one, 'ds1', 'ds2', 'f1')], 'dependencies': set()}, 'f2': {'name': 'f2', 'type': 'node', 'nodes': [Node(<lambda>, 'ds2', ['ds3', 'ds4'], 'f2')], 'dependencies': {'f1'}}, 'f3': {'name': 'f3', 'type': 'node', 'nodes': [Node(identity_one_to_one, 'ds3', 'ds5', 'f3')], 'dependencies': {'f2'}}, 'f4': {'name': 'f4', 'type': 'node', 'nodes': [Node(identity_one_to_one, 'ds3', 'ds6', 'f4')], 'dependencies': {'f2'}}, 'f5': {'name': 'f5', 'type': 'node', 'nodes': [Node(identity_one_to_one, 'ds4', 'ds8', 'f5')], 'dependencies': {'f2'}}, 'f6': {'name': 'f6', 'type': 'node', 'nodes': [Node(identity_one_to_one, 'ds6', 'ds7', 'f6')], 'dependencies': {'f4'}}, 'f7': {'name': 'f7', 'type': 'node', 'nodes': [Node(<lambda>, ['ds3', 'ds6'], 'ds9', 'f7')], 'dependencies': {'f4', 'f2'}}})
```

On the last `dependencies` entry, the value is `{'f4', 'f2'}`, but sometimes it comes as `{'f2', 'f4'}}`. This is causing some flakiness on tests. I had initially solved this by using the `sorted()` function to make the results predictable. However, this causes the issue that @ankatiyar pointed out, as the dependencies are sorted in an incorrect order. I'm trying to find a way to solve it but if the issue. Any suggestions?

kedro-airflow/kedro_airflow/grouping.py
```diff
@@ -113,3 +113,47 @@ def dfs(cur_node_name: str, component: int) -> None:
...
+                parent_ns = node_to_namespace.get(parent)
+                if (
+                    parent_ns
+                    and parent_ns != ns
+                    and parent_ns not in dependent_namespaces
+                ):
+                    dependent_namespaces.append(parent_ns)
+        dependencies_by_namespace[ns] = dependent_namespaces
+
+    return nodes_by_namespace, dependencies_by_namespace
```
ankatiyar, line 159: The task order still is reversed. I think it's because the second dict `dependecies_by_namespace` is meant to be `{node/namespace: [list of downstream namespaces/nodes]}`  not the other way around. You can check the `group_dependencies` dict from the `group_memory_nodes()` function and how that's constructed.

DimedS (commented):
Thanks for the PR, @lrcouto. I’ve left a comment regarding the API in relation to Kedro. I also think we should replace the DAG execution command with `kedro run --namespace=namespace1` instead of `kedro run --nodes=node1,node2` when the task represents a namespace rather than individual nodes. We can handle that change in a separate PR.

Issue #135, opened 2023-03-21
Rewrite `SparkDataset`


noklam:
## Description

While working on Update SparkDataSet's doc and adding support for spec #128, I found there are multiple bugs that are buried in this dataset, some of the tests are not testing anything.

This dataset is frustrating to work with, as it works very differently from other datasets and there are lots of surprises.

Some observations from #128 :
1.  Is the `hdfs` client useful at all? Some of the tests has not been running correctly for years
2. We don't have anything other than `s3` supported
3. The dataset invents its own `_split_filepath` logic which resembles `get_protocol_and_path` a lot, the only difference is the `hdfs` pathway, is this actually a bug that we should fix in `kedro`?
4. The use of `fs_prefix` and `protocol` is confusing, some of the code is read via `fsspec` some is read with Spark directly. I suspect it will break since there are inconsistencies. It works fine for `s3` and `dbfs` which is the majority of our user base I suspect.
5. Overall it just looks a lot different from other datasets, partly due to the I/O is mainly via `Spark` instead of `fsspec`

## What do we need to do?
1. Ensure we have proper testing - There is a lot of effort in setting up the correct environment to test different pathways. If we ever want to refactor this dataset. Getting the testing process in place is a pre-requisite since this dataset is heavily used.
2. Simplify the logic and have besting testing coverage.

## Why is this important?
Overall there has been quite a lot of bug report related to `SparkDataSet`, it's one of the most popular datasets so making sure it works well consistently is crucial.

The pain point here is we don't have a good test to make sure it will work on Databricks/S3/gcs or even HDFS.

## Context
How has this bug affected you? What were you trying to accomplish?

## Steps to Reproduce
1. [First Step]
6. [Second Step]
7. [And so on...]

## Expected Result
Tell us what should happen.

## Actual Result
Tell us what happens instead.

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`):
* Kedro plugin and kedro plugin version used (`pip show kedro-airflow`):
* Python version used (`python -V`):
* Operating system and version:

merelcht:
Related: https://github.com/kedro-org/kedro-plugins/issues/849

PR #1039, opened 2025-03-14
feat:(datasets): Enrich databricks connect error message

### LINKED ISSUE #1038: SparkDataset failing with databricks-connect serveless cluster
## Description

SparkDataset isn't working with serverless cluster retrieved by databricks-connect. Is there a way to overwrite the way spark session is created in the dataset class?

## Context
I'm trying to run simple kedro pipeline locally. It uses the catalog item of the type `SparkDataset`. In this project I create spark session via databricks-connect as:
```python
DatabricksSession.builder.profile("profile_name").serverless(enabled=True).getOrCreate()
```
But the `get_spark` function inside the dataset does:
```python
DatabricksSession.builder.getOrCreate()
```
And my pipeline fails with: `Cluster id or serverless are required but were not specified`

## Steps to Reproduce
1. Install Databricks-connect
2. Authenticate in the workspace
3. Create serverless executor in the Databricks workspace
4. Add data catalog item `"dataset"` to catalog, with type `SparkDataset` 
5. Define simple pipeline:`pipeline([node(lambda x: x), inputs="dataset", outputs="dataset"])`
6. Try to run kedro pipeline

## Expected Result
The pipeline runs w/o issues

## Actual Result
The pipeline runs with `Cluster id or serverless are required but was not specified`

## Your Environment

Operating system and version: Win 10

```
python==3.10.10
Kedro==0.19.11
kedro-datasets==6.0.0
databricks-connect==16.1.0
databricks-sdk==0.40.0
```

Related to #700

star-yar:
## Description
Resolves #1038 

## Development notes
Catches error raised by databricks-connect and reraises with suggestions on resolution.

## Checklist

- [x] Added tests to cover my changes (_no tests exist for `get_spark` so I'm marking this as done_)


star-yar:
Any ideas why this might fail? Not sure this is caused by my PR, wdyt?

@merelcht

merelcht:
> Any ideas why this might fail? Not sure this is caused by my PR, wdyt?
> 
> @merelcht

No this isn't related to your PR. This failure started showing up on our `main` branch too. We'll have a look at fixing it.

merelcht:
@star-yar, the test failures are resolved, but now there's a coverage issue which is related to the changes you made. Can you add tests to cover the behaviour?

star-yar:
Before implementing such a change, I wanted to share some additional learning.

We should not invoke Databricks Connect here. I think the newer Databricks Connect is intended for establishing the connection once and then retrieving the created session via `pyspark.sql.SparkSesssion`.

So, in the context of a kedro project, the user should create a hook that'll establish the connection via databricks connect. Then, the catalog code should always **get** spark session, not make it, relying on the user creating the connection first.

So maybe we should mention this approach somewhere in docs. And rely on the fact that spark session is created not handling the creation case. This will impact the error message we output.

star-yar:
Any reflections @merelcht ?

merelcht:
Hi @star-yar , I'm really sorry for the late response. 

You raise a very good point. In fact, this is what we recommend when using spark (without databricks) in the docs: https://docs.kedro.org/en/stable/integrations/pyspark_integration.html#initialise-a-sparksession-using-a-hook 

We've had an open issue to rewrite our spark datasets for ever (https://github.com/kedro-org/kedro-plugins/issues/135), but never got round to it so the Spark related datasets have just evolved through contributions but not with a proper architecture in mind. 

We're currently working on a major release, but afterwards we can put this back on our priority list. You're also more than welcome to make a contribution if you have time.

merelcht:
Closing because: https://github.com/kedro-org/kedro-plugins/issues/1038#issuecomment-3738276907

PR #1053, opened 2025-04-08
fix(datasets): Modify test error messages with changes from PR #4580

DimedS:
## Description

Modify dataset test error messages to align with updates introduced in https://github.com/kedro-org/kedro/pull/4580.


ankatiyar (approved):
Thanks @DimedS!

rashidakanchwala (approved):
thank u!!!

PR #1070, opened 2025-05-07
feat(datasets): Fix `PartitionedDataset` caching conflicts in a parallel environment

SajidAlamQB:
## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
+++ b/kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
@@ -263,14 +263,30 @@ class PartitionedDataset(AbstractDataset[dict[str, Any], dict[str, Callable[[], 
             return urlparse(self._path)._replace(scheme="s3").geturl()
         return self._path
 
-    @cachedmethod(cache=operator.attrgetter("_partition_cache"))
-    def _list_partitions(self) -> list[str]:
-        dataset_is_versioned = VERSION_KEY in self._dataset_config
-        return [
-            _grandparent(path) if dataset_is_versioned else path
-            for path in self._filesystem.find(self._normalized_path, **self._load_args)
-            if path.endswith(self._filename_suffix)
-        ]
+    def _list_partitions(self, force_refresh=False) -> list[str]:
+        """List partitions with option to force cache refresh.
+
+        Args:
+            force_refresh: If True, clear the cache before listing partitions.
+                Defaults to False.
+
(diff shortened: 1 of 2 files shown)


ElenaKhaustova (commented):
Thank you, @SajidAlamQB, the solution looks clear to me!

I have a few clarifying questions:
1. Does that mean that we never use cache anymore as we always update it when `load` is called?
> So now every time PartitionedDataset.load() is invoked, it performs a fresh scan of the filesystem to discover partitions.
2. If 1 is true how does it affect the performance?
3. Does it make sense to consider invalidation on `save()` call instead, so if nothing changes we keep reading from cache without rewriting it?

Maybe we can run performance benchmarks (if we have such tests in place) to see how it affects?

SajidAlamQB:
> 1. Does that mean that we never use cache anymore as we always update it when `load` is called?

Yes, specifically for the `_list_partitions` result when `load()` is called, that's correct.

> 2. If 1 is true how does it affect the performance?

It will likely be negligible for datasets with few partitions or on local filesystems. Maybe for datasets with a very large number of partitions could introduce overhead compared to if a previously cached list was reused (But the reuse was the source of the bug).

> 3. Does it make sense to consider invalidation on `save()` call instead, so if nothing changes we keep reading from cache without rewriting it?

That's a good thought. We do already invalidate the cache at the end of the `save()` method (`self._invalidate_caches()` is called there). But the original bug showed that the cache for a `PartitionedDataset` could be populated before any data is saved to its path by an upstream node (e.g., if `exists()` is called), or by a save operation from a different process.

If instance A's cache is populated (e.g., finding no files) and then instance B saves files to that path, instance A's `load()` would still use its old, stale cache without the fix.

By invalidating directly in `load()`, we make sure that the instance performing the load always re-scans. Relying only on `save()` to invalidate wouldn't cover the case where an instance's cache is stale due to some external factor.

> Maybe we can run performance benchmarks (if we have such tests in place) to see how it affects?

That's a very good point. I'm not sure if we have tests for that specifically, but definitely something we should consider.

ElenaKhaustova (approved):
Thank you for the explanation, then I would go on with this solution but consider running/extending the performance tests.

merelcht (approved):
Thanks @SajidAlamQB! This looks good to me and great PR description 👍  Don't forget to add this change to the release notes. 

Let's keep an eye on performance. Can you add a ticket for the benchmark tests and we can prioritise based on user complaints.

PR #1075, opened 2025-05-13
feat(telemetry): Update telemetry to support both catalogs

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4620

## Development notes
Since we decided to keep the old catalog name - `DataCatalog` (https://github.com/kedro-org/kedro/issues/4721) -importing it for both versions will not cause the issue. However, we use the old list API to list datasets, which was deprecated for the new catalog. In this PR, we replace the old API call (list) with the new one (keys) and fallback to the old one in case the new one is not found.

## Checklist

### DIFF
--- a/kedro-telemetry/RELEASE.md
+++ b/kedro-telemetry/RELEASE.md
@@ -1,4 +1,5 @@
 # Upcoming release
+* Updated catalog API usage to comply with both new and old catalogs.
 
 # Release 0.6.2
 * Removed support for Python 3.8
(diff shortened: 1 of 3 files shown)


ravi-kumar-pilla (approved):
This looks good and we will do similar thing on Kedro-Viz too. Thank you @ElenaKhaustova

DimedS (approved):
Thanks, @ElenaKhaustova, LGTM!

merelcht (approved):
👍

PR #1076, opened 2025-05-16
docs(kedro-datasets): Migrate docs to mkdocs

merelcht:
## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Updated `jsonschema/kedro-catalog-X.XX.json` if necessary
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes
- [ ] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)

PR #1078, opened 2025-05-20
chore: Remove mentions of doctest since it doesn't work with mkdocs format

merelcht:
## Description
Follow up on #1076 

## Checklist

### DIFF
--- a/kedro-datasets/CONTRIBUTING.md
+++ b/kedro-datasets/CONTRIBUTING.md
@@ -38,11 +38,10 @@ Core datasets are maintained by the [Kedro Technical Steering Committee (TSC)](h
 
 1. Must be something that the Kedro TSC is willing to maintain.
 2. Must be fully documented.
-3. Must have working doctests (unless complex cloud/DB setup required, which can be discussed in the review).
-4. Must run as part of the regular CI/CD jobs.
-5. Must have 100% test coverage.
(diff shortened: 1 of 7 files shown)


astrojuanlu (commented):
Couldn't we run doctests with https://docs.pytest.org/en/stable/how-to/doctest.html and/or https://pypi.org/project/pytest-doctest-mkdocstrings/ ?

astrojuanlu:
Hmmm okay I just read

> ❗I had to update all docstrings to make them rendered nicely in markdown. However, this now means we can no longer use doc-test to check that the code-examples in the docstrings actually work.

from https://github.com/kedro-org/kedro-plugins/pull/1076

astrojuanlu:
I'm confused, can't mkdocstrings parse Sphinx-style docstrings? https://mkdocstrings.github.io/griffe/reference/docstrings/#docstrings

merelcht:
From what I found the only way to do this is to is to duplicate the code part and have it in a markdown readable way which is with the "```python" opener and then duplicate it for doc-test with the >>> syntax.

astrojuanlu (approved):

merelcht:
Let's discuss this after the styling has been applied to see if there's more we can do. It was (as you can imagine with 60+ datasets) a pretty fiddly job to change this for all datasets, so my suggestion would be to leave it for now and continue if needed after the higher priority things are finished.

Issue #908, opened 2024-10-18
Polars Datatype Catalog Entry Cannot Partition on Saving Parquet


alexdavis24:
## Description
- I would like to save partitioned Polars parquet datasets which currently relies on Pyarrow using write_parquet
  - Following documentation: [https://docs.pola.rs/api/python/version/0.18/reference/api/polars.DataFrame.write_parquet.html](url)
- When running with default system (Rust-based implementation) or with Pyarrow, Kedro returns errors (see below)
- I have encountered this writing locally and writing to s3 
- Attempting to explicitly pass a `filesystem` within catalog (see code below) does not work.
- I believe this is due to handling as a BytesIO object in the save, rather than a direct write of an expected dataframe:
  - https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-4.1.0/_modules/kedro_datasets/polars/lazy_polars_dataset.html#LazyPolarsDataset
  - https://docs.kedro.org/projects/kedro-datasets/en/latest/_modules/kedro_datasets/polars/eager_polars_dataset.html 

## Context
- Trying to partition a large dataframe using polars across a single column

## Steps to Reproduce
1. Sample dataset that runs locally with no issues:
```
df = pl.DataFrame(
    {"A": [1, 2, 3],
        "B": [4, 5, 6]}
)
path = "tmp/test.parquet"
df.write_parquet(
    path,
    use_pyarrow=True,
    pyarrow_options={"partition_cols": ["B"]}
)
# this also runs with no issues
df.write_parquet(
    path,
    partition_by=["B"]
)
```
2. Sample code following the same implementation:
```
# pipelines.py
from kedro.pipeline import Pipeline, node
from kedro.pipeline.modular_pipeline import pipeline
import polars as pl
def my_func():
    return pl.DataFrame(
    {"A": [1, 2, 3],
        "B": [4, 5, 6]}
    )
def create_pipeline() -> Pipeline:
    return pipeline(
    node(
                func=my_func,
                inputs={}
                outputs="my_entry",
                name="partition_polars"        
    )
```

```
# catalog.yml
# using Rust
my_entry:
  # also tried with polars.LazyPolarsDatset
  type: polars.EagerPolarsDataset 
  filepath: /tmp/test.parquet
  file_format: parquet
  save_args:
    partition_by: 
      - B
```
```
# catalog.yml
# using pyarrow (C++)
my_entry:
  type: polars.EagerPolarsDataset
[...]

PR #1087, opened 2025-05-28
fix(datasets): LazyPolarsDataset save/load for partitioned Parquet with `use_pyarrow=True`

gitgud5000:
Implemented fallback to PyArrow for saving partitioned Parquet files in `LazyPolarsDataset`, addressing issue #908. This change ensures compatibility with object storage systems and maintains consistency with Polars' capabilities. 

`EagerPolarsDataset` remains unchanged.

## Description

`LazyPolarsDataset` previously did not support saving partitioned Parquet files when `use_pyarrow=True`. Polars defers to `pyarrow.dataset.write_dataset()` under the hood when using `partition_cols`, which writes a directory of files—not a single file.

However, the existing implementation used `self._fs.open(...)` to pass a file-like object, whereas PyArrow expects a filesystem object that represents a folder-like destination. This mismatch caused errors...

This PR fixes that by injecting a proper `pyarrow.PyFileSystem` (wrapping the `fsspec` filesystem) into `pyarrow_options["filesystem"]`.

Additionally, partitioning information in `load_args` (e.g., `"partitioning": "hive"`) was ignored on remote file systems, resulting in incorrect reads or errors

## Development notes

- In `load()`: 
  - Dropped `"partitioning"` from `self._load_args` when using Polars’ native scanner (local FS).
  - For non-local filesystems, the dataset is loaded via `pyarrow.dataset()`, which correctly handles partitioned folders using the provided `partitioning` arg.
- In `save()`: 
  - If `use_pyarrow=True`, wraps the target fsspec filesystem using `pyarrow.PyFileSystem(FSSpecHandler(...))` and injects it into `pyarrow_options["filesystem"]`, allowing Polars to delegate partitioned writes to Arrow.

This was tested against both local filesystems and Azure Blob Storage to confirm that partitioned read/write works as expected and resolves the original and replicated issues.

Here’s my comment for a [detailed explanation](https://github.com/kedro-org/kedro-plugins/issues/908#issuecomment-2848982838)

## Checklist


ankatiyar:
Hi @gitgud5000, thanks for the fix! The code looks good to me, any chance you could add a unit test for this as well?

gitgud5000:
@ankatiyar Sorry for the delay in getting back to you! I'm happy to add tests.

Could you clarify what level of coverage you're expecting and where these should be added? This will be my first time writing tests in this codebase, but I'm eager to dive in.

ravi-kumar-pilla:
Hi @gitgud5000 , Thank you for your contribution. We would like to take this further and add the required tests to get this in next release. I will be adding tests today.

PR #1093, opened 2025-05-31
feat(datasets): `ibis.TableDataset` add configurable save mode for table materialization

### LINKED ISSUE #834: Support for inserting data in ibis.TableDataset
## Description
In ETL pipelines, loading transformed data into various data warehouses is a critical requirement. Currently, the `ibis.TableDataset ` connector in Kedro does not support data insertion into Ibis backends. 

### Why is this change important to me?
We are developing ETL pipelines in our organization, and inserting records into data warehouses is an essential requirement. At present, without support for data insertion, we must bypass the Kedro `DataCatalog` and rely on external ORM tools to handle native data storage operations, such as `SQLAlchemy` , `dataset` etc .

### How would I use it?
Supporting data insertion in `ibis.TableDataset` would allow us to maintain a clean and consistent pipeline, avoiding the need for custom load operations within nodes. This would simplify the workflow and allow Kedro to manage the complete I/O process.

### How can it benefit other users?
By enabling this feature, users could avoid writing custom loading logic, thereby keeping their pipelines cleaner and more efficient. This would enhance Kedro's usability in scenarios where heavy I/O operations are involved, particularly for teams working with data warehouses or similar storage backends.

gitgud5000:
feat: Introduces a `mode` parameter for save operations, allowing "append", "overwrite", "error", and "ignore" options to control write behavior.
  - Supports legacy `overwrite` flag with backward compatibility and deprecation warning.
  - Adds mode dispatching logic to handle different write scenarios.
  - Updates examples and docs to reflect new parameter.
  - Prevents simultaneous use of both `mode` and legacy `overwrite` to avoid ambiguity.
  - docs: Improves documentation and usage examples for new save modes.

Semi-breaking change: Replaces `overwrite` save argument with `mode`; users should update configurations to use `mode`.

## Description

This PR adds a configurable `mode` parameter to `ibis.TableDataset.save()`, allowing users to control table materialization behavior. Supported modes are:

- `"append"`: Insert data into an existing table (requires the backend to implement `insert()`).
- `"overwrite"`: Drop and recreate the table/view.
- `"error"`: Fail if the table already exists.
- `"ignore"`: Do nothing if the table exists; otherwise create it.

It preserves backward compatibility by mapping the legacy `overwrite: True/False` flag to `mode="overwrite"` or `mode="error"`, and raises a `ValueError` if both `mode` and the deprecated `overwrite` are specified simultaneously.

## Development notes

- Closes #834 (requested by @deepyaman on Sep 14, 2024, to add “insert” support via a `mode` argument). 
  - In the parent issue kedro-org/kedro#4190, the main ask was to allow an “append”/“insert” operation for Ibis datasets.
  - This implementation covers:
    - Rework the API akin to Spark’s `DataFrameWriter.mode` and Pandas’s `to_csv(mode=…)`.
    - Backward compatibility with `overwrite` behavior.
    - Clear, explicit behavior for each mode option.

<details>
  <summary>Expand for details:</summary>

1. **`DEFAULT_SAVE_ARGS`** was updated:
   - Removed the boolean `overwrite` default.
   - Introduced `"mode": "overwrite"` as the new default.

2. **`__init__` logic**:
   - Accepts `save_args={"mode": ...}` (must be one of `"append"|"overwrite"|"error"|"ignore"`).
   - Detects deprecated `overwrite` in `save_args`, mapping it to `"mode"` internally.
   - Raises `ValueError` if both `mode` and `overwrite` are specified.

3. **`save()` dispatch**:
   - If `mode == "append"`, attempts `self.connection.insert(self._table_name, data, **save_args)`.
     - Raises `NotImplementedError` if `insert()` is not supported by the Ibis backend.
   - If `mode` is `"overwrite"|"error"|"ignore"`, calls `create_{table_or_view}` with appropriate `overwrite=...` flag and existence checks.

4. **Docstrings and examples**:
   - Updated YAML and Python usage samples to include `mode` arg.
   - Clarified that `mode` supersedes `overwrite`.

5. **`_describe()`** now returns `"mode": <current_mode>` so it appears in Kedro’s dataset description.
</details>

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/ibis/table_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/table_dataset.py
@@ -26,6 +26,7 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
         database: company.db
       save_args:
         materialized: table
+        mode: append
 
     motorbikes:
       type: ibis.TableDataset
@@ -33,6 +34,9 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
       connection:
         backend: duckdb
         database: company.db
+      save_args:
+        materialized: view
+        mode: overwrite
     ```
 
     ### Example usage for the [Python API](https://docs.kedro.org/en/stable/data/advanced_data_catalog_usage.html):
@@ -45,7 +49,7 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
     dataset = TableDataset(
         table_name="test",
         connection={"backend": "duckdb", "database": tmp_path / "file.db"},
-        save_args={"materialized": "table"},
+        save_args={"materialized": "table", "mode": "overwrite"},
     )
     dataset.save(data)
     reloaded = dataset.load()
@@ -60,7 +64,7 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
     DEFAULT_LOAD_ARGS: ClassVar[dict[str, Any]] = {}
     DEFAULT_SAVE_ARGS: ClassVar[dict[str, Any]] = {
         "materialized": "view",
-        "overwrite": True,
+        "mode": "overwrite",
     }
 
     _CONNECTION_GROUP: ClassVar[str] = "ibis"
@@ -107,7 +111,9 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
                 `create_{materialized}` method. By default, ``ir.Table``
                 objects are materialized as views. To save a table using
                 a different materialization strategy, supply a value for
-                `materialized` in `save_args`.
+                `materialized` in `save_args`. You can also include a
+                `mode` parameter ("append", "overwrite", "error", "ignore")
+                to control write behavior.
             metadata: Any arbitrary metadata. This is ignored by Kedro,
                 but may be consumed by users or external plugins.
         """
@@ -119,19 +125,30 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
 
         # Set load and save arguments, overwriting defaults if provided.
         self._load_args = deepcopy(self.DEFAULT_LOAD_ARGS)
-        if load_args is not None:
+        if load_args:
             self._load_args.update(load_args)
         if database is not None:
             self._load_args["database"] = database
 
         self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
-        if save_args is not None:
+        if save_args:
             self._save_args.update(save_args)
         if database is not None:
             self._save_args["database"] = database
 
         self._materialized = self._save_args.pop("materialized")
 
+        # Handle mode / overwrite conflict
+        if save_args and "mode" in save_args and "overwrite" in self._save_args:
+            raise ValueError(
+                "Cannot specify both 'mode' and deprecated 'overwrite'."
+            )
+        # Map legacy overwrite if present
+        if "overwrite" in self._save_args:
+            legacy = self._save_args.pop("overwrite")
+            self._mode = "overwrite" if legacy else "error"
+        else:
+            self._mode = self._save_args.pop("mode")
     def _connect(self) -> BaseBackend:
         import ibis
 
@@ -148,8 +165,38 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
         return self.connection.table(self._table_name, **self._load_args)
 
     def save(self, data: ir.Table) -> None:
+        if self._mode == "append":
+            if hasattr(self.connection, "insert"):
+                self.connection.insert(
+                    self._table_name, data, **self._save_args
+                )
+            else:
+                raise NotImplementedError(
+                    f"Insert mode is not supported by the {self.connection!r} backend."
+                )
+            return
+
         writer = getattr(self.connection, f"create_{self._materialized}")
-        writer(self._table_name, data, **self._save_args)
+
+        def overwrite():
+            writer(self._table_name, data, overwrite=True, **self._save_args)
+        def error():
+            writer(self._table_name, data, overwrite=False, **self._save_args)
+        def ignore():
+            if self._exists():
+                return
+            writer(self._table_name, data, overwrite=False, **self._save_args)
+
+        mode_dispatch = {
+            "overwrite": overwrite,
+            "error": error,
+            "ignore": ignore,
+        }
+
+        try:
+            mode_dispatch[self._mode]()
+        except KeyError:
+            raise ValueError(f"Unknown mode: {self._mode!r}")
 
     def _describe(self) -> dict[str, Any]:
         load_args = deepcopy(self._load_args)
@@ -163,6 +210,7 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
             "load_args": load_args,
             "save_args": save_args,
             "materialized": self._materialized,
+            "mode": self._mode,
         }
 
     def _exists(self) -> bool:


kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -107,7 +112,9 @@ def __init__(  # noqa: PLR0913
                 `create_{materialized}` method. By default, ``ir.Table``
                 objects are materialized as views. To save a table using
                 a different materialization strategy, supply a value for
-                `materialized` in `save_args`.
+                `materialized` in `save_args`. You can also include a
+                `mode` parameter ("append", "overwrite", "error", "ignore")
```
deepyaman, line 116: It would be nice to explain these, as done in https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.mode.html#pyspark.sql.DataFrameWriter.mode.

I wonder if "append" or "insert" makes more sense (or if should just support both). As I think about it more, maybe "append" is good, because it's clear it doesn't overwrite (e.g. [dbt supports "append" and "insert_overwrite", among other things](https://docs.getdbt.com/docs/build/incremental-strategy)).

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -148,8 +165,38 @@ def load(self) -> ir.Table:
...
+            writer(self._table_name, data, overwrite=False, **self._save_args)
+
+        mode_dispatch = {
+            "overwrite": overwrite,
+            "error": error,
+            "ignore": ignore,
+        }
+
+        try:
+            mode_dispatch[self._mode]()
```
deepyaman, line 197: Why implement this as a dispatch? It feels a bit convoluted—define function, create dictionary, call function, try/except for an unsupported item in the dict—when it could just be an if/else.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -119,19 +126,29 @@ def __init__(  # noqa: PLR0913
...
         self._load_args = deepcopy(self.DEFAULT_LOAD_ARGS)
-        if load_args is not None:
+        if load_args:
             self._load_args.update(load_args)
         if database is not None:
             self._load_args["database"] = database
 
         self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
-        if save_args is not None:
+        if save_args:
```
deepyaman, line 135: I think this is technically slightly more efficient. IIRC I tried to be consistent with the other dataset implementations; maybe it makes sense to undo this for now, and potentially (separately) just update all the datasets to do this?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -119,19 +126,29 @@ def __init__(  # noqa: PLR0913
...
             self._save_args.update(save_args)
         if database is not None:
             self._save_args["database"] = database
 
         self._materialized = self._save_args.pop("materialized")
 
+        # Handle mode / overwrite conflict
```
deepyaman, line 142: Wonder if insert + view needs to be disallowed; I haven't actually checked yet.

deepyaman (commented):
Thanks for the PR!

Did a quick-pass review. I see you're still actively updating, so I'll revisit this later.

deepyaman:
> I've been activily working with this because I need the features now 😅.

Best reason. 😁 

> I've also added support for passing credentials to the `ibis.TableDataset` to this PR I'll update the OP shortly to reflect this. Is that ok or should I create a separate PR for this feature? @deepyaman

Sorry, I didn't check in time; ideal is a separate PR, but it's probably no big deal to have it in the same.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -149,9 +204,49 @@ def connection(self) -> BaseBackend:
     def load(self) -> ir.Table:
         return self.connection.table(self._table_name, **self._load_args)
 
-    def save(self, data: ir.Table) -> None:
+    def save(self, data: pd.DataFrame | ir.Table) -> None:
```
deepyaman: Why does this PR introduce support for saving pandas dataframes?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -101,22 +108,51 @@ def __init__(  # noqa: PLR0913
                 (e.g. `("catalog", "database")`) or a dotted string path
                 (e.g. `"catalog.database"`) to reference a table or view
                 in a multi-level table hierarchy.
-            connection: Configuration for connecting to an Ibis backend.
-                If not provided, connect to DuckDB in in-memory mode.
+            credentials: (Preferred) Connection information for the Ibis backend.
+                Can be a connection string or a dict of parameters. Supersedes
+                `connection`.
+            connection: (Deprecated) Configuration for connecting to an Ibis
+                backend. Use `credentials` instead.
```
deepyaman: I don't think this is right. It should be fine to keep `connection` visible, and use `credentials` on top of that. For example, `connection` could include information like the database name, and there's no reason to hide it, whereas `user` and `password` should of course be obsured.

Look at something like https://github.com/kedro-org/kedro-plugins/blob/609a38c579fe4d7808526ee178d33e1d397f68b7/kedro-datasets/kedro_datasets/pickle/pickle_dataset.py#L178 for an example of how `credentials` is typically used.

In the Ibis case, I think this could be as simple as updating the result of https://github.com/kedro-org/kedro-plugins/blob/609a38c579fe4d7808526ee178d33e1d397f68b7/kedro-datasets/kedro_datasets/ibis/table_dataset.py#L119 and adding the `connection` keys. (`_connection_config` is used for the cache of connections, so it needs to be the one with the final correct values.)

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -101,22 +108,51 @@ def __init__(  # noqa: PLR0913
...
-                `materialized` in `save_args`.
-            metadata: Any arbitrary metadata. This is ignored by Kedro,
-                but may be consumed by users or external plugins.
+                `materialized` in `save_args`. The `mode` parameter controls
+                the behavior when saving data:
+                - _"overwrite"_: Overwrite existing data in the table.
+                - _"append"_: Append contents of the new data to the existing table (does not overwrite).
+                - _"error"_ or _"errorifexists"_: Throw an exception if the table already exists.
+                - _"ignore"_: Silently ignore the operation if the table already exists.
+                These options are similar to those in Spark's DataFrameWriter (see: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.mode.html).
```
deepyaman: Nit: I don't think this is necessary, since an Ibis user may not be familiar with Spark. Should just be documented sufficiently here.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -101,22 +108,51 @@ def __init__(  # noqa: PLR0913
...
+        metadata: Any arbitrary metadata. This is ignored by Kedro,
+            but may be consumed by users or external plugins.
```
deepyaman:
```suggestion
            metadata: Any arbitrary metadata. This is ignored by Kedro,
                but may be consumed by users or external plugins.
```
I think this got accidentally dedented?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -101,22 +108,51 @@ def __init__(  # noqa: PLR0913
...
         """
 
+        if credentials is not None and connection is not None:
+            import warnings  # noqa: PLC0415
+
+            warnings.warn(
+                "Both 'credentials' and deprecated 'connection' were provided. 'credentials' will be used.",
+                DeprecationWarning,
+            )
+
```
deepyaman: See above comment.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -1,10 +1,12 @@
 """Provide data loading and saving functionality for Ibis's backends."""
+
```
deepyaman: Unnecessary change.
```suggestion
```

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -47,7 +53,7 @@ class TableDataset(ConnectionMixin, AbstractDataset[ir.Table, ir.Table]):
         >>> dataset = TableDataset(
         ...     table_name="test",
         ...     connection={"backend": "duckdb", "database": tmp_path / "file.db"},
-        ...     save_args={"materialized": "table"},
+        ...     save_args={"materialized": "table", "mode": "overwrite"},
```
deepyaman:
```suggestion
        ...     save_args={"materialized": "table"},
```
Nit: Doesn't really matter, but I think keeping this example very simple (and leaving it as assumed overwrite) is fine.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -72,6 +78,7 @@ def __init__(  # noqa: PLR0913
         *,
         table_name: str,
         database: str | None = None,
+        credentials: dict[str, Any] | str | None = None,
```
deepyaman: Is it necessary to introduce connection string support in this PR? It's a lot of changes happening in one PR:

- Support `credentials`
- Support connection string
- Support `append` mode

Ideally these should all be separate PRs. I know agreed to leave the `credentials` change in here, but then connection string is yet another change.

The other thing is, changes like `credentials` and supporting `connection` string need to be supported in `ibis.FileDataset`, too.

I would suggest getting things to work for now, and then once everything is nice and passing, I can try and help split it into 2-3 PRs (if you'd like help on that). In case something ends up accidentally introducing breaking or incorrect behavior, I'd like to be able to isolate the changes as much as possible (in addition to it just generally being better practice).

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -101,22 +108,51 @@ def __init__(  # noqa: PLR0913
...
+                - _"append"_: Append contents of the new data to the existing table (does not overwrite).
+                - _"error"_ or _"errorifexists"_: Throw an exception if the table already exists.
+                - _"ignore"_: Silently ignore the operation if the table already exists.
+                These options are similar to those in Spark's DataFrameWriter (see: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.mode.html).
+        metadata: Any arbitrary metadata. This is ignored by Kedro,
+            but may be consumed by users or external plugins.
         """
 
+        if credentials is not None and connection is not None:
+            import warnings  # noqa: PLC0415
```
deepyaman: Why `import warnings` here? `warnings` is built-in and there's no reason to specifically delay this import.

(But maybe this all not necessary if we don't deprecate `connection` to begin with.)

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -149,9 +204,49 @@ def connection(self) -> BaseBackend:
...
+            return
         writer = getattr(self.connection, f"create_{self._materialized}")
-        writer(self._table_name, data, **self._save_args)
+        if self._mode == "append":
+            if not self._exists():
+                writer(self._table_name, data, overwrite=False, **self._save_args)
+            elif hasattr(self.connection, "insert"):
+                self.connection.insert(self._table_name, data, **self._save_args)
+            else:
+                raise NotImplementedError(
```
deepyaman: Nit: Maybe this should be a `DatasetError`? It's not a missing implementation on the Kedro-Datasets side.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -149,9 +204,49 @@ def connection(self) -> BaseBackend:
...
+                    f"Insert mode is not supported by the {self.connection!r} backend."
```
deepyaman:
```suggestion
                    f"Insert mode is not supported by the {self.connection.name} backend."
```

or maybe even more explicit

```suggestion
                    f"The {self.connection.name} backend for Ibis does not support inserts."
```

Or something. I think my main point is to use the backend name rather than dumping the config.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -149,9 +204,49 @@ def connection(self) -> BaseBackend:
...
+        if self._mode == "overwrite":
+            writer(self._table_name, data, overwrite=True, **self._save_args)
+        elif self._mode in ("error", "errorifexists"):
+            writer(self._table_name, data, overwrite=False, **self._save_args)
+        elif self._mode == "ignore":
+            if self._exists():
+                return
+            writer(self._table_name, data, overwrite=False, **self._save_args)
+        else:
+            raise ValueError(f"Unknown mode: {self._mode!r}")
```
deepyaman: I feel like this would be better to validate in `__init__`? No need to wait until failure.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -149,9 +204,49 @@ def connection(self) -> BaseBackend:
...
+
+    def _get_backend_name(self) -> str | None:
```
deepyaman: I can't remember if there's any downside in getting it from `self.connection` (i.e. `self.connection.name`). Maybe there are some places where it's used where `self.connection` isn't yet initialized?

kedro-datasets/kedro_datasets/ibis/table_dataset.py
gitgud5000: @deepyaman  
The output of a node would be a pandas df, similar to `SQLTableDataset`.

In my use case, I load ibis table objects and perform manipulations in-database, then collect and do final manipulation in-memory and save the result. 

Is this the intended behavior, or should it be?

[Ibis Docs](https://ibis-project.org/reference/expression-tables) examples of `con.create_table` expect a DataFrame-like object, so supporting `pandas.DataFrame` here seems consistent.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
gitgud5000: removed L:128

kedro-datasets/kedro_datasets/ibis/table_dataset.py
gitgud5000: @deepyaman I'll go ahead and address your comments and get things working first. I'd really appreciate your help with splitting this into separate PRs once everything is copacetic.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: I would recommend proposing this separately.

`con.create_table` does take a dataframe-like object, but that extends beyond pandas (including PyArrow tables, Polars frames, etc.).

The way this is also supported (at least, for the most part I think) is by wrapping whatever object in a `ibis.memtable`, like in https://github.com/ibis-project/ibis/blob/df4f1858e7f9bc36d589dcf2a53f9f16b1acd92a/ibis/backends/duckdb/__init__.py#L180. I think users can wrap it themselves before return as a workaround for now.

The main issue I have with supporting pandas dataframes here is it breaks the usual assumption of a Kedro user that you can save and get back the same object. In this case, you would save a `pd.DataFrame` but get back an `ibis.Table`, which doesn't fit with the usual model of datasets.

That said, I think there is definite value to being able to leverage Ibis as a more universal dataset, that could save and return other dataframe types. For example, if you loaded an `ibis.TableDataset` as an input to a node that expects something with a `pd.DataFrame` type hint, and it did the conversion for you automatically, that would be cool. (Some sort of type-aware conversion.) I do think this would require some design and alignment, though.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: >  I'd really appreciate your help with splitting this into separate PRs once everything is copacetic.

:100: Of course, happy to help! And thanks for all the effort into improving the Ibis integration!

kedro-datasets/kedro_datasets/ibis/table_dataset.py
gitgud5000: @deepyaman Something like this?

```python
self._table_name = table_name
self._database = database
self._connection_config = connection or self.DEFAULT_CONNECTION_CONFIG
  if credentials is not None:
    if isinstance(credentials, str):
      # Treat string credentials as a connection string under key 'con'
      self._connection_config.update({"con": credentials})
    else:
      self._connection_config.update(deepcopy(credentials))
self.metadata = metadata
```

kedro-datasets/kedro_datasets/ibis/table_dataset.py
gitgud5000: @deepyaman ah, makes sense — I didn’t realize the connection bits were already handled by the `ConnectionMixin` parent class (tbh I just hacked something together that worked 😅).  
I'm removing the `_get_backend_name` method altogether.

gitgud5000: Now that I’ve been writing the tests… it looks like if we want to support connection strings, we’d still need `_get_backend_name` for cases where `self.connection` hasn’t been initialized yet.  

So in scenarios where credentials are passed as a connection string, having `_get_backend_name` available seems necessary.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -134,6 +161,22 @@ def __init__(  # noqa: PLR0913
 
         self._materialized = self._save_args.pop("materialized")
 
+        # Handle mode/overwrite conflict.
+        if save_args and "mode" in save_args and "overwrite" in self._save_args:
+            raise ValueError("Cannot specify both 'mode' and deprecated 'overwrite'.")
```
deepyaman, line 167: There's actually no `DeprecationWarning`, but I'm also not 100% sure there needs to be one. However, if there isn't a `DeprecationWarning`, it doesn't make sense to call it "deprecated", and it shouldn't be on the list of breaking changes in the release notes.

kedro-datasets/RELEASE.md
```diff
@@ -7,13 +8,20 @@
...
 
+- `ibis.TableDataset`: Added configurable save modes via `save_args.mode`, supporting "append", "overwrite", "error"/"errorifexists", and "ignore". Legacy `save_args.overwrite` is mapped to `mode` for backward compatibility; specifying both is now an error.
+
 ## Bug fixes and other changes
 
 - Added primary key constraint to BaseTable.
 
 ## Breaking Changes
+
+- `ibis.TableDataset`: Deprecated `save_args.overwrite` and the `connection` parameter in favor of `save_args.mode` and `credentials`. Using both `overwrite` and `mode` together raises an error; providing both `credentials` and `connection` emits a deprecation warning. The deprecated options will be removed in a future release.
```
```suggestion
```
I don't think it's a breaking change even if raise a `DeprecationWarning` for using 'overwrite' (separate issue about actually raising the `DeprecationWarning`; see below).

(Also, the `credentials` and `connection` bit shouldn't have been left in here when I split out the PR; my bad.)

kedro-datasets/RELEASE.md
```diff
@@ -32,6 +40,7 @@
 - Fixed `PartitionedDataset` to reliably load newly created partitions, particularly with `ParallelRunner`, by ensuring `load()` always re-scans the filesystem .
 - Add a parameter `encoding` inside the dataset `SQLQueryDataset` to choose the encoding format of the query.
 - Corrected the `APIDataset` docstring to clarify that request parameters should be passed via `load_args`, not as top-level arguments.
+- Improved `_connect` and `_describe` for `ibis.TableDataset`; saving an empty pandas DataFrame is now a no-op.
```
```suggestion
```
This was also supposed to get split out, sorry.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -1,18 +1,36 @@
...
-from kedro.io import AbstractDataset
+from kedro.io import AbstractDataset, DatasetError
 
 from kedro_datasets._utils import ConnectionMixin
 
 if TYPE_CHECKING:
     from ibis import BaseBackend
 
 
+class SaveMode(StrEnum):
```
deepyaman, line 25: @gitgud5000 I took the liberty of refactoring save mode logic into an enum.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
```diff
@@ -1,18 +1,36 @@
...
+    """`SaveMode` is used to specify the expected behavior of saving a table."""
+
+    APPEND = auto()
+    OVERWRITE = auto()
+    ERROR = auto()
+    ERRORIFEXISTS = auto()
```
deepyaman, line 31: Do we need both "error" and "errorifexists"? "errorifexists" is a bit more explicit, but also a bit more verbose.

I decided to check if one or the other is preferred by Spark (maybe we could just use that), but it turns out it's not so simple. https://issues.apache.org/jira/browse/SPARK-21640 added "errorifexists" to the Python (and other API) options—"error" used to be the only option—but https://github.com/apache/spark/blob/v4.0.0/sql/api/src/main/java/org/apache/spark/sql/SaveMode.java only defines "errorifexists". Since we don't have the baggage, I'd be fine with just keeping "error" (feels simpler), but I also don't have any strong opinion against supporting both. Happy to have another maintainer (or @gitgud5000) weigh in.

kedro-datasets/pyproject.toml
```diff
@@ -11,6 +11,7 @@ description = "Kedro-Datasets is where you can find all of Kedro's data connecto
 requires-python = ">=3.10"
 license = {text = "Apache Software License (Apache 2.0)"}
 dependencies = [
+    "backports.strenum; python_version < '3.11'",
```
deepyaman, line 14: It's a backport of a standard lib module; I don't personally think it's a big deal to add this dependency for everything, and just for Python 3.10.

kedro-datasets/tests/ibis/test_table_dataset.py
```diff
@@ -35,10 +37,16 @@ def connection_config(request, database):
...
+    # Default table name when not explicitly parametrized
+    return "test" if request.param is _SENTINEL else request.param
+
+
 @pytest.fixture
-def table_dataset(database_name, connection_config, load_args, save_args):
+def table_dataset(table_name, database_name, connection_config, load_args, save_args):
     return TableDataset(
-        table_name="test",
+        table_name=table_name,
```
deepyaman: @gitgud5000 Was making this parametrizable necessary? Just curious.

deepyaman: @gitgud5000 So, I ended up removing this. I see why it's necessary, but I think the correct thing to do is to clean up the environment by deleting any tables created (FWIW that's how Ibis tests also handle it). I've refactored it as such.

kedro-datasets/kedro_datasets/ibis/table_dataset.py
ankatiyar: If we're doing a deprecation and then removal then we could emit a `DeprecationWarning` if `overwrite` is used whether or not it's used in conjunction with the `mode` and keep this `ValueError` as it is. Either way deprecation is not a breaking change so I would suggest not adding it to the "breaking change" section in the docs

kedro-datasets/RELEASE.md
```diff
@@ -15,14 +16,20 @@
 | ------------------------------ | ------------------------------------------------------------- | ------------------------------------ |
 | `polars.PolarsDatabaseDataset` | A dataset to load and save data to a SQL backend using Polars | `kedro_datasets_experimental.polars` |
 
+- Added `mode` save argument to `ibis.TableDataset`, supporting "append", "overwrite", "error"/"errorifexists", and "ignore" save modes. The `overwrite` save argument is mapped to `mode` for backward compatibility; specifying both results in an error.
```
ankatiyar: I'd add that `overwrite` argument is deprecated here and remove the entry from "breaking changes" entirely

deepyaman: Done!

kedro-datasets/kedro_datasets/ibis/table_dataset.py
deepyaman: Done!

PR #1095, opened 2025-06-02
feat(datasets): add docx dataset

soyamimi:
## Description
Adds support for .docx files, which require special handling. Since direct writing is complicated, the document is first saved to a buffer before being written to filesystem.

## Development notes
Built using python-docx. Follows the pattern of other Kedro datasets with custom logic for handling Word documents.

## Checklist

- [x] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)

### DIFF
--- a/kedro-datasets/tests/openxml/test_docx_dataset.py
+++ b/kedro-datasets/tests/openxml/test_docx_dataset.py
@@ -62,7 +62,7 @@ class TestDocxDataset:
 
     def test_load_missing_file(self, docx_dataset):
         """Check the error when trying to load missing file."""
-        pattern = r"Failed while loading data from dataset DocxDataset\(.*\)"
+        pattern = r"Failed while loading data from dataset kedro_datasets.openxml.docx_dataset.DocxDataset\(.*\)"
         with pytest.raises(DatasetError, match=pattern):
             docx_dataset.load()
 
@@ -122,7 +122,7 @@ class TestDocxDatasetVersioned:
 
     def test_no_versions(self, versioned_docx_dataset):
         """Check the error if no versions are available for load."""
-        pattern = r"Did not find any versions for DocxDataset\(.+\)"
+        pattern = r"Did not find any versions for kedro_datasets.openxml.docx_dataset.DocxDataset\(.+\)"
         with pytest.raises(DatasetError, match=pattern):
             versioned_docx_dataset.load()
 
@@ -137,7 +137,7 @@ class TestDocxDatasetVersioned:
         corresponding docx file for a given save version already exists."""
         versioned_docx_dataset.save(dummy_data)
         pattern = (
-            r"Save path \'.+\' for DocxDataset\(.+\) must "
+            r"Save path \'.+\' for kedro_datasets.openxml.docx_dataset.DocxDataset\(.+\) must "
             r"not exist if versioning is enabled\."
         )
         with pytest.raises(DatasetError, match=pattern):
@@ -156,7 +156,7 @@ class TestDocxDatasetVersioned:
         the subsequent load path."""
         pattern = (
             rf"Save version '{save_version}' did not match load version "
-            rf"'{load_version}' for DocxDataset\(.+\)"
+            rf"'{load_version}' for kedro_datasets.openxml.docx_dataset.DocxDataset\(.+\)"
         )
         with pytest.warns(UserWarning, match=pattern):
             versioned_docx_dataset.save(dummy_data)


kedro-datasets/kedro_datasets_experimental/docx/docx_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+    AbstractVersionedDataset,
+    DatasetError,
+    Version,
+    get_filepath_str,
+    get_protocol_and_path,
+)
+from six import BytesIO
+
+
+class DOCXDataset(AbstractVersionedDataset[dict, dict]):
```
ankatiyar: Could this be renamed to `DocxDataset` to maintain consistency with all the other datasets? I don't believe we have any other name in capitals

astrojuanlu:
> I’m just wondering (with a bit of ambition), how can it be turned into one of the core datasets ? 😆

https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/CONTRIBUTING.md#core-datasets

Condition 1 is a bit out of your control 😄 but you can work on the other 5!

ravi-kumar-pilla:
Hi @soyamimi , Are you still actively working on pushing this dataset to be part of core datasets or you are fine to draft this as experimental ?

merelcht:
Hi @soyamimi, apologies this fell of my radar a bit. Sadly we can't use the poll as it's anonymous and the vote needs to happen by our TSC. I'll tag people and ask them to review.

kedro-datasets/kedro_datasets/docx/docx_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+        type: docx.DocxDataset
+        filepath: soya.docx
+    ```
+    ### Example usage for the [Python API](https://docs.kedro.org/en/stable/data/advanced_data_catalog_usage.html):
+
+    ```python
+
+    from kedro_datasets.docx import DocxDataset
+
+    data = Document()
```
merelcht: What type is this Document? I'm guessing it's a class you'd need to import?

merelcht (approved):

rashidakanchwala (approved):

yetudada (approved):
All approved and I'm very curious - @soyamimi what use case prompted creating this dataset?

datajoely:
is this going into experimental datasets?

lrcouto (approved):

kedro-datasets/RELEASE.md
```diff
@@ -3,6 +3,11 @@
 ## Major features and improvements
 
 - Migrated docs to mkdocs
+- Added the following new datasets:
+
+| Type                           | Description                                                         | Location                          |
+|--------------------------------|---------------------------------------------------------------------|-----------------------------------|
+| `docx.DocxDataset`             | A dataset loading and saving .docx files (Word) using `python-docx` | `kedro_datasets.docx` |
```
```suggestion
| `docx.DocxDataset`             | A dataset for loading and saving .docx files (Microsoft Word) using `python-docx` | `kedro_datasets.docx` |
```

kedro-datasets/kedro_datasets/docx/__init__.py
deepyaman, line 1: Thinking about the future, what if somebody adds a `pptx` dataset? I think it would be less duplicative and maybe more forward-thinking to put it under `openxml`.

kedro-datasets/kedro_datasets/docx/docx_dataset.py
deepyaman: I didn't see doctests got killed in https://github.com/kedro-org/kedro-plugins/pull/1078. :( Would be nice to have it back (understand there are challenges), because otherwise you end up with technically non-working snippets like this.

deepyaman (changes requested):
Happy for others to weigh in on the module for this, but I think `docx` is overly (and unnecessarily) restrictive.

merelcht:
@deepyaman To answer to your comment 

> I didn't see doctests got killed in https://github.com/kedro-org/kedro-plugins/pull/1078. :( Would be nice to have it back (understand there are challenges), because otherwise you end up with technically non-working snippets like this.

We tried a lot of things to get this to work, but just couldn't figure out how... 😣 I know it's not ideal, so would appreciate any help to get this working again. We didn't want to let this block the complete doc migration though so that's why we decided to remove it in the end.

kedro-datasets/kedro_datasets/docx/docx_dataset.py
soyamimi: > What type is this Document? I'm guessing it's a class you'd need to import?

Exactly, it is a class from docx.document.Document which is used to handle .docx documents.

kedro-datasets/kedro_datasets/docx/docx_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+    ```yaml
+    cats:
+        type: docx.DocxDataset
+        filepath: soya.docx
+    ```
+    ### Example usage for the [Python API](https://docs.kedro.org/en/stable/data/advanced_data_catalog_usage.html):
+
+    ```python
+
+    from kedro_datasets.docx import DocxDataset
```
deepyaman:
```suggestion
    from docx import Document
    from kedro_datasets.docx import DocxDataset
```

kedro-datasets/kedro_datasets/docx/docx_dataset.py
```diff
@@ -0,0 +1,177 @@
...
+
+    data = Document()
+    data.add_paragraph("Hello, World !")
+    dataset = DocxDataset(filepath=tmp_path / "test.docx")
+    dataset.save(data)
+
+    reloaded = dataset.load()
+    assert data == reloaded
+    assert dummy_data.paragraphs[0].text == reloaded.paragraphs[0].text
```
deepyaman:
```suggestion
    assert data.paragraphs[0].text == reloaded.paragraphs[0].text
```

soyamimi:
I've made all the requested changes to my best. Let me know if anything else needs to be updated. Thanks.

soyamimi:
> > > I'd still be in favor of doing this; think there's no downside, only upside: [#1095 (comment)](https://github.com/kedro-org/kedro-plugins/pull/1095#discussion_r2166951444)
> > > @merelcht @astrojuanlu any thoughts?
> > 
> > 
> > I'm happy to add it like this and create an issue to extend it to use `openxml`.
> 
> Sorry, maybe I wasn't clear. I'm saying should put the dataset under `openxml` (or something more generic to capture future `openxml`-based datasets in the same group) rather than `docx`. Nothing would be different other than the dataset would be `openxml.DocxDataset` instead of `docx.DocxDataset`. This way, let's say @soyamimi or somebody else has a similar requirement to produce PPTX reports, it can be added under the same group like `openxml.PptxDataset` (or even if you wanted to add a generic `XlsxDataset` that isn't through pandas, Polars, etc.). We wouldn't generally want to proliferate additional groups for each file format.

Thanks for your comment. All updated :)

astrojuanlu (approved):
Thank you @soyamimi !

PR #1118, opened 2025-06-27
docs(datasets): apply filters for private attributes

Huongg:
## Checklist

### DIFF
--- a/kedro-datasets/mkdocs.yml
+++ b/kedro-datasets/mkdocs.yml
@@ -112,8 +112,8 @@ extra_javascript:
   - javascript/deindex-old-docs.js
 
 nav:
-  - Welcome: https://docs.kedro.org/en/develop/
-  - Kedro: https://docs.kedro.org/en/develop/getting-started/course/
+  - Welcome: https://docs.kedro.org/en/1.0.0rc1/
+  - Kedro: https://docs.kedro.org/en/1.0.0rc1/getting-started/course/
   - Kedro-Viz: https://docs.kedro.org/projects/kedro-viz/en/feature-branch-mkdocs/
   - Kedro-Datasets:
     - API documentation:
--- a/kedro-datasets/mkdocs.yml
+++ b/kedro-datasets/mkdocs.yml
@@ -56,6 +56,8 @@ plugins:
       handlers:
         python:
           options:
+            filters:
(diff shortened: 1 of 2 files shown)


kedro-datasets/mkdocs.yml
```diff
@@ -112,8 +114,8 @@ extra_javascript:
   - javascript/deindex-old-docs.js
 
 nav:
-  - Welcome: https://docs.kedro.org/en/develop/
-  - Kedro: https://docs.kedro.org/en/develop/getting-started/course/
+  - Welcome: https://docs.kedro.org/en/1.0.0rc1/
```
Huongg: This isn't related to the filters change — I'm pretty sure I had already updated this. why it's showing up differently here? 🤔

https://github.com/kedro-org/kedro-plugins/blob/feature/8.0/kedro-datasets/mkdocs.yml#L115

Huongg: oh never mind, i havent merged the lates one from feature branch..oops

kedro-datasets/mkdocs.yml
```diff
@@ -56,6 +56,8 @@ plugins:
       handlers:
         python:
           options:
+            filters:
+            - "!^_[^_]"
```
deepyaman, line 60: I tried this with `ibis.TableDataset` too, and just `connection` was showing there; filter seems to be behaving weirdly somehow.

Another possibility is just explicitly enumerate all possibilities like:

```suggestion
            members:
            - DEFAULT_FS_ARGS
            - DEFAULT_LOAD_ARGS
            - DEFAULT_SAVE_ARGS
            - load
            - save
            - preview
```

That seems to work, although it means having to have the full set (e.g. Ibis has `DEFAULT_CONNECTION_CONFIG`, which won't show up for anything else if listed, but you still need to add all possibilities to the list). Guess could have some script validate the list is complete.

Huongg: Do you know how much effort was estimated to create this script? I'm happy either way—perhaps @astrojuanlu  can weigh in on whether it's worth investigating further

astrojuanlu: Hardcoding the possible values and having a script to validate them seems like overkill. How do other projects handle this?

PR #1142, opened 2025-07-11
fix(datasets): Add primary key constraint to base table

### LINKED ISSUE #1140: `primary_key` in `ManagedTableDataset` doesn't set primary key
## Description
`ManagedTableDataset` has a `primary_key` property inherited from `BaseTableDataset`:

https://github.com/kedro-org/kedro-plugins/blob/86a971bf4549609d13ae638ee443a37e6b9ff0eb/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py#L229

However, this is not being used to set the actual primary key of the table, but to create merge predicates:

https://github.com/kedro-org/kedro-plugins/blob/86a971bf4549609d13ae638ee443a37e6b9ff0eb/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py#L471-L486

## Context
I was trying to [use an existing Delta Table in the Databricks Feature Store](https://learn.microsoft.com/en-gb/azure/databricks/machine-learning/feature-store/uc/feature-tables-uc#use-an-existing-delta-table-in-unity-catalog-as-a-feature-table), which requires a primary key.

## Steps to Reproduce
1. Create a `ManagedTableDataset` with an appropriate `primary_key` (for example, `primary_key="id"` if using the Companies dataset from Spaceflights, see https://github.com/kedro-org/kedro-academy/blob/ec4b1c8/kedro-databricks-bootcamp/02_databricks/rocketfuel/notebooks/02_1-Kedro-on-Databricks-part-1.ipynb)
2. Perform a `DESCRIBE TABLE EXTENDED`, and verify that there are no Constraints ❌ 
3. Manually `ALTER` the table to add a primary key
4. Repeat a `DESCRIBE TABLE EXTENDED` and verify that there are Constraints ✔️ 

## Expected Result
A primary key should be set.

Or, alternatively, the parameter should be called something else to avoid giving a false impression.

## Actual Result
Nothing.

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

ravi-kumar-pilla:
## Description

Resolves #1140 

## Development notes

* Updated validation logic for a primary key
* Added a method to alter table and add primary key specified

**NOTE:** 
- Please have a look at - https://github.com/kedro-org/kedro-plugins/issues/1140#issuecomment-3112006382

## QA notes

- Try [databricks free edition ](https://docs.databricks.com/aws/en/getting-started/free-edition) in case you do not have Platform McK Azure Analytics Service
- Follow the [databricks setup](https://github.com/kedro-org/kedro-academy/tree/main/kedro-databricks-bootcamp#3-databricks-setup)
- Follow the notebook for initial setup - https://github.com/kedro-org/kedro-academy/blob/main/kedro-databricks-bootcamp/02_databricks/02_0-Unity-Catalog-checks.ipynb
- Follow the notebook to create connection with Unity Catalog completing Exercise 1 - https://github.com/kedro-org/kedro-academy/blob/main/kedro-databricks-bootcamp/02_databricks/rocketfuel/notebooks/02_1-Kedro-on-Databricks-part-1.ipynb . Though you need to modify the companies catalog entry. 

Make sure your conf at `kedro-databricks-bootcamp/02_databricks/rocketfuel/conf/databricks/catalog.yml` is as below. Note that the companies entry contains a primary_key prop.

```yaml
_uc_catalog: <your catalog name>
_uc_schema: <your schema/database name>

companies_raw:
  type: spark.SparkDataset
  filepath: /Volumes/${_uc_catalog}/${_uc_schema}/bronze/companies.csv
  file_format: csv
  load_args:
    header: True
    inferSchema: True

reviews_raw:
  type: spark.SparkDataset
  filepath: /Volumes/${_uc_catalog}/${_uc_schema}/bronze/reviews.csv
  file_format: csv
  load_args:
    header: True
    inferSchema: True

shuttles_raw:
  type: pandas.ExcelDataset
  filepath: /Volumes/${_uc_catalog}/${_uc_schema}/bronze/shuttles.xlsx

companies:
  type: databricks.ManagedTableDataset
  catalog: ${_uc_catalog}
  database: ${_uc_schema}
  table: companies
  write_mode: overwrite
  primary_key: "id"

reviews:
  type: databricks.ManagedTableDataset
  catalog: ${_uc_catalog}
  database: ${_uc_schema}
  table: reviews
  write_mode: overwrite

shuttles:
  type: databricks.ManagedTableDataset
  catalog: ${_uc_catalog}
  database: ${_uc_schema}
  table: shuttles
  write_mode: overwrite
  dataframe_type: pandas
```

- Once you are done with `catalog.save("companies", df_companies)`, you should see a companies ManagedTableDataset being created in `<your-catalog>/<your-schema>`
- The `companies` table will not have a primary key constraint
- Uninstall `kedro-datasets` - `!pip uninstall -y kedro-datasets` in your notebook on databricks
- Install `kedro-datasets` pointing to current PR branch - `!pip install git+https://github.com/kedro-org/kedro-plugins.git@fix/primary_key#subdirectory=kedro-datasets`
- Uninstall rich - `!pip uninstall -y rich`
- Restart python - `%restart_python`
- Load ext kedro - `%load_ext kedro`
- Reload kedro pointing to databricks env - `%reload_kedro --env databricks`
- Try saving `companies_raw` data to `companies` again 

```python
df_companies = catalog.load("companies_raw")
df_companies.show(1)
```

```python
catalog.save("companies", df_companies)
```

- Go back to see `companies` ManagedTableDataset at `<your-catalog>/<your-schema>`. It should have a primary key constraint at the id column. You can verify this by running below query in the notebook too -

```bash
%sql DESCRIBE TABLE EXTENDED `your-catalog`.your-schema.companies
```

image

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
+++ b/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
@@ -268,6 +268,17 @@ class BaseTable:
             DatasetError: If retrieving of primary key column names fail.
         """
         try:
+            # Check if running in Unity Catalog
+            catalog_impl = get_spark().conf.get(
+                "spark.sql.catalogImplementation", "hive"
+            )
+            if catalog_impl == "hive":
+                # Fallback for non-Unity Catalog environments
+                logger.warning(
+                    "Unity Catalog is not available. Skipping primary key retrieval."
+                )
+                return None
+
             columns_df = (
                 get_spark()
                 .table("system.information_schema.key_column_usage")
--- a/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
+++ b/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
@@ -164,11 +164,6 @@ class BaseTable:
                     f"Unable to retrieve schema for table '{self.full_table_location()}': {exc}"
                 )
 
-        # if table_schema is None:
-        #     raise DatasetError(
-        #         f"Cannot validate primary key because the schema for table '{self.full_table_location()}' is not available."
-        #     )
-
         primary_keys = (
             [self.primary_key]
             if isinstance(self.primary_key, str)
--- a/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
+++ b/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
@@ -138,14 +138,56 @@ class BaseTable:
     def _validate_primary_key(self):
         """Validates the primary key of the table.
 
+        Ensures that:
+        1. A primary key is provided when `write_mode` is "upsert".
+        2. The primary key column(s) exist in the table schema.
+        3. The primary key column(s) do not contain null values.
+
         Raises:
-            DatasetError: If no `primary_key` is specified.
+            DatasetError: If no `primary_key` is specified, if the primary key
+            column(s) do not exist in the table schema, or if the primary key
+            column(s) contain null values.
         """
         if self.primary_key is None or len(self.primary_key) == 0:
             if self.write_mode == "upsert":
                 raise DatasetError(
                     f"`primary_key` must be provided for"
                     f"`write_mode` {self.write_mode}"
+                )
+
+        table_schema = self.schema()
+        if table_schema is None:
+            raise DatasetError(
+                f"Cannot validate primary key because the schema for table '{self.full_table_location()}' is not available."
+            )
+
+        primary_keys = (
+            [self.primary_key]
+            if isinstance(self.primary_key, str)
+            else self.primary_key
+        )
+
+        # Check if primary key columns exist in the schema
+        missing_columns = [
+            column for column in primary_keys if column not in table_schema.fieldNames()
+        ]
+        if missing_columns:
+            raise DatasetError(
+                f"Primary key column(s) {missing_columns} do not exist in the table schema."
+            )
+
+        # Check if primary key columns contain null values
+        for col in primary_keys:
+            null_count = (
+                get_spark()
+                .table(self.full_table_location())
+                .filter(f"{col} IS NULL")
+                .count()
+            )
+            if null_count > 0:
+                raise DatasetError(
+                    f"Primary key column '{col}' contains {null_count} null value(s). "
+                    "Primary key columns must not contain null values."
                 )
 
     def full_table_location(self) -> str | None:
@@ -201,6 +243,54 @@ class BaseTable:
         except (ParseException, AnalysisException) as exc:
             logger.warning("error occured while trying to find table: %s", exc)
             return False
+
+    def add_primary_key_constraint(self) -> None:
+        """Adds a primary key constraint to the table.
+
+        This method uses Delta Lake's `ALTER TABLE` command to add a primary key
+        constraint to the table. Note that this requires Delta Lake's constraints
+        feature to be enabled in your environment.
+
+        Raises:
+            DatasetError: If the table does not exist or if the primary key is not defined.
+
+        Example:
+            >>> table = BaseTable(
+            ...     database="default",
+            ...     table="example_table",
+            ...     primary_key="id",
+            ...     write_mode="upsert",
+            ... )
+            >>> table.add_primary_key_constraint()
+        """
+        if not self.exists():
+            raise DatasetError(f"Table '{self.full_table_location()}' does not exist.")
+
+        if not self.primary_key:
+            raise DatasetError("Primary key is not defined for this table.")
+
+        primary_key = (
+            self.primary_key
+            if isinstance(self.primary_key, list)
+            else [self.primary_key]
+        )
+
+        primary_key_columns = ", ".join(f"`{col}`" for col in primary_key)
+
+        try:
+            get_spark().sql(
+                f"ALTER TABLE {self.full_table_location()} "
+                f"ADD CONSTRAINT pk_constraint PRIMARY KEY ({primary_key_columns})"
+            )
+            logger.info(
+                "Primary key constraint added to table '%s' on columns: %s",
+                self.full_table_location(),
+                primary_key_columns,
+            )
+        except Exception as exc:
+            raise DatasetError(
+                f"Failed to add primary key constraint to table '{self.full_table_location()}': {exc}"
+            )
 
 
 class BaseTableDataset(AbstractVersionedDataset):
@@ -291,6 +381,8 @@ class BaseTableDataset(AbstractVersionedDataset):
 
         self.metadata = metadata
         self._version = version
+        # [TODO: Confirm on where to make this call]
+        self._table.add_primary_key_constraint()
 
         super().__init__(
             filepath=None,  # type: ignore[arg-type]
--- a/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
+++ b/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
@@ -202,7 +202,7 @@ class BaseTable:
             logger.warning("error occured while trying to find table: %s", exc)
             return False
 
-    def add_primary_key_constraint(self, primary_keys: list[str]) -> None:
+    def _add_primary_key_constraint(self, primary_keys: list[str]) -> None:
         """Adds a primary key constraint to the table.
 
         This method uses Delta Lake's `ALTER TABLE` command to add a primary key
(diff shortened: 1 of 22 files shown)


kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
```diff
@@ -403,6 +509,9 @@ def _save(self, data: DataFrame | pd.DataFrame) -> None:
         if method:
             method(data)
 
+        # [TODO: Confirm on where to make this call]
+        self._table.add_primary_key_constraint()
```
ravi-kumar-pilla: Hi Team,

Trying to fix - https://github.com/kedro-org/kedro-plugins/issues/1140

The code here works and adds primary key as intended. However, I haven't seen any kedro-dataset which modifies the metadata i.e., adding constraints or modifying schema (please point me to anything you are aware of). So I am not sure how to approach this. Do we only care about read and write abstractions or modifying metadata is also of interest ?

1. If we add primary constraint, do we also provide a way to modify or update it if needed ?

2. Do we also provide a way for modifying foreign keys ?

3. I am not sure about the placement of this call, I added it after the save operation (obviously need to add a check of not repeating the pk addition if exists)

4. Primary key also needs the columns to be not null, so I added another [alter statement
](https://github.com/kedro-org/kedro-plugins/pull/1142/files#diff-91c0902f5e866bb40a2704ed17796704c423b33c2ad22a44cda999e5a4bd1b18R284) to set that. But I feel this should be part of schema provided by the user.

5. The sql syntax (ALTER statements) -
- Works on Databricks runtime environments that:
   > Have Unity Catalog enabled.
   > Use DBR versions supporting constraints (e.g., DBR 13+).
   > On managed Unity Catalog tables.
 
 - Does not work:
   > On local Spark sessions.
   > On Databricks without Unity Catalog.
   > On plain Delta Lake environments.
   
 6. Because of the limitations mentioned in (5), should we leave the constraint addition to the user ? 
 
 cc: @astrojuanlu @ankatiyar @noklam
 
 Thank you

rashidakanchwala: also  @merelcht

rashidakanchwala: Some thoughts here:

This is a Databricks-specific feature - all Feature Tables in Databricks require a primary key. In this case, users are using `ManagedTableDataset` to write a Delta table that they want to register as a Feature Table. Since a Feature Table must have a primary key, this becomes a necessary step.

So while this constraint is specific to Feature Store (and not to generic Delta usage), given that ManagedTableDataset is part of our Databricks integration, it might make sense to support primary key constraints here - we have it as opt-in behavior anyway, we just need to make sure if we do this we highlight it properly in our docs.

astrojuanlu: There's another thing about this change, which is that we're assuming that the user wants to use the primary key for UPSERT operations (reusing the same parameter for 2 different things). This might or might not be a big deal, we should consult with users of this dataset.

astrojuanlu: About adding the `ALTER TABLE` logic here, answering to your points @ravi-kumar-pilla :

> 1. If we add primary constraint, do we also provide a way to modify or update it if needed ?

Kedro doesn't have the notion of "modifying" a dataset, so I think we should refrain from it. They can always do it out of band.

> 2. Do we also provide a way for modifying foreign keys ?

I think this should be out of scope for now.

> 3. I am not sure about the placement of this call, I added it after the save operation (obviously need to add a check of not repeating the pk addition if exists)

That's about right I think

> 4. Primary key also needs the columns to be not null, so I added another alter statement to set that. But I feel this should be part of schema provided by the user.

Yeah we're doing the right thing I think. If the user tries to set the primary key on a column that contains nulls, they will get an error. It should be the user responsibility to fix this.

> 5. The sql syntax (ALTER statements) [...] Does not work
> - On local Spark sessions.
> - On plain Delta Lake environments.

No problem, this is the `databricks.ManagedTableDataset` after all.

> - On Databricks without Unity Catalog.

Potentially an issue to a minority of users but I think Databricks is pushing everyone to use the Unity Catalog. I'm not sure our `ManagedTableDataset` works outside of UC even before this PR.

> 6. Because of the limitations mentioned in (5), should we leave the constraint addition to the user ?

I don't think so (see above)

ravi-kumar-pilla: Thank you @astrojuanlu

ravi-kumar-pilla:
> Thanks @ravi-kumar-pilla, would you be able to add a minimal guide on how to QA this.

Done !

rashidakanchwala:
I tried this a, but it doesn't show me PK. maybe we cna connect later today. thanks @ravi-kumar-pilla 
Screenshot 2025-07-28 at 12 20 20

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
```diff
@@ -16,6 +16,7 @@
     VersionNotFoundError,
 )
 from pyspark.sql import DataFrame
+from pyspark.sql.functions import col as col_func
```
noklam: Is there a specific reason for this alias? If this is due to name conflicts, you may consider a more common approach:
```
from pyspark.sql import functions as F # or f, then use `f.col`
```

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
```diff
@@ -399,10 +536,32 @@ def _save(self, data: DataFrame | pd.DataFrame) -> None:
...
+        )
+
         method = getattr(self, f"_save_{self._table.write_mode}", None)
         if method:
             method(data)
 
+        # Add the primary key constraint only if it is overwrite or is different than existing
+        if new_primary_keys and (
+            self._table.write_mode == "overwrite"
+            or set(existing_primary_keys) != set(new_primary_keys)
```
noklam: Should this be done automatically or we should throw error at this point?

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
479BetaCoder: It is mainly due to name conflicts. okay let me do that. thanks !

ravi-kumar-pilla: It is mainly due to name conflicts. okay let me do that. thanks !

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
ravi-kumar-pilla: Not sure if I understood it completely but do you mean to remove this check and let databricks error out ? Similar to this question, had a discussion with Ankita and was thinking if we should even validate anything about primary key in the kedro-datasets side or just assign it as passed by user and databricks throw an error if something is weird

ravi-kumar-pilla:
> My general feeling is that we shouldn't be doing any validation on our end in the dataset but let Databricks throw an error - `kedro-datasets` should just be passing on the config to the underlying library. 

As discussed, I also do partially [agree](https://github.com/kedro-org/kedro-plugins/pull/1142#discussion_r2205762464) (point 4) on this. Most of the time `kedro-datasets` abstracts the load and save operations but I do not see we modify schema of a table (I might be wrong). Thank you

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
noklam: Yes - the original issue in #1140 was mainly around a `primary_key` is specified but it has no effect on table creation. I think this make sense as a default which is implemented in this PR.

The second part of the PR does validation on save. While it may be useful, I advise against it because it should be an error from Databricks/Spark but not on Kedro. It may have some performance hit also if we are checking null everytime.

We can leave it as a helper function/optional thing, but then I am not sure if this is needed at the first place. If we start checking for NULL constrains, then there is a question why don't we check for duplicates as well? I think this is not a simple problem that should be addressed from Kedro side. If this is simple enough, Databricks would have addressed it already. I think the best we can do here is, having some pointer/docs to advise user when using primary key they should check these constrains.

ravi-kumar-pilla: I agree it is an overhead. I removed the validation now.

ravi-kumar-pilla: Another question is should we save the table even if adding of primary constraint fails ? Right now, the table gets saved but just the constraint is not added.

ankatiyar: Maybe a warning?

noklam (approved):

PR #1151, opened 2025-07-24
feat(telemetry): don't warn if consent is explicit

### LINKED ISSUE #1150: Suppress telemetry warning without disabling
## Description
I'd like to allow telemetry data, but don't want to see the noisy output when running the CLI.

How do I disable the warning while still sending data?

## Context
I understand the utility, but don't want to be constantly reminded that I can opt out.

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

deepyaman:
## Description
Proposal to resolve #1150 by only showing the message that Kedro is sending telemetry if the user hasn't explicitly granted permission. It seems reasonable that, if a user has created a `.telemetry` file and marked `consent: true`, they understand and are OK with sending telemetry.

## Development notes
https://github.com/kedro-org/kedro/blob/main/docs/about/telemetry.md will also need to be updated separately; I can do that if this goes in.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,9 +1,16 @@
 # Upcoming Release
 ## Major features and improvements
+
+- Added the following new experimental datasets:
+
+| Type                           | Description                                                   | Location                             |
+|--------------------------------|---------------------------------------------------------------|--------------------------------------|
+| `polars.PolarsDatabaseDataset` | A dataset to load and save data to a SQL backend using Polars | `kedro_datasets_experimental.polars` |
+
 ## Bug fixes and other changes
 ## Breaking Changes
 ## Community contributions
-
+- [Minura Punchihewa](https://github.com/MinuraPunchihewa)
 
 # Release 8.0.0
 
--- a/kedro-datasets/docs/api/kedro_datasets_experimental/index.md
+++ b/kedro-datasets/docs/api/kedro_datasets_experimental/index.md
@@ -12,6 +12,7 @@ Name | Description
 [langchain.ChatOpenAIDataset](langchain.ChatOpenAIDataset.md) | OpenAI dataset used to access credentials at runtime.
 [langchain.OpenAIEmbeddingsDataset](langchain.OpenAIEmbeddingsDataset.md) | ``OpenAIEmbeddingsDataset`` loads a OpenAIEmbeddings `langchain` model.
 [netcdf.NetCDFDataset](netcdf.NetCDFDataset.md) | ``NetCDFDataset`` loads/saves data from/to a NetCDF file using an underlying filesystem (e.g.: local, S3, GCS). It uses xarray to handle the NetCDF file.
+[polars.PolarsDatabaseDataset](polars.PolarsDatabaseDataset.md) | ``PolarsDatabaseDataset`` implementation to access databases as Polars DataFrames. It supports reading from a SQL query and writing to a database table.
 [prophet.ProphetModelDataset](prophet.ProphetModelDataset.md) | ``ProphetModelDataset`` loads/saves Facebook Prophet models to a JSON file using an underlying filesystem (e.g., local, S3, GCS). It uses Prophet's built-in serialisation to handle the JSON file.
 [pytorch.PyTorchDataset](pytorch.PyTorchDataset.md) | ``PyTorchDataset`` loads and saves PyTorch models' `state_dict` using PyTorch's recommended zipfile serialization protocol. To avoid security issues with Pickle.
 [rioxarray.GeoTIFFDataset](rioxarray.GeoTIFFDataset.md) | Loads and saves raster data files as xarray DataArrays. Supports single and multiband GeoTIFFs with CRS validation.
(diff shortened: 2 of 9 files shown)


kedro-telemetry/tests/test_plugin.py
```diff
@@ -192,6 +192,8 @@ def test_before_command_run(self, mocker, fake_metadata, caplog):
             "No personal data or IP addresses are stored on our side. "
             "If you want to opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, "
             "or create a `.telemetry` file in the current working directory with the contents `consent: false`. "
+            "If you don't want to see this message again, "
+            "create a `.telemetry` file in the current working directory with the contents `consent: true`. "
```
patcon: This all looks great!

but as for this line, I'd almost suggest that the project file is the only place to not allow explicit consent? Since someone could version-control it, and then a user silently gives consent. I'd suggest it has to be an envvar or something, as that's a little harder to confuse for being set :)

deepyaman: Hmm, as it stands the env vars are only to disable telemetry. The docs do also say not to version control the `.telemetry` file, and it is in `.gitignore`, so it may be safe enough. 

I don't actually have access to/consume telemetry data directly, so I'll let somebody else review/weigh in.

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -230,13 +234,16 @@ def after_catalog_created(self, catalog):
...
-        )
+        if self._consent_is_implicit:
+            logger.info(
+                "Kedro is sending anonymous usage data with the sole purpose of improving the product. "
+                "No personal data or IP addresses are stored on our side. "
+                "If you want to opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, "
+                "or create a `.telemetry` file in the current working directory with the contents `consent: false`. "
+                "If you don't want to see this message again, "
+                "create a `.telemetry` file in the current working directory with the contents `consent: true`. "
+                "Read more at https://docs.kedro.org/en/stable/configuration/telemetry.html"
```
merelcht: Totally agree with this change, but I'm wondering if this message is now getting a bit long and also slightly confusing at the end. It now sounds like you can only get rid of the message if you set consent to true, which isn't the case. Maybe we can change it to just say: "If you don't want to see this message again, explicitly give or deny consent"

deepyaman: That works, and maybe give the instructions on how to explicitly grant or deny consent in the docs? @merelcht is this what you meant:

```suggestion
        if self._consent_is_implicit:
            logger.info(
                "Kedro is sending anonymous usage data with the sole purpose of improving the product. "
                "No personal data or IP addresses are stored on our side. "
                "If you want to opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, "
                "or create a `.telemetry` file in the current working directory with the contents `consent: false`. "
                "If you don't want to see this message again, explicitly grant or deny consent. "
                "Read more at https://docs.kedro.org/en/stable/configuration/telemetry.html"
```

(I assume the opt-out instructions still need to be here?)

patcon: I like your revision and find it clearer :)

I think with the mention of removing message, I'd recognize that setting `consent: true` does what I want, even without consulting further docs.

Fwiw I find it clearer to read documentation more like directions (or recipes), without me entering as second person via "if _you_ want" etc:

> To opt out, set the `KEDRO_DISABLE_TELEMETRY` or...

> To hide this message, explicitly grant or deny consent.

patcon:
```suggestion
        if self._consent_is_implicit:
            logger.info(
                "Kedro is sending anonymous usage data with the sole purpose of improving the product. "
                "No personal data or IP addresses are stored on our side. "
                "To opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, "
                "or create a `.telemetry` file in the current working directory with the contents `consent: false`. "
                "To hide this message, explicitly grant or deny consent. "
                "Read more at https://docs.kedro.org/en/stable/configuration/telemetry.html"
```

kedro-telemetry/tests/test_plugin.py
patcon: I'm satisfied with that response. Thanks!

patcon (approved):

kedro-telemetry/kedro_telemetry/plugin.py
merelcht: Yes this last suggestion looks great to me 👍

merelcht (approved):

kedro-telemetry/tests/test_plugin.py
patcon: Am on desktop already, so figured I'd make it a button-click for you @deepyaman :)

```suggestion
            "To opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment variables, "
            "or create a `.telemetry` file in the current working directory with the contents `consent: false`. "
            "To hide this message, explicitly grant or deny consent. "
```

deepyaman: Sorry, I got delayed/distracted and also just saw this. Appreciate it, but I realized I'd also forgotten to update the expected message in another place.

> I can also raise a PR later on the Kedro repo to add information of granting explicit consent.

@patcon If you feel like doing this, though, not going to stop you. :)

PR #1185, opened 2025-09-11
feat(datasets): SparkDataset Rewrite

SajidAlamQB:
## Description

This PR introduces `SparkDatasetV2`, a cleaner alternative to `SparkDataset` that addresses long-standing issues outlined in #135.

### Problems with Current SparkDataset
* Custom filepath parsing (`split_filepath` vs. `get_protocol_and_path`) causes inconsistencies
* Multiple code pathways (fsspec for metadata, Spark for I/O) work for S3/DBFS but break for others
* Forces installation of ~300MB dependencies even on Databricks, causing conflicts with databricks-connect
* Circular dependencies between Databricks datasets and SparkDataset utilities
* Incomplete test coverage and filepath handling bugs

## Development notes

**Dependency Improvements:**
* `spark-core` with zero dependencies - no forced PySpark on Databricks!
* Environment-specific installations: `spark-local`, `spark-databricks`, `spark-emr`
* Optional filesystem support: `spark-s3`, `spark-gcs`, `spark-azure`
* HDFS via PyArrow instead of deprecated hdfs package

**Code Improvements:**
* Lazy imports via `TYPE_CHECKING` 
* Consistent fsspec usage (keept the dbutils optimisation for DBFS performance, I can't test since databricks free doesn't allow use of DBFS)
* Proper protocol translation (s3:// → s3a://)
* Unity Catalog volumes support (`/Volumes/...`)
* Spark Connect compatibility
* Spark Connect and Databricks Connect compatibility
* Automatic Pandas → Spark DataFrame conversion (like `SnowparkTableDataset`)
* Session creation inside dataset no `SparkHooks` required
* Spark Session Handling:
`get_spark_with_remote_support()` automatically detects environment:
- Databricks Connect (when `DATABRICKS_HOST` and `DATABRICKS_TOKEN` set and databricks-connect installed)
- Spark Connect (when `SPARK_REMOTE` set)
- Classic local SparkSession (fallback)

### Current Status
- Local development with `spark-local`
- Unity Catalog volumes on Databricks
- S3/GCS/Azure storage via optional dependencies
- Spark Connect (Spark 4.0+)

**Known Limitations:**
- DBFS public root (`/FileStore`) is deprecated by Databricks not very easy to test if legacy works.

### Breaking Changes
- Users of `kedro-datasets[spark]` must choose specific bundles
- HDFS users need to use `spark-hdfs` with PyArrow

Now:

- Databricks users: No more PySpark conflicts with databricks-connect
- Reduced installation size: ~310MB → 0MB for cloud environments
- Clearer installation paths based on environment
- Users relying on `kedro-datasets[spark]` will need to choose specific bundles
- HDFS support is deprecated (still available via `spark-hdfs`)

## Checklist


kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,309 @@
...
+        try:
+            import fsspec  # noqa: PLC0415
+        except ImportError:
+            raise ImportError("fsspec is required")
+
+        # Normalise protocols
+        protocol_map = {
+            "s3a": "s3",
+            "s3n": "s3",  # Spark S3 variants
+            "dbfs": "file",  # DBFS is mounted as local
```
deepyaman: I think need to be very careful here. https://github.com/kedro-org/kedro/commit/6bf1066463fb4c9d6f1f572996f59fd86e345730 added `_dbfs_glob` because regular glob (i.e. treating DBFS paths as local) is—or at least was—_extremely_ slow on Databricks. There's history somewhere that I can't find right now; maybe it's on `private-kedro` or you can dig through historical KED-1458 JIRA issue for context? I'm 99% sure I created a PR with more details/investigation, and then Andrii ended up merging a separate PR.

kedro-datasets/pyproject.toml
```diff
@@ -18,15 +18,45 @@ dynamic = ["readme", "version"]
...
+spark-core = []  # No dependencies
+spark-local = ["pyspark>=2.2,<4.0"]  # For local development
+spark-databricks = []  # Uses Databricks runtime Spark
+spark-emr = []  # Uses EMR runtime Spark
+
+# Filesystem specific packages for Spark
+spark-s3 = ["s3fs>=2021.4"]
+spark-gcs = ["gcsfs>=2023.1, <2023.7"]
+spark-azure = ["adlfs>=2023.1"]
+spark-hdfs = ["pyarrow>=7.0"]  # PyArrow includes HDFS support
```
deepyaman, line 40: 🆒

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,309 @@
...
+)
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame, SparkSession
+    from pyspark.sql.types import StructType
+
+logger = logging.getLogger(__name__)
+
+
+class SparkDatasetV2(AbstractVersionedDataset):
```
ravi-kumar-pilla, line 43: Hi @SajidAlamQB , In general the code looks good to me. I am not sure about the history of changes made as mentioned by @deepyaman . I am testing this dataset and will update here if there are any issues.

Regarding the rewrite vs refactor, as discussed I think it is a good idea to have a separate version to avoid breaking changes. At the same time keeping in mind if we want to continue maintain both versions in future ? 

May be an input from @noklam on these changes would help understand better. Thank you

deepyaman: > Regarding the rewrite vs refactor, as discussed I think it is a good idea to have a separate version to avoid breaking changes.

@ravi-kumar-pilla Yeah, sorry if I wasn't clear. If there needs to be breaking changes, absolutely agree it should be a separate version; I just didn't see so much potential breaking changes (removing HDFS dependency doesn't _have_ to be breaking). Maybe I treated the `fsspec` changes too lightly when I looked through the PR; I got the impression it was mostly bugfixing to map the correct protocols, in addition to removing the custom logic to get protocol and such, but didn't understand it to be an overall behavior change.

SajidAlamQB:
> * When should I use `databricks` specific dataset and spark.SparkDataset on Databricks? I recalled there are already something that is only possible with the databricks one. If we are re-writing this I think we should have a look at this.
> * dbfs is a bit annoying - Databricks already deprecated it, new cluster are default to UC's volume but still a lot of people are using dbfs in older cluster.
> * Is there a goal/additional things that this rewrite improve? Or is it more like refactoring?

Hey @noklam thanks,

I think the Databricks datasets are more for TABLE operations while the SparkDataset is for FILE operations.

The new V2 handles both DBFS and UC Volumes properly, they still supports /dbfs/, dbfs:/, and /Volumes/ paths and we only do the DBFS optimisations only when needed.

This goes a bit beyond a refactor I think, we solving some long standing issues such as the Databricks users can now actually use it, we add Spark Connect for Spark 4.0 and now the users can choose their deps instead of installing everything via pyproject.toml changes. It makes the dataset more usable.

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,459 @@
...
+    dbfs_glob,
+    deployed_on_databricks,
+    get_dbutils,
+    strip_dbfs_prefix,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class SparkDatasetV2(AbstractVersionedDataset):
```
ravi-kumar-pilla, line 43: We need to mention this dataset in `spark/__init__.py` too if we go with v2.

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,459 @@
...
+            )
+            and not any(
+                filepath.startswith(f"{p}://")
+                for p in ["s3", "s3a", "s3n", "gs", "abfs", "wasbs"]
+            )
+        ):
+            logger.warning(
+                "Using SparkDatasetV2 on Databricks without the `/dbfs/`, `dbfs:/`, or `/Volumes` prefix "
+                "in the filepath is a known source of error. You must add this prefix to %s",
+                filepath,
```
merelcht: As far as I know, Databricks isn't recommending using the `dbfs` file format anymore, so we should make sure this dataset is also up to date with the latest practice on Databricks and probably remove this error message.

ravi-kumar-pilla:
Hi @SajidAlamQB ,

I have done some basic testing on databricks. Below is my notebook which I tried after a basic setup using UC volumes:

```python
from kedro_datasets.spark import SparkDataset, SparkDatasetV2

test_catalog = "spark_v2_catalog"
test_schema = "default"
volume_name = "spark_test_volume"

uc_base_path = f"/Volumes/{test_catalog}/{test_schema}/{volume_name}"
temp_dir = f"{uc_base_path}/basic"
dbutils.fs.mkdirs(temp_dir)

# Test 1: Parquet format (fails with SparkDatasetV2 but passes with SparkDataset)
parquet_path = f"{temp_dir}/basic_test.parquet"
dataset_parquet = SparkDatasetV2(filepath=parquet_path, file_format="parquet")
    
start_time = time.time()
dataset_parquet.save(test_df)
save_time = time.time() - start_time
print(f"✅ Saved parquet in {save_time:.3f}s to: {parquet_path}")
    
start_time = time.time()
loaded_df = dataset_parquet.load()
load_time = time.time() - start_time
print(f"✅ Loaded parquet in {load_time:.3f}s - {loaded_df.count()} rows")
    
# Verify data integrity
assert test_df.count() == loaded_df.count(), "Row count mismatch"
print("✅ Data integrity verified")

# Test 2: CSV with custom args (fails with SparkDatasetV2 but passes with SparkDataset)
print("\n📁 Test 2: CSV with Custom Arguments")
csv_path = f"{temp_dir}/basic_test.csv"
dataset_csv = SparkDataset(
    filepath=csv_path, 
    file_format="csv",
    save_args={"header": True, "sep": "|"},
    load_args={"header": True, "sep": "|", "inferSchema": True}
)

dataset_csv.save(test_df)
loaded_csv_df = dataset_csv.load()
print(f"✅ CSV save/load successful - {loaded_csv_df.count()} rows")
loaded_csv_df.show(3)

# Test 3: JSON format (fails with SparkDatasetV2 but passes with SparkDataset)
print("\n📁 Test 3: JSON Format")
json_path = f"{temp_dir}/basic_test.json"
dataset_json = SparkDataset(filepath=json_path, file_format="json")

# DatasetError: Failed while saving data to dataset kedro_datasets.spark.spark_dataset_v2.SparkDatasetV2(filepath='file:///Volumes/spark_v2_catalog/default/spark_test_volume/basic/basic_test.csv', file_format='csv', load_args={'header': True, 'sep': '|', 'inferSchema': True}, save_args={'header': True, 'sep': '|'}, protocol='file').
# (java.lang.SecurityException) Cannot use com.databricks.backend.daemon.driver.WorkspaceLocalFileSystem - local filesystem access is forbidden
```

**Observations:**
1. For UC volumes there is no issue with SparkDataset but SparkDatasetV2 fails with the error`local filesystem access is forbidden`. Detailed error in the notebook
2. For DBFS both SparkDataset and SparkDatasetV2 fails with the error `Public DBFS root is disabled. Access is denied on path`. I am not sure if I am doing something wrong. This is my path - `dbfs:/FileStore/spark_dataset_test/` and `/dbfs/FileStore/spark_dataset_test/`
3. `%pip install hdfs s3fs` is required for SparkDataset but SparkDatasetV2 does not require them (dependency restructuring works great)
4. Code looks good to me.

Thank you

kedro-datasets/kedro_datasets/_utils/databricks_utils.py
```diff
@@ -103,3 +108,102 @@ def dbfs_exists(pattern: str, dbutils: "DBUtils") -> bool:
...
+    Args:
+        filepath: Original filepath (used if protocol/path not provided).
+        protocol: Detected protocol.
+        path: Path component without protocol.
+
+    Returns:
+        Spark-compatible path string.
+    """
+    if not protocol:
+        protocol, path = parse_spark_filepath(filepath)
```
DimedS: could you comment please in what cases it can be `not protocol `if you made that parsing in `init()`?

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,321 @@
...
+            data: PySpark dataframe to save.
+        """
+        save_path = self._get_save_path()
+        spark_save_path = to_spark_path(str(save_path))
+
+        # Prepare writer
+        writer = data.write
+
+        # Apply mode if specified
+        mode = self._save_args.pop("mode", None)
```
DimedS: why are you mutating `_save_args` here?

kedro-datasets/kedro_datasets/_utils/databricks_utils.py
```diff
@@ -103,3 +108,99 @@ def dbfs_exists(pattern: str, dbutils: "DBUtils") -> bool:
...
+        and not any(
+            filepath.startswith(f"{p}://")
+            for p in ["s3", "s3a", "s3n", "gs", "abfs", "wasbs"]
+        )
+    ):
+        logger.warning(
+            "Using SparkDataset on Databricks without the `/dbfs/`, `dbfs:/`, or `/Volumes` prefix "
+            "in the filepath is a known source of error. You must add this prefix to %s",
+            filepath,
+        )
```
merelcht, line 176: Is this warning still valid?

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,320 @@
...
+    def _load(self) -> DataFrame:
+        """Loads data from filepath.
+
+        Returns:
+            Data from filepath as pyspark dataframe.
+        """
+        load_path = self._get_load_path()
+        spark_load_path = to_spark_path(str(load_path), self._protocol, str(load_path))
+
+        spark = get_spark_with_remote_support()
```
merelcht: Nit: maybe call this `spark_session` to make it clearer what this is.

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,320 @@
...
+
+        # Save with remaining options
+        writer.format(self._file_format).options(**save_args).save(spark_save_path)
+
+    def _exists(self) -> bool:
+        load_path = self._get_load_path()
+        spark_load_path = to_spark_path(str(load_path), self._protocol, str(load_path))
+
+        try:
+            spark = get_spark_with_remote_support()
```
merelcht: Same comment as earlier, call this `spark_session` instead.

kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -0,0 +1,320 @@
...
+            protocol: Filesystem protocol.
+            filepath: Original filepath.
+            credentials: Credentials for filesystem access.
+
+        Returns:
+            Tuple of (exists_function, glob_function).
+        """
+        if protocol == "dbfs" and deployed_on_databricks():
+            try:
+                spark = get_spark_with_remote_support()
```
merelcht: Same here

kedro-datasets/pyproject.toml
```diff
@@ -27,7 +27,38 @@ plotly-base = ["plotly>=4.8.0, <6.0"]
 delta-base = ["delta-spark>=1.0, <5.0"]
 networkx-base = ["networkx~=3.4"]
 
-# Individual Datasets
+# Spark dependencies
+spark-core = []  # No dependencies
```
merelcht, line 31: Are these empty definitions needed?

kedro-datasets/pyproject.toml
```diff
@@ -27,7 +27,38 @@ plotly-base = ["plotly>=4.8.0, <6.0"]
...
+spark-local = ["pyspark>=2.2,<4.0"]  # For local development
```
merelcht: Why is this `<4.0` when above spark-base is set to `<5.0`

kedro-datasets/pyproject.toml
SajidAlamQB: `spark-core` is a base that other packages can depend on without pulling in any specific Spark implementation. This allows users on Databricks to avoid installing PySpark since it's already in the runtime. We can add a comment to clarify this.

kedro-datasets/kedro_datasets/_utils/databricks_utils.py
SajidAlamQB: Yes, the warning is still valid on Databricks, using a plain path like /data/file.parquet without the proper prefix will fail because, DBFS paths need `/dbfs/,` Unity Catalog volumes need `/Volumes/catalog/schema/volume/...`
Cloud storage needs the appropriate protocol (`s3a://,` `gs://`, `abfs://`, etc.)

SajidAlamQB:
> These changes look good to me!
> 
> Just a couple of thoughts/questions not related to the code:

Hi, thank you for the feedback @merelcht!

> I'm thinking it might be helpful to add a comment somewhere for users on which SparkDataset they should use when.

Good idea I'll add guidance in the docstring and our docs. In general I think:

Use `SparkDatasetV2` when:
- Starting a new project
- Using Unity Catalog Volumes (`/Volumes/...`)
- Using Databricks with databricks-connect (avoids PySpark conflicts)
- You want cleaner dependency management (install only what you need)
- Using Spark Connect (Spark 3.4+)

    
Use `SparkDataset` (legacy) when:
- You have existing pipelines that depend on its specific behavior
- You need HDFS support via the hdfs package (V2 uses `PyArrow` instead)

For the most part `SparkDatasetV2` can be dropin replacement.

> Related to the above: Should we keep the old SparkDataset? And if not, I'd vote for removing it in the next major release.

I agree we should for now add `SparkDatasetV2` alongside `SparkDataset` then in the next minor patch add deprecation warning to `SparkDataset` pointing users to V2. Then in next major we can rename `SparkDatasetV2` to `SparkDataset`.

> The spark tests are not running on Windows because of the issues with the old implementation. Is it possible to run this new one on Windows?

`SparkDatasetV2` should work on Windows I will enable it and test it out.

DimedS:
Remote connections to Databricks should be established using Databricks Connect.
I’ve updated the implementation here: https://github.com/kedro-org/kedro-plugins/pull/1185/commits/f11f1c129f8b21376721496a4681567cc7b3ad61, and it now works correctly in remote mode.

PR #1188, opened 2025-09-16
build(telemetry): Release 0.6.5

DimedS:
## Description
Release to support #1178, which resolves [kedro-org/kedro#4999](https://github.com/kedro-org/kedro/issues/4999).


Huongg (approved):
thank you @DmitrySorokinQB

PR #1200, opened 2025-09-29
feat(datasets): Add LangchainPromptDataset to experimental datasets

lrcouto:
## Description
Solves https://github.com/kedro-org/kedro/issues/5089

### What is this?

A dataset that can be used to load `.txt`, `.json` and `.yaml` files and convert them into LangChain prompt objects. Currently works with `PromptTemplate` and `ChatPromptTemplate`.

I recommend using it with Python >= 3.10 for safety/compatibility, and LangChain >= 0.3.0.

### Expected data format

`PromptTemplate` expects a string by default. It accepts a set of parameters from the user that can be used to generate a prompt for a language model. For example:

`Hello {name}, welcome to Kedro!`

Or with a defined list of input variables:

```json
{
  "template": "You are an expert in {field}. Answer the following question: {question}",
  "input_variables": ["field", "question"]
}
```
Or in YAML format, same thing can be done:

```yaml
template: |
  Context: {context}
  
  Task: {task}
  
  Please provide a {detail_level} response.
input_variables:
  - context
  - task
  - detail_level
  ```

`ChatPromptTemplate` expects as input a dictionary or a list of tuples, with pairs of role and content parameters. For example:

```py
[
    ("system", "You are a helpful AI bot. Your name is {name}."),
    ("human", "Hello, how are you doing?"),
    ("ai", "I'm doing well, thanks!"),
    ("human", "{user_input}"),
]
```

or

```json
{
  "messages": {
    "system": "You are a coding assistant specialized in {language}.",
    "human": "Help me with: {problem}"
  } 
}
```

In yaml:

```yaml
messages:
  - role: system
    content: "You are a {role}."
  - role: human
    content: "Please help with: {request}"
  - role: ai
    content: "I'll help you with {request}. Let me break it down."
  - role: human
    content: "Thanks! Can you also explain {additional_topic}?"
```

For further detail see the LangChain documentation - 
PromptTemplate: https://python.langchain.com/v0.2/api_reference/core/prompts/langchain_core.prompts.prompt.PromptTemplate.html
ChatPromptTemplate: https://python.langchain.com/v0.2/api_reference/core/prompts/langchain_core.prompts.chat.ChatPromptTemplate.html

### Data Catalog configuration.

Example:

```yaml
json_chat_prompt:
  type: langchain_prompt_dataset.datasets.langchain_prompt_dataset.LangChainPromptDataset
  filepath: data/prompts/chat_simple.json
  template: ChatPromptTemplate
  dataset:
    type: json.JSONDataset
    fs_args:
      load_args:
        encoding: utf-8
      save_args:
        ensure_ascii: false
  credentials: dev_creds
  metadata:
    kedro-viz:
      layer: raw
```

- `filepath`: Path to the dataset file.
- `template`: Which LangChain template should be used for that dataset, either `PromptTemplate` or `ChatPromptTemplate`. If none is chosen, `PromptTemplate` is the default.
- `dataset`: Arguments related to which underlying dataset is chosen.
  - `type`: Which underlying dataset should be used to load the file. Can be `text.TextDataset`, `json.JSONDataset` or `yaml.YAMLDataset`. If none is chosen, it'll be inferred through the file extension.
    - `fs_args`: If the chosen underlying dataset allows for extra arguments, they can be passed here.
- `credentials`:  Will be passed to the underlying dataset. Work the same as whichever dataset is chosen.

The `kedro-viz: layer: raw` parameter in the metadata allow the data preview to be displayed in the Kedro Viz metadata panel (https://github.com/kedro-org/kedro-viz/pull/2490)

Looks like this:

image

### Node example:

You can take a YAML file like:

```yaml
messages:
  - role: system
    content: "You are a {role}."
  - role: human
    content: "Please help with: {request}"
  - role: ai
    content: "I'll help you with {request}. Let me break it down."
  - role: human
    content: "Thanks! Can you also explain {additional_topic}?"
```

And pass it to a Kedro node, where the inputs for those variables can be passed using the `format_messages` method, which returns a list of LangChain [BaseMessages](https://python.langchain.com/v0.2/api_reference/core/messages/langchain_core.messages.base.BaseMessage.html#langchain_core.messages.base.BaseMessage),

```py
def test_yaml_chat_prompt(yaml_chat_prompt: ChatPromptTemplate) -> Dict[str, str]:
    """Test YAML chat prompt."""
    messages = yaml_chat_prompt.format_messages(
        role="technical writer",
        request="documentation for API endpoints",
        additional_topic="best practices for API versioning"
    )
```

The output of `messages` will be like this:

```py
[
    SystemMessage(
        content='You are a technical writer.',
        additional_kwargs={},
        response_metadata={}
    ),
    HumanMessage(
        content='Please help with: documentation for API endpoints',
        additional_kwargs={},
        response_metadata={}
    ),
    AIMessage(
        content="I'll help you with documentation for API endpoints. Let me break it down.",
        additional_kwargs={},
        response_metadata={}
    ),
    HumanMessage(
        content='Thanks! Can you also explain best practices for API versioning?',
        additional_kwargs={},
        response_metadata={}
    )
]
```

Which is the right format to be fed into one of LangChain's chat models.

## Checklist


kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+class LangChainPromptDataset(AbstractDataset[PromptTemplate | ChatPromptTemplate, Any]):
+    """Kedro dataset for loading LangChain prompts using existing Kedro datasets."""
+
+
+    TEMPLATES = {
+        "PromptTemplate": PromptTemplate,
+        "ChatPromptTemplate": ChatPromptTemplate,
+    }
+
+    def __init__(
```
ElenaKhaustova: Add credentials handling for the underlying dataset

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+# Minimum number of elements required for a message (role, content)
+MIN_MESSAGE_LENGTH = 2
+
+
+class LangChainPromptDataset(AbstractDataset[PromptTemplate | ChatPromptTemplate, Any]):
+    """Kedro dataset for loading LangChain prompts using existing Kedro datasets."""
+
+
+    TEMPLATES = {
+        "PromptTemplate": PromptTemplate,
```
ElenaKhaustova: Here we can use direct mapping to the target function

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+                return self._create_prompt_template(raw_data)
+        except Exception as e:
+            raise DatasetError(f"Failed to create {self._template_class.__name__}: {e}")
+
+    def _create_prompt_template(self, raw_data: Any) -> PromptTemplate:
+            """Create a PromptTemplate from loaded data."""
+            if isinstance(raw_data, str):
+                return PromptTemplate.from_template(raw_data)
+
+            if isinstance(raw_data, dict):
```
ElenaKhaustova: Looks like `from_template()` can handle this case as is: https://python.langchain.com/api_reference/_modules/langchain_core/prompts/prompt.html#PromptTemplate

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+                # If it has a template key, handle it specially, otherwise pass all kwargs
+                if "template" in raw_data and "input_variables" not in raw_data:
+                    return PromptTemplate.from_template(raw_data["template"])
+                return PromptTemplate(**raw_data)
+
+            raise DatasetError(f"Unsupported data type for PromptTemplate: {type(raw_data)}")
+
+    def _create_chat_prompt_template(self, raw_data: Any) -> ChatPromptTemplate:
```
ElenaKhaustova: Here, I would expect an input format as a list of tuples or a dictionary:

```yaml
    [
        ("system", "You are a helpful AI bot. Your name is {name}."),
        ("human", "Hello, how are you doing?"),
        ("ai", "I'm doing well, thanks!"),
        ("human", "{user_input}"),
    ]
```

or 

```yaml
{
        "system": "You are a helpful AI bot. Your name is {name}.",
        "human": "Hello, how are you doing?",
        "ai": "I'm doing well, thanks!",
}
```

ElenaKhaustova: We can create a validation function to handle the format and then use `ChatPromptTemplate()` or  `ChatPromptTemplate.from_messages()`.

https://python.langchain.com/api_reference/core/prompts/langchain_core.prompts.chat.ChatPromptTemplate.html

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+        try:
+            raw_data = self._dataset.load()
+        except Exception as e:
+            raise DatasetError(f"Failed to load data from {self._filepath}: {e}")
+
+        if raw_data is None:
+            raise DatasetError(f"No data loaded from {self._filepath}")
+
+        try:
+            if self._template_class == ChatPromptTemplate:
```
ElenaKhaustova: We don't need `self._template_class` to be a type; it can be a string. If we replace this mapping `TEMPLATES ` constant with the mapping to the function (as suggested above), then we would not need to have a condition here.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
ElenaKhaustova: I've reopened it as it doesn't seem to be done but still makes sense to me.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+                    for msg in messages]
+
+        if any(not isinstance(msg, tuple) or len(msg) != MIN_MESSAGE_LENGTH for msg in validated_data):
+            raise DatasetError(f"Unsupported message type found in messages: {validated_data}")
+
+        return validated_data
+
+    def _create_chat_prompt_template(self, data: dict | list[tuple[str, str]]) -> ChatPromptTemplate:
+        """Create a ChatPromptTemplate from validated data."""
+        messages = self._validate_chat_prompt_data(data)
```
ElenaKhaustova, line 286: The validation method should not change the input but only validate.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+
+        try:
+            if self._template_class == ChatPromptTemplate:
+                return self._create_chat_prompt_template(raw_data)
+            else:
+                return self._create_prompt_template(raw_data)
+        except Exception as e:
+            raise DatasetError(f"Failed to create {self._template_class.__name__}: {e}")
+
+    def _create_prompt_template(self, raw_data: Any) -> PromptTemplate:
```
ElenaKhaustova: We need to specify the type for the `raw_data` here

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+        """Create a PromptTemplate from loaded data."""
```
ElenaKhaustova: Can we be clear about the format we expect for this method and add it to the docstring function as an example?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+        if isinstance(raw_data, str):
+            return PromptTemplate.from_template(raw_data)
+
+        if isinstance(raw_data, dict):
+           return PromptTemplate(**raw_data)
+
+        raise DatasetError(f"Unsupported data type for PromptTemplate: {type(raw_data)}")
+
+    def _validate_chat_prompt_data(self, data: dict | list[tuple[str, str]]) -> list[tuple[str, str]]:
+        """Validate and normalize chat prompt data."""
```
ElenaKhaustova: The format is the same here. Can we be clear about the format we expect for this method and add it to the docstring function as an example?

I've also resolved the previous comment with the suggested formats.

ElenaKhaustova (changes requested):
Thank you, @lrcouto!

I made another pass and left several things to address.

Apart from them can we please also:
1. Add an example on how to test it as we do with normal datasets 
2. Spcify a minimum `LangChain` and `Python` versions this dataset will be supporting
3. Test it with the https://github.com/kedro-org/kedro-viz/pull/2490.  It will require `.preview()` method to be implemented.

@rashidakanchwala, can you please help us to build a branch with the preview and node colouring features, so we can use it when testing?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,169 @@
...
+from pathlib import Path
+from typing import Any
+
+from kedro.io import AbstractDataset, DatasetError
+from kedro.io.catalog_config_resolver import CREDENTIALS_KEY
+from kedro.io.core import get_filepath_str, parse_dataset_definition
+from langchain.prompts import ChatPromptTemplate, PromptTemplate
+
+# Minimum number of elements required for a message (role, content)
+MIN_MESSAGE_LENGTH = 2
```
ElenaKhaustova: Do we really need this? Can we maybe just validate that the message is not empty?

lrcouto: Yeah that's probably cleaner, I did this to circumvent a linter error.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: Do you think it'd be better to change the input if needed on the function that creates the template, or put it on another separate one?

ElenaKhaustova: I don't think we should change it at all.  We should allow formats that `ChatPromptTemplate` constructor or `from_messages()` expect. Or use just one of them - `ChatPromptTemplate`.

lrcouto:
@ElenaKhaustova everything should be corrected now.

For the format validation, I've left it only checking if the message isn't empty. If the data does not contain the correct format, the underlying dataset throws an exception (for example, bad JSON formatting), so it doesn't even load and doesn't even get to the validation function.

kedro-datasets/RELEASE.md
```diff
@@ -3,6 +3,12 @@
 
 - Group datasets documentation according to the dependencies to clean up the nav bar.
 
+- Added the following new **experimental** datasets:
```
ElenaKhaustova, line 6: We also need to update `pyproject.toml` to add requirements for this dataset.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        if isinstance(raw_data, str):
+            return PromptTemplate.from_template(raw_data)
+
+        if isinstance(raw_data, dict):
+           return PromptTemplate(**raw_data)
+
+        raise DatasetError(f"Unsupported data type for PromptTemplate: {type(raw_data)}")
+
+    def _validate_chat_prompt_data(self, data: dict | list[tuple[str, str]]) -> None:
```
ElenaKhaustova: Return type needs to be updated

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            "path": self._filepath,
+            "template": self._template_name,
+            "underlying_dataset": self._dataset.__class__.__name__,
+            "dataset_config": clean_config,
+        }
+
+    def _exists(self) -> bool:
+        return self._dataset._exists() if hasattr(self._dataset, "_exists") else True
+
+    def preview(self) -> JSONPreview:
```
ElenaKhaustova, line 306: For this method, I would suggest not to modify the content, like adding keys, etc. Instead we can use `HTMLPreview` added recently: https://github.com/kedro-org/kedro-viz/pull/2490

So, based on the underlying dataset type, we can return markdown content for text, JSON, or YAML wrapped according to the type.

```python
return f"```yaml\n{data}```"
```

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+
+from kedro.io import AbstractDataset, DatasetError
+from kedro.io.catalog_config_resolver import CREDENTIALS_KEY
+from kedro.io.core import get_filepath_str, parse_dataset_definition
+from langchain.prompts import ChatPromptTemplate, PromptTemplate
+
+from kedro_datasets._typing import JSONPreview
+
+
+class LangChainPromptDataset(AbstractDataset[PromptTemplate | ChatPromptTemplate, Any]):
```
ElenaKhaustova: Shall we use `Union` here instead? In case someone uses this dataset with a lower Python version.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            ...     "template": "Hello {name}!",
+            ...     "input_variables": ["name"]
+            ... })
+            PromptTemplate(template='Hello {name}!', input_variables=['name'])
+        """
+        if isinstance(raw_data, str):
+            return PromptTemplate.from_template(raw_data)
+
+        if isinstance(raw_data, dict):
+           return PromptTemplate(**raw_data)
```
ElenaKhaustova: Indentation mismatch (1 space less)

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+from kedro.io import AbstractDataset, DatasetError
+from kedro.io.catalog_config_resolver import CREDENTIALS_KEY
+from kedro.io.core import get_filepath_str, parse_dataset_definition
+from langchain.prompts import ChatPromptTemplate, PromptTemplate
+
+from kedro_datasets._typing import JSONPreview
+
+
+class LangChainPromptDataset(AbstractDataset[PromptTemplate | ChatPromptTemplate, Any]):
+    """Kedro dataset for loading LangChain prompts using existing Kedro datasets."""
```
ElenaKhaustova: Can we please extend class docstrings like we do for the rest of the datasets? It is used in the docs, so it should be quite informative.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                raise DatasetError(f"Cannot auto-detect dataset type for file: {self._filepath}")
+
+        dataset_config = dataset if isinstance(dataset, dict) else {"type": dataset}
+        dataset_config = deepcopy(dataset_config)
+        dataset_config["filepath"] = self._filepath
+
+        return dataset_config
+
+    def load(self) -> PromptTemplate | ChatPromptTemplate:
+        """Load data using underlying dataset and wrap in LangChain template."""
```
ElenaKhaustova: Public methods should have proper docstrings, with all the required sections like `Returns`.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        if isinstance(raw_data, str):
+            return PromptTemplate.from_template(raw_data)
+
+        if isinstance(raw_data, dict):
+           return PromptTemplate(**raw_data)
+
+        raise DatasetError(f"Unsupported data type for PromptTemplate: {type(raw_data)}")
+
+    def _validate_chat_prompt_data(self, data: dict | list[tuple[str, str]]) -> None:
+        """Validate that chat prompt data exists and is not empty."""
```
ElenaKhaustova: It should also mention that it returns the validated and unpacked messages.

ElenaKhaustova (commented):
Thank you @lrcouto! The implementation looks much cleaner now and thanks for extending the PR description with examples ✨

I've added a few comments to make it ready for merging!

After you add the requirements to the `pyproject.toml`, could you please also share installation commands in the PR description, so the reviewers could easily test it?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: The linter yells at me if I use Union =(

I changed and noqa'd it for now.

kedro-datasets/pyproject.toml
```diff
@@ -183,7 +183,8 @@ langchain-chatopenaidataset = ["langchain-openai~=0.1.7"]
 langchain-openaiembeddingsdataset = ["langchain-openai~=0.1.7"]
 langchain-chatanthropicdataset = ["langchain-anthropic~=0.1.13", "langchain-community~=0.2.0"]
 langchain-chatcoheredataset = ["langchain-cohere~=0.1.5", "langchain-community~=0.2.0"]
-langchain = ["kedro-datasets[langchain-chatopenaidataset,langchain-openaiembeddingsdataset,langchain-chatanthropicdataset,langchain-chatcoheredataset]"]
+langchain-langchainpromptdataset = ["langchain~=0.3.0"]
+langchain = ["kedro-datasets[langchain-chatopenaidataset,langchain-openaiembeddingsdataset,langchain-chatanthropicdataset,langchain-chatcoheredataset,langchain-langchainpromptdataset]"]
```
ElenaKhaustova: Let’s avoid including it in the main langchain group for now.

The existing langchain-* datasets (OpenAI, Anthropic, Cohere, etc.) are still compatible to the older LangChain 0.1–0.2.x ecosystem.

The new LangChainPromptDataset depends on langchain~=0.3.0, which introduces major breaking changes and will likely cause dependency conflicts if installed together.

kedro-datasets/pyproject.toml
```diff
@@ -183,7 +183,8 @@ langchain-chatopenaidataset = ["langchain-openai~=0.1.7"]
 langchain-openaiembeddingsdataset = ["langchain-openai~=0.1.7"]
 langchain-chatanthropicdataset = ["langchain-anthropic~=0.1.13", "langchain-community~=0.2.0"]
 langchain-chatcoheredataset = ["langchain-cohere~=0.1.5", "langchain-community~=0.2.0"]
-langchain = ["kedro-datasets[langchain-chatopenaidataset,langchain-openaiembeddingsdataset,langchain-chatanthropicdataset,langchain-chatcoheredataset]"]
+langchain-langchainpromptdataset = ["langchain~=0.3.0"]
```
ElenaKhaustova: Are we sure it's not gonna work with the `langchain~=0.2.0`?

lrcouto: I'll look through all the methods on the two classes we're using again to be sure. I wanted to play it a bit safe but it should work.

ElenaKhaustova: Can we maybe just test it out?

kedro-datasets/pyproject.toml
```diff
@@ -183,6 +183,7 @@ langchain-chatopenaidataset = ["langchain-openai~=0.1.7"]
 langchain-openaiembeddingsdataset = ["langchain-openai~=0.1.7"]
 langchain-chatanthropicdataset = ["langchain-anthropic~=0.1.13", "langchain-community~=0.2.0"]
 langchain-chatcoheredataset = ["langchain-cohere~=0.1.5", "langchain-community~=0.2.0"]
+langchain-langchainpromptdataset = ["langchain~=0.2.0"]
```
ElenaKhaustova: So this version is working fine?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,284 @@
...
+from kedro.io.catalog_config_resolver import CREDENTIALS_KEY
+from kedro.io.core import get_filepath_str, parse_dataset_definition
+from langchain.prompts import ChatPromptTemplate, PromptTemplate
+
+from kedro_datasets._typing import JSONPreview
+
+
+class LangChainPromptDataset(AbstractDataset[Union[PromptTemplate, ChatPromptTemplate], Any]): # noqa UP007
+    """
+    A Kedro dataset for loading LangChain prompt templates from text, JSON, or YAML files.
```
ElenaKhaustova, line 16: There also has to be a usage example, so it's included in the docs like here: https://kedro--1200.org.readthedocs.build/projects/kedro-datasets/en/1200/api/kedro_datasets_experimental/langchain.ChatOpenAIDataset/

ElenaKhaustova:
> Regarding the preview() method

I see, that makes sense! Let's stick to the current solution then, as discussed.

And if the following is the case, we still need to open a ticket for Viz:

> I think Kedro-Viz cannot handle a preview function that returns both HTMLPreview and JSONPreview, if I try to mix output types it doesn't render anything.

> Returning everything as HTMLPreview causes the text to truncate on the preview window if it's too long, and it's not displaying syntax highlighting even if the correct yaml or json tags are passed on the code block.

kedro-datasets/kedro_datasets_experimental/langchain/__init__.py
```diff
@@ -7,19 +7,23 @@
...
     ChatCohereDataset: Any
+    LangChainPromptDataset: Any
 
 __getattr__, __dir__, __all__ = lazy.attach(
     __name__,
     submod_attrs={
         "_openai": ["ChatOpenAIDataset", "OpenAIEmbeddingsDataset"],
         "_anthropic": ["ChatAnthropicDataset"],
         "_cohere": ["ChatCohereDataset"],
+        "_langchain_prompt_dataset": ["LangChainPromptDataset"],
```
ElenaKhaustova:
```suggestion
        "langchain_prompt_dataset": ["LangChainPromptDataset"],
```

kedro-datasets/pyproject.toml
lrcouto: I've tested it yesterday and didn't find any different behaviors. Double checked the langchain docs for any deprecation notices or anything, so I think it'll be OK.

ElenaKhaustova: Nice, thank you!

kedro-datasets/kedro_datasets_experimental/langchain/__init__.py
```diff
@@ -7,19 +7,23 @@
...
     ChatCohereDataset: Any
+    LangChainPromptDataset: Any
 
 __getattr__, __dir__, __all__ = lazy.attach(
     __name__,
     submod_attrs={
         "_openai": ["ChatOpenAIDataset", "OpenAIEmbeddingsDataset"],
         "_anthropic": ["ChatAnthropicDataset"],
         "_cohere": ["ChatCohereDataset"],
+        "langchain_prompt_dataset": ["LangChainPromptDataset"],
```
ankatiyar, line 27: QQ: Why're the other ones starting with an underscore and not this one?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,307 @@
...
+            fs_args: Optional filesystem arguments for remote storage.
+        credentials: Optional credentials for accessing remote filesystems.
+        metadata: Arbitrary metadata for catalog introspection.
+
+    Example YAML catalog:
+        my_prompt:
+          type: kedro_datasets_experimental.langchain.LangChainPromptDataset
+          filepath: data/prompts/my_prompt.json
+          template: PromptTemplate
+            dataset:
```
ankatiyar: The indentation here is wonky I think

kedro-datasets/kedro_datasets_experimental/langchain/__init__.py
lrcouto: My question would be why these datasets specifically start with an underscore when none of the other experimental datasets do. 🤔

ElenaKhaustova: The rest of the modules have names with an underscore. I'm not sure what the main purpose was—probably to prevent people from importing from them directly.

lrcouto: Yes, but only the LangChain ones. Is there anything specific about them?

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,315 @@
...
+
+class LangChainPromptDataset(AbstractDataset[Union[PromptTemplate, ChatPromptTemplate], Any]): # noqa UP007
+    """
+    A Kedro dataset for loading LangChain prompt templates from text, JSON, or YAML files.
+
+    This dataset wraps existing Kedro datasets (such as TextDataset, JSONDataset, or YAMLDataset)
+    to load prompt configurations and convert them into LangChain `PromptTemplate` or
+    `ChatPromptTemplate` objects.
+
+    Args:
```
ElenaKhaustova: It looks like something is still wrong with this class docstring cause it's not rendered properly with the docs: https://kedro--1200.org.readthedocs.build/projects/kedro-datasets/en/1200/api/kedro_datasets_experimental/langchain.LangChainPromptDataset/

I suggest following existing examples like https://kedro--1200.org.readthedocs.build/projects/kedro-datasets/en/1200/api/kedro_datasets_experimental/langchain.LangChainPromptDataset/ or https://kedro--1200.org.readthedocs.build/projects/kedro-datasets/en/1200/api/kedro_datasets_experimental/polars.PolarsDatabaseDataset/#kedro_datasets_experimental.polars.PolarsDatabaseDataset--example-usage-for-the-yaml-api

kedro-datasets/pyproject.toml
ElenaKhaustova: `~=` means compatible with version `0.2.0` that is, `≥0.2.0` and `<0.3.0`
```suggestion
langchain-langchainpromptdataset = ["langchain>=0.2.0"]
```

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        Validate that chat prompt data exists and is not empty.
+        Raises an error if data is a plain string, which is only compatible with PromptTemplate.
+
+        Returns validated and unpacked messages as a dictionary or a list of tuples.
+
+        Raises:
+            DatasetError: If the data is empty or is a plain string.
+        """
+        if isinstance(data, str):
```
ElenaKhaustova, line 241: Tested this on the example shared before and it works as expected

ElenaKhaustova (commented):
I've tested recent changes - all works as expected.

There are still issues with the version and class docstrings. Happy to approve when they're resolved.

kedro-datasets/kedro_datasets_experimental/langchain/__init__.py
ankatiyar: Okay this makes sense, they probably just wanted to split out the different datasets into different files but keep the import still `from kedro_dataset_experimental.langchain import ...`

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,315 @@
...
+            else:
+                dataset_config["fs_args"] = deepcopy(self._fs_args)
+
+        try:
+            dataset_class, dataset_kwargs = parse_dataset_definition(dataset_config)
+            self._dataset = dataset_class(**dataset_kwargs)
+        except Exception as e:
+            raise DatasetError(f"Failed to create underlying dataset: {e}")
+
+    def _build_dataset_config(self, dataset: dict[str, Any] | str | None) -> dict[str, Any]:
```
ankatiyar, line 141: I wonder if it's better to specify strictly what the underlying datasets can be - just `TextDataset`, `YAMLDataset` and `JSONDataset` and error out if it isn't instead of inferring it. Unlike `PartitionedDataset` it's not like this can be any underlying dataset type

ElenaKhaustova: Well, yeah, it makes sense cause now one can set a random dataset that will load data incompatible with the langchain template.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: Ah I see, it needs to be encased in markdown tags.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: Added some validation for that.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Returns:
+            dict: A normalized dataset configuration dictionary.
+        """
+
+        valid_datasets = {"text.TextDataset", "json.JSONDataset", "yaml.YAMLDataset"}
+
+        if dataset is not None:
+            dataset_type = dataset["type"] if isinstance(dataset, dict) else str(dataset)
+            if dataset_type not in valid_datasets:
```
ElenaKhaustova: This will not work if the user sets the dataset using the full name, like `kedro_datasets.text.TextDataset`

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+                  ensure_ascii: false
+            credentials: dev_creds
+            metadata:
+              kedro-viz:
+                  layer: raw
+    ```
+
+    ### Example usage for the [Python API](https://docs.kedro.org/en/stable/catalog-data/advanced_data_catalog_usage/):
+    ```python
+        >>> from kedro_datasets_experimental.langchain import LangChainPromptDataset
```
ElenaKhaustova: We probably don't need `>>>` if wrapping into the python code block

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+class LangChainPromptDataset(AbstractDataset[Union[PromptTemplate, ChatPromptTemplate], Any]): # noqa UP007
+    """
+    A Kedro dataset for loading LangChain prompt templates from text, JSON, or YAML files.
+
+    This dataset wraps existing Kedro datasets (such as TextDataset, JSONDataset, or YAMLDataset)
+    to load prompt configurations and convert them into LangChain `PromptTemplate` or
+    `ChatPromptTemplate` objects.
+
+    ### Example usage for the [YAML API](https://docs.kedro.org/en/stable/catalog-data/data_catalog_yaml_examples/):
+    ```yaml
```
ElenaKhaustova, line 23: Can we please shift it left, so the indentation looks better when rendered like here: https://kedro--1200.org.readthedocs.build/projects/kedro-datasets/en/1200/api/kedro_datasets_experimental/langchain.OpenAIEmbeddingsDataset/#kedro_datasets_experimental.langchain.OpenAIEmbeddingsDataset?

lrcouto: Should be good now. I also changed the indent to 4 spaces since that's what the example is using.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
ankatiyar: What I meant was, instead of allowing for a case where the user hasn't specified an underlying `dataset` config and inferring from file extension, we can just error out if the `dataset` config is not provided. And if the config is provided, maybe we can check if the `type` is only `TextDataset` `YAMLDataset` or `JSONDataset` (on further discussion with @ElenaKhaustova, this might limit the user incase they wanted to use a custom underlying dataset so I am not too fussed about if we include this validation or not. If the data is not in the correct format `langchain` should complain anyway. But it might be fine to limit this to these types for now)

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if dataset is not None:
+            dataset_type = dataset["type"] if isinstance(dataset, dict) else str(dataset)
+            if dataset_type not in valid_datasets:
+                raise DatasetError(
+                    f"Unsupported dataset type '{dataset_type}'. "
+                    f"Allowed dataset types are: {', '.join(valid_datasets)}"
+                )
+
+        if dataset is None:
```
ankatiyar: re: the above comment, this bit I don't think is necessary, just error out if `dataset` is none.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+            kedro-viz:
+                layer: raw
+    ```
+
+    ### Example usage for the [Python API](https://docs.kedro.org/en/stable/catalog-data/advanced_data_catalog_usage/):
+    ```python
+        from kedro_datasets_experimental.langchain import LangChainPromptDataset
+        dataset = LangChainPromptDataset(filepath="data/prompts/my_prompt.json", template="PromptTemplate")
+        prompt = dataset.load()
+        print(prompt.format(name="Kedro"))
```
ankatiyar: It would be nicer to add blank lines and indentation here to make it more readable.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: That'd make the `dataset: type:` entry in the catalog mandatory. Do you think that'd be better?

The way it is right now you can quickly throw in just the filepath and type and have it work.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
lrcouto: I like the idea of allowing users to use a custom underlying dataset but I think this can be a future addition. I'd like to see if this dataset is something that people are actually interested in using first.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
ankatiyar: Yeah that's what I'm suggesting. We shouldn't be doing any inference - it'll be better if the user is explicit in their config rather than us assuming

lrcouto: Alright, applied the changes. It's mandatory now.

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,329 @@
...
+                    ensure_ascii: false
+        credentials: dev_creds
+        metadata:
+            kedro-viz:
+                layer: raw
+    ```
+
+    ### Example usage for the [Python API](https://docs.kedro.org/en/stable/catalog-data/advanced_data_catalog_usage/):
+    ```python
+        from kedro_datasets_experimental.langchain import LangChainPromptDataset
```
ElenaKhaustova: Should help with the indentation a bit

```suggestion
    ```python
    from kedro_datasets_experimental.langchain import LangChainPromptDataset
    
    dataset = LangChainPromptDataset(
        filepath="data/prompts/my_prompt.json",
        template="PromptTemplate",
        dataset={"type": "json.JSONDataset"}
    )
    prompt = dataset.load()
    print(prompt.format(name="Kedro"))
    ```
```

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,329 @@
...
+
+        Raises:
+            DatasetError: If the dataset type is unsupported.
+            Currently supported dataset types are: text.TextDataset, json.JSONDataset, yaml.YAMLDataset
+
+        Returns:
+            dict: A normalized dataset configuration dictionary.
+        """
+
+        valid_datasets = {"text.TextDataset", "json.JSONDataset", "yaml.YAMLDataset"}
```
ElenaKhaustova: Nit: would be a bit cleaner if this became a constant and the validation logic is moved to a separate method

kedro-datasets/kedro_datasets_experimental/langchain/langchain_prompt_dataset.py
```diff
@@ -0,0 +1,329 @@
...
+
+        if dataset is None:
```
ElenaKhaustova: Can we please also add at least two unit tests to check that proper errors are raised?

PR #1201, opened 2025-09-30
test docker ci

SajidAlamQB:
## Checklist

### DIFF
--- a/kedro-docker/features/steps/cli_steps.py
+++ b/kedro-docker/features/steps/cli_steps.py
@@ -1,4 +1,4 @@
-"""Behave step definitions for the cli_scenarios feature."""
+"""Behave step definitions for the cli_scenarios feature. test"""
 
 import re
 import sys


deepyaman (approved):
Thanks for addressing this long-running issue!

kedro-docker/kedro_docker/template/Dockerfile.spark
```diff
@@ -5,6 +5,16 @@ FROM $BASE_IMAGE as runtime-environment
...
 
+# set JVM options to allow Unsafe operations
+ENV JAVA_OPTS="-Dio.netty.tryReflectionSetAccessible=true \
+    --add-opens=java.base/java.nio=ALL-UNNAMED \
+    --add-opens=java.base/sun.nio.ch=ALL-UNNAMED \
+    --add-opens=java.base/jdk.internal.misc=ALL-UNNAMED"
+
+# set Spark configuration to handle Arrow properly
+ENV SPARK_SUBMIT_OPTS="-Dio.netty.tryReflectionSetAccessible=true"
+ENV ARROW_PRE_0_15_IPC_FORMAT=1
```
deepyaman, line 9: This is ridiculously old--like 6 years ago. I can't imagine people should be on pre-0.15 Arrow in reality. Should this need to be done?

deepyaman: Looking more into it, it _should not_ be required when working with Spark 3. What version of Spark are the e2e tests using?

SajidAlamQB: You're right that `ARROW_PRE_0_15_IPC_FORMAT=1` is outdated but after testing without it the same `sun.misc.Unsafe` error to returns.

Is the flag effecting more than just IPC format compatibility?

deepyaman: > Is the flag effecting more than just IPC format compatibility?

Diving into the PySpark source, I would expect it throw an error if you're using Spark 3: https://github.com/apache/spark/blob/ee8d66105885929ac0c0c087843d70bf32de31a1/python/pyspark/sql/pandas/utils.py#L58-L60

I wouldn't be surprised if this is still necessary for Spark 2; Spark 3 really pushed to be more Arrow-native.

I think it's worth confirming:

1. Is Spark 2 used in the tests? I couldn't tell easily just from scanning the code and looking at the CI logs.
2. If so, why is Spark 2 being used? As far as I can tell, the starter should get you set up including Spark 3.

SajidAlamQB: Yeh this is confusing, I will keep digging into this a bit more.

PR #1205, opened 2025-10-02
fix(datasets): default `plotly.JSONDataset` save encoding to UTF-8

### LINKED ISSUE #741: plotly.JSONDataset not saved as utf-8
## Description
It can happen that the saved plotly.JSONDataset is not encoded as utf-8. Supplying the file system args as follows fixes the issue:
```yml 
myfile:
  filepath: myfilepath
  fs_args:
    open_args_save:
      encoding: utf-8
```
However, that should be the default behaviour. The question is why there is a problem when the encoding is not explicitly set here.

## Context
I had an issue with the encoding of saved plotly plots (as json) via the kedro data catalog. After saving the plots I could not read the plots anymore via the catalog. It failed with the error `'utf-8' codec can't decode byte 0xe8 in position 6570: invalid continuation byte`.  Investigating that further, I managed to read those files with a different encoding (e.g. latin-1). I did not understand though why the files are not valid utf-8 in the first place. Adding that fs_args mentioned above solved the issues.

## Steps to Reproduce
1. Save some plotly plot as plotly.JSONDataset with special characters.
2. Try to load that plot via the catalog again.
3. Eventually there should occur the error mentioned above.
4. Change the fs_args as indicated and repeat steps 1 and 2. Now it should work without issues.

## Expected Result
There should not be any encoding issues happening, because it is expected that files are saved as utf-8.

## Actual Result
The file was not saved in utf-8.

```
utf-8' codec can't decode byte 0xe8 in position 6570: invalid continuation byte
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.19.5
* Kedro plugin and kedro plugin version used (`pip show kedro-airflow`): 3.0.0
* Python version used (`python -V`): 3.11.7
* Operating system and version: Windows 11

SRIKAR-8-77:
## Description
Fixes #741.

The plotly.JSONDataset lacks a file encoding on save, resulting in a UnicodeDecodeError when reloading charts containing special characters on non-UTF-8 systems.

This PR resolves the bug by modifying the DEFAULT_FS_ARGS class attribute on line 55 to add encoding="utf-8" by default for save operations. This makes the save action consistent with the current utf-8 decode on load to provide consistent reliability.

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -17,6 +17,7 @@
 
 ## Bug fixes and other changes
 - Updated `pandas.DeltaTableDataset` to be compatible with `deltalake` version 1.x.
+- Fix `plotly.JSONDataset` encoding errors by defaulting thesave encoding to UTF-8.
 
 ## Community contributions
 
--- a/kedro-datasets/kedro_datasets/plotly/json_dataset.py
+++ b/kedro-datasets/kedro_datasets/plotly/json_dataset.py
@@ -52,7 +52,9 @@ class JSONDataset(AbstractVersionedDataset[go.Figure, go.Figure | go.FigureWidge
 
     DEFAULT_LOAD_ARGS: dict[str, Any] = {}
     DEFAULT_SAVE_ARGS: dict[str, Any] = {}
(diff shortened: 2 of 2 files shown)


ravi-kumar-pilla (approved):
LGTM 👍

Issue #1203, opened 2025-10-02
`kedro-datasets`: Implement `LangfuseTraceDataset`


ElenaKhaustova:
## Description
Implement an experimental `LangfuseTraceDataset` that provides Langfuse clients or callbacks with credentials managed via Kedro config.

## Context
https://github.com/kedro-org/kedro/issues/5088#issuecomment-3328361496

PR #1208, opened 2025-10-07
feat(datasets): Added the Experimental LangfuseTraceDataset

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-plugins/issues/1203

This PR adds `LangfuseTraceDataset` as an experimental dataset to enable Langfuse integration for tracing in Kedro pipelines. The implementation for this dataset can be found here: https://github.com/kedro-org/kedro-academy/pull/77.

The dataset provides three modes of integration:

langchain: Returns `CallbackHandler` for LangChain tracing
openai: Returns wrapped OpenAI client with automatic tracing
sdk: Returns raw Langfuse client for manual instrumentation

## Development notes

- Added `LangfuseTraceDataset` to `kedro_datasets_experimental.langfuse` module
- The dataset manages Langfuse credentials and sets required environment variables during initialisation
- Implements read-only pattern as tracing clients are not data to be persisted
- Added documentation
- Updated RELEASE.md

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
@@ -25,7 +25,7 @@ class LangfuseTraceDataset(AbstractDataset):
 
         ```yaml
         langfuse_trace:
-          type: kedro_datasets_experimental.langfuse.LangfuseTraceDataset
+          type: langfuse.LangfuseTraceDataset
           credentials: langfuse_credentials
           mode: openai
         ```
--- a/kedro-datasets/pyproject.toml
+++ b/kedro-datasets/pyproject.toml
@@ -187,7 +187,7 @@ langchain-chatcoheredataset = ["langchain-cohere~=0.1.5", "langchain-community~=
 langchain-langchainpromptdataset = ["langchain>=0.2.0"]
 langchain = ["kedro-datasets[langchain-chatopenaidataset,langchain-openaiembeddingsdataset,langchain-chatanthropicdataset,langchain-chatcoheredataset]"]
 langfuse-langfusetracedataset = ["langfuse>=2.0.0"]
-langfuse = ["kedro-datasets[langfuse-langfusetracedataset]"]
+langfuse = ["kedro-datasets[langfuse-langfusetracedataset]", "openai>=2.3.0"]
 
 netcdf-netcdfdataset = ["h5netcdf>=1.2.0","netcdf4>=1.6.4","xarray>=2023.1.0"]
 netcdf = ["kedro-datasets[netcdf-netcdfdataset]"]


kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,260 @@
...
+                dataset = LangfuseTraceDataset(credentials=creds, mode="openai")
+                client = dataset.load()
+                response = client.chat.completions.create(model="gpt-4", messages=[...])
+
+            # SDK mode
+                dataset = LangfuseTraceDataset(credentials=creds, mode="sdk")
+                langfuse = dataset.load()
+                trace = langfuse.trace(name="my-trace")
+        """
+        if self._mode == "langchain":
```
ElenaKhaustova, line 243: I wonder if it makes sense to cache the loaded client to prevent extra API setups when load() is called multiple times.

```python
self._cached_client = None

def load(self):
    if self._cached_client is not None:
        return self._cached_client
    self._cached_client = self._create_client()
    return self._cached_client
```

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,260 @@
...
+            from langfuse.langchain import CallbackHandler  # noqa PLC0415
+            return CallbackHandler()
+        elif self._mode == "openai":
+            from langfuse.openai import OpenAI  # noqa PLC0415
+            client_params = self._build_openai_client_params()
+            return OpenAI(**client_params)
+        else:
+            return Langfuse()
```
ElenaKhaustova: Can we please double-check here if we should use this API or `get_client()`, which they use in their recent examples: https://langfuse.com/docs/observability/get-started

Or at least we need to understand what constraints they add to the langfuse version.

Whether we should have something like that:

```python
try:
    from langfuse import get_client
    return get_client()
except ImportError:
    return Langfuse()

```

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+    - langchain: Returns CallbackHandler for LangChain integration
+    - openai: Returns wrapped OpenAI client with automatic tracing
+    - sdk: Returns raw Langfuse client for manual tracing
+
+    Examples:
+        Using catalog YAML configuration:
+
+        ```yaml
+        langfuse_trace:
+          type: kedro_agentic_workflows.datasets.langfuse_trace_dataset.LangfuseTraceDataset
```
ravi-kumar-pilla:
```suggestion
          type: langfuse.LangfuseTraceDataset
```

ravi-kumar-pilla: It is a bit confusing to say type as langfuse.LangfuseTraceDataset. Though I see langchain following similar pattern. Since we have the main langfuse library from the actual Langfuse module, may be we should rename our folders as kedro_langfuse ? So users can reference it like type: kedro_langfuse.LangfuseTraceDataset. Not sure if this is some convention followed in kedro datasets. 

On first glance it seems like LangfuseTraceDataset is coming from the official langfuse application

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+          credentials: langfuse_credentials
+          mode: openai
+        ```
+
+        Using Python API:
+
+        ```python
+        from kedro_agentic_workflows.datasets.langfuse_trace_dataset import LangfuseTraceDataset
```
ravi-kumar-pilla: Need to update these docstrings accordingly through out this file.

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Returns:
+            Dictionary with validated OpenAI client parameters. Always includes
+            'api_key', optionally includes 'base_url' if provided.
+
+        Raises:
+            ValueError: If OpenAI credentials are missing or invalid.
+
+        Examples:
+            # With API key only
```
ravi-kumar-pilla, line 175: This might need some formatting ?

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        # Check if openai section exists
+        if "openai" not in self._credentials:
+            raise ValueError("OpenAI mode requires 'openai' section in credentials")
+
+        openai_creds = self._credentials["openai"]
+
+        # Check for required API key
+        if "openai_api_key" not in openai_creds:
+            raise ValueError("Missing required OpenAI credential: 'openai_api_key'")
```
ravi-kumar-pilla: I see we throw DatasetError instances in most of our core datasets and experimental for any issues related to missing values, issues with business logic etc. 

Apart from something like NotImplementedError or Network related errors, we try to only throw DatasetError instance. Should we stick to the convention ?

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
SajidAlamQB: wouldn't it actually be `kedro_datasets_experimental.langfuse.LangfuseTraceDataset`

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+    - langchain: Returns CallbackHandler for LangChain integration
+    - openai: Returns wrapped OpenAI client with automatic tracing
+    - sdk: Returns raw Langfuse client for manual tracing
+
+    Examples:
+        Using catalog YAML configuration:
+
+        ```yaml
+        langfuse_trace:
+          type: kedro_datasets_experimental.langfuse.LangfuseTraceDataset
```
ravi-kumar-pilla: I am a bit confused on the usage. If I see other experimental datasets, we mention the usage as module.ds without kedro_datasets_experimental i.e., instead of `kedro_datasets_experimental.langfuse.LangfuseTraceDataset` -> `langfuse.LangfuseTraceDataset`. 

If I test the local version (not published in pypi) `langfuse.LangfuseTraceDataset` does not work. I am not sure what should we suggest here or I am missing something.

kedro-datasets/pyproject.toml
```diff
@@ -186,6 +186,8 @@ langchain-chatanthropicdataset = ["langchain-anthropic~=0.1.13", "langchain-comm
 langchain-chatcoheredataset = ["langchain-cohere~=0.1.5", "langchain-community~=0.2.0"]
 langchain-langchainpromptdataset = ["langchain>=0.2.0"]
 langchain = ["kedro-datasets[langchain-chatopenaidataset,langchain-openaiembeddingsdataset,langchain-chatanthropicdataset,langchain-chatcoheredataset]"]
+langfuse-langfusetracedataset = ["langfuse>=2.0.0"]
+langfuse = ["kedro-datasets[langfuse-langfusetracedataset]"]
```
ravi-kumar-pilla:
```suggestion
langfuse = ["kedro-datasets[langfuse-langfusetracedataset]", "openai>=2.3.0"]
```

since openai is required for mode=openai. Without this the test at - https://github.com/kedro-org/kedro-plugins/pull/1208/files#diff-14da23f8d7e88ab9dc1c5873bd77295c52fbd1a7cd135cf84f1b59c83f372011R171
fails

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
SajidAlamQB: We should use `langfuse.LangfuseTraceDataset` as this will work when its published.

Issue #1179, opened 2025-09-08
`kedro-datasets`: Assess existing experimental `LangChain` datasets for graduation


ElenaKhaustova:
## Description
Kedro has a number of datasets that are currently considered experimental. One example is the family of `LangChain` datasets - https://github.com/kedro-org/kedro-plugins/tree/main/kedro-datasets/kedro_datasets_experimental/langchain, which were introduced experimentally.

These datasets have shown their usefulness in real-world applications and do not require significant additional maintenance.

We propose assessing the `LangChain` datasets with the goal of graduating them into the core `kedro_datasets` package.

## Context

- `LangChain` has become a widely adopted framework for LLMs and agentic workflows.
- In projects such as [kedro-rag-chatbot](https://github.com/kedro-org/kedro-academy/blob/main/kedro-rag-chatbot/conf/base/catalog.yml) `LangChain` datasets proved valuable for `LLM` usage within `Kedro` pipelines.
- Moving them to core datasets would give users more confidence in adopting them for production workflows.

ElenaKhaustova:
> Do not forget to add the support for the recently released `langchain>=1.0`

It was confirmed by @ankatiyar that langchain datasets are not affected by the 1.0 langchain release because they use the `langchain-openai`, `langchain-anthropic` and `langchain-cohere`

Issue #517, opened 2024-01-16
Proposal for Adding Contributions Space for Experimental Datasets


merelcht:
### Description
Introduce a `contrib` folder within the Kedro datasets repository to accommodate contributions that are more experimental and may not fully adhere to the usual standards, such as being fully tested. This space will allow for the inclusion of datasets that are in the early stages of development or might not meet the criteria for being part of the core Kedro datasets.

An example of such datasets are the `langchain` based datasets for which we have an open draft PR: https://github.com/kedro-org/kedro-plugins/pull/434

**Key Points:**
1. The `contrib` folder is designated for experimental contributions and should not be held to the same maintenance standards as the core Kedro datasets.
2. Contributions within the `contrib` folder are owned by their primary authors, and the core Kedro team is not responsible for their active maintenance.
3. Datasets within the `contrib` folder may evolve and improve over time. Successful and well-maintained contributions can graduate from the `contrib` folder and move to the regular `kedro_datasets` space.
4. Establish clear criteria and guidelines for determining when a contribution is considered a regular (non-experimental) contribution versus an experimental one. This will help contributors and maintainers understand the expectations and classification of datasets.

**Considerations:**
- Define specific criteria for moving a dataset from the `contrib` folder to the regular `kedro_datasets` space.
- Ensure that the `contrib` folder is clearly communicated as an experimental space, encouraging users to be cautious when relying on datasets from this folder.
- Document the rules and guidelines for contributing to both the `contrib` and regular `kedro_datasets` spaces in the project's documentation.

### Next steps
- Seek feedback from the Kedro TSC on the proposed structure and guidelines.
- Finalise and document the criteria for contributions to the `contrib` and regular `kedro_datasets` spaces.
- Implement the `contrib` folder and update the documentation accordingly.

**Note:**
This issue serves as a proposal and discussion point. Further details and decisions will be made in collaboration with the Kedro community and maintainers.

### Examples
Projects that have a similar contribution space:

- https://github.com/great-expectations/great_expectations/tree/develop/contrib
- https://github.com/git/git/tree/master/contrib

noklam:
Add `tensorflow` as an exampler https://www.tensorflow.org/guide/migrate/migrate_tf2#:~:text=Remove%20old%20tf.contrib%20symbols%20(check%20TF%20Addons%20and%20TF%2DSlim).

[...]

PR #1249, opened 2025-11-13
refactor(datasets)!: graduate LangChain chat datasets

ankatiyar:
## Description

#1179 
Would like TSC votes for the proposal to graduate the langchain chat datasets - `ChatOpenAIDataset` `ChatAnthropicDataset` `ChatCohereDataset` & `OpenAIEmbeddingsDataset`

## Development notes

- ~Rename `OpenAIDataset` -> `AbstractOpenAIDataset` and moved to `_utils` (Open to suggestions on where this goes) This was done in discussion with @ElenaKhaustova as this abstract class will be useful for implementation of an OpenAI dataset without `langchain`~
- Separate `OpenAIEmbeddingsDataset` and `ChatOpenAIDataset` into separate files
- Rename the file names from `_openai`, `_anthropic` and `_cohere` 
- Move datasets to core
- Move tests to core tests
- Move listing in docs to core
- Add to release notes

TODO (Separate PR):
- Add test for abstract class 
- Update tests to use fixtures 

## Checklist

### DIFF
--- a/kedro-datasets/tests/langchain/test_openai_dataset.py
+++ b/kedro-datasets/tests/langchain/test_openai_dataset.py
@@ -31,115 +31,58 @@ def embeddings_kwargs():
     return {"model": "text-embedding-ada-002"}
 
 
-@pytest.fixture
-def test_openai_dataset():
-    """Fixture that creates a concrete TestDataset for base class testing."""
+class TestOpenAIEmbeddingsDataset:
+    """Test the OpenAIEmbeddingsDataset class."""
 
-    class TestDataset(AbstractOpenAIDataset):
-        @property
-        def constructor(self):
-            return Mock
-
-    return TestDataset
-
-
-class TestOpenAIDataset:
-    """Test the base OpenAIDataset class."""
-
-    def test_init_with_credentials(
-        self, test_openai_dataset, openai_credentials, chat_kwargs
-    ):
+    def test_init_with_credentials(self, openai_credentials, chat_kwargs):
         """Test dataset initialization with credentials."""
-        dataset = test_openai_dataset(
-            credentials=openai_credentials, kwargs=chat_kwargs
+        dataset = OpenAIEmbeddingsDataset(
+            credentials=openai_credentials, kwargs=embeddings_kwargs
         )
 
         assert dataset.credentials == openai_credentials
-        assert dataset.kwargs == chat_kwargs
+        assert dataset.kwargs == embeddings_kwargs
 
-    def test_init_without_credentials(self, test_openai_dataset):
+    def test_init_without_credentials(self):
         """Test dataset initialization without credentials (uses environment variables)."""
-        dataset = test_openai_dataset()
+        dataset = OpenAIEmbeddingsDataset()
 
         assert dataset.credentials == {}
         assert dataset.kwargs == {}
 
-    def test_init_with_partial_credentials(self, test_openai_dataset):
+    def test_init_with_partial_credentials(self):
         """Test dataset initialization with partial credentials works."""
-        credentials = {"openai_api_key": "sk-test-key"}  # pragma: allowlist-secret
-        dataset = test_openai_dataset(credentials=credentials)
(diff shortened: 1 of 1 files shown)


kedro-datasets/pyproject.toml
```diff
@@ -78,6 +78,12 @@ ibis = ["ibis-framework"]
 json-jsondataset = []
 json = ["kedro-datasets[json-jsondataset]"]
 
+langchain-chatopenaidataset = ["langchain-openai~=0.1.7"]
```
ElenaKhaustova: Since we tested it with langhcain 1.0 we may replace `~=` with `>=` here and below

rashidakanchwala (approved):

kedro-datasets/pyproject.toml
ElenaKhaustova: I think we need to replace for all the groups as well `langchain-*`

ElenaKhaustova (approved):

merelcht:
> Why couple to langchain and not make a generic approach?

This is just graduating the existing experimental datasets that are `langchain` based. We can implement generic ones if there's a user need as a separate task.

kedro-datasets/kedro_datasets/_utils/abstract_openai_dataset.py
```diff
@@ -0,0 +1,61 @@
...
+from abc import abstractmethod
+from typing import Any, Generic, NoReturn, TypeVar
+
+from kedro.io import AbstractDataset, DatasetError
+from langchain_openai import ChatOpenAI, OpenAIEmbeddings
+
+OPENAI_TYPE = TypeVar("OPENAI_TYPE")
+
+
+class AbstractOpenAIDataset(AbstractDataset[None, OPENAI_TYPE], Generic[OPENAI_TYPE]):
```
deepyaman: In terms of reviewing the content of the datasets, still need to look deeply but one point that comes to mind:

Historically, we explicitly haven't supported any abstract dataset hierarchies. For example, while the `pandas.*Dataset` implementations share a great deal, they are all very explicit in their implementation. Part of the stated reason for this is that it makes it easy for contributors to understand, fork, and extend datasets. For example, let's say somebody wants to customize `ChatOpenAIDataset` in their project, they just copy-paste the dataset code and modify it; everything they should touch is there.

Given how little this `AbstractOpenAIDataset` does, what's the significant benefit of it being a separate abstract class? If anything, it seems like it could be a mix-in, but I don't think that's even necessary in this case.

merelcht:
Since I'm the person who introduced the concept of the **experimental datasets**, I just want to clarify and remind everyone of the purpose: "Experimental datasets are community-contributed and more flexible. They are not officially maintained and can be in early stages or less thoroughly tested.". Whereas: "Core datasets are officially maintained by the Kedro Technical Steering Committee (TSC) and must meet high standards, including full documentation, 100% test coverage, cross-platform support, and compatibility with supported Python versions." 

The term **experimental** in this context has nothing to do with usage, adoption, user need and so on. 

According to these rules (_our rules_), these LangChain datasets fit in the **core** bucket as we've added tests and have actively maintained them ever since they were contributed by an external contributor. 



(See the original discussions about introducing this: https://github.com/kedro-org/kedro-plugins/issues/517)

kedro-datasets/kedro_datasets/_utils/abstract_openai_dataset.py
ankatiyar: I've removed the abstract class

kedro-datasets/tests/langchain/test_openai_dataset.py
```diff
@@ -4,21 +4,19 @@
 
 import pytest
 from kedro.io import DatasetError
+from kedro_datasets._utils.abstract_openai_dataset import AbstractOpenAIDataset
```
deepyaman:
```suggestion
```

ankatiyar:
Thanks everyone for the discussion! There's 11 votes now, I'll merge it. From next time, we'll do votes before the PR is opened like @deepyaman suggested 😄

PR #1256, opened 2025-11-21
feat(telemetry): Collect dataset type information with telemetry

ankatiyar:
## Description

Partial solve https://github.com/kedro-org/kedro/issues/5177

## Development notes

- Collect dataset type info using `get_type()` functionality of the new catalog
This only works for the new catalog, and is therefore under the condition for the new catalog. The `get_type()` function doesn't materialise datasets so this won't add much overhead for telemetry. I didn't add it to the old catalog because we would have to materialise datasets which may cause problems for remote datasets 
- Update tests
- starters in a separate PR
## Checklist

### DIFF
--- a/kedro-telemetry/kedro_telemetry/plugin.py
+++ b/kedro-telemetry/kedro_telemetry/plugin.py
@@ -312,7 +312,7 @@ def _format_project_statistics_data(
         dataset_names = catalog.keys()
         for ds_name in dataset_names:
             if not ds_name.startswith(("parameters", "params:")):
-                ds_type = catalog.get_type(ds_name)
+                ds_type = catalog.get_type(ds_name) or ""
                 if (
                     ds_type.startswith("kedro_datasets.")
                     or ds_type.startswith("kedro.io.")
--- a/kedro-telemetry/tests/integration/dummy-project/pyproject.toml
+++ b/kedro-telemetry/tests/integration/dummy-project/pyproject.toml
@@ -13,7 +13,7 @@ dummy-project = "dummy_project.__main__:main"
 [tool.kedro]
 package_name = "dummy_project"
 project_name = "dummy_project"
-kedro_init_version = "1.0.0rc1"
+kedro_init_version = "1.1.0"
 tools = [ "None",]
 example_pipeline = "True"
 source_dir = "src"


kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -306,22 +306,26 @@ def _format_project_statistics_data(
 ):
     """Add project statistics to send to Heap."""
     # Support both catalog.list() for `kedro < 1.0` and catalog.keys() for `kedro >= 1.0`
+    dataset_types: dict[str | None, int] = {}
     if hasattr(catalog, "keys") and callable(catalog.keys):
         dataset_names = catalog.keys()
+        for ds_name in dataset_names:
+            if not ds_name.startswith(("parameters", "params:")):
+                ds_type = catalog.get_type(ds_name)
```
ElenaKhaustova: Shall we add a comment here that storing dataset type only works for kedro>=1.0?

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -306,22 +306,26 @@ def _format_project_statistics_data(
...
+                dataset_types[ds_type] = dataset_types.get(ds_type, 0) + 1
```
ElenaKhaustova: I'm not sure if we can use the full type here as a key (`kedro.io.memory_dataset.MemoryDataset`), because custom dataset types can include information about the user's project directories. I'm doubting if we want to store that as well.

So maybe we can use the full name in case the dataset comes from `kedro.io`, and otherwise add a separate category like `custom dataset`, so unnecessary information is not sent to the heap.

ankatiyar: done!

kedro-telemetry/kedro_telemetry/plugin.py
ankatiyar: done!

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -308,11 +308,15 @@ def _format_project_statistics_data(
     # Support both catalog.list() for `kedro < 1.0` and catalog.keys() for `kedro >= 1.0`
     dataset_types: dict[str | None, int] = {}
     if hasattr(catalog, "keys") and callable(catalog.keys):
+        # Only collect dataset types for kedro >= 1.0 because `get_type` method is not available in earlier versions
         dataset_names = catalog.keys()
         for ds_name in dataset_names:
             if not ds_name.startswith(("parameters", "params:")):
                 ds_type = catalog.get_type(ds_name)
-                dataset_types[ds_type] = dataset_types.get(ds_type, 0) + 1
+                if ds_type.startswith("kedro_datasets.") or ds_type.startswith("kedro.io."):
```
ElenaKhaustova: And `kedro_datasets_experimental.`

ankatiyar: whoops yes

kedro-telemetry/kedro_telemetry/plugin.py
```diff
@@ -306,22 +306,34 @@ def _format_project_statistics_data(
...
+            if not ds_name.startswith(("parameters", "params:")):
+                ds_type = catalog.get_type(ds_name)
+                if (
+                    ds_type.startswith("kedro_datasets.")
+                    or ds_type.startswith("kedro.io.")
+                    or ds_type.startswith("kedro_datasets_experimental.")
+                ):
+                    dataset_types[ds_type] = dataset_types.get(ds_type, 0) + 1
+                else:
+                    dataset_types["custom"] = dataset_types.get("custom", 0) + 1
```
merelcht, line 323: Should we add a test to check the custom tracking is done properly?

ankatiyar: Added

PR #1280, opened 2026-01-08
fix(datasets): Add mlrun datasets

rokatyy:
## Description

This PR adds experimental MLRun dataset integrations for Kedro, enabling seamless tracking of ML artifacts, models, dataframes, and results through [MLRun](https://www.mlrun.org/) - an open-source MLOps orchestration framework.

**Why?**
- MLRun provides robust experiment tracking, model versioning, and artifact management
- This integration allows Kedro users to leverage MLRun's capabilities without changing their pipeline code
- Artifacts are automatically logged to MLRun projects with proper metadata and lineage tracking

**New datasets:**
| Dataset | Purpose |
|---------|---------|
| `MLRunModel` | Save/load ML models with framework metadata and configurable format |
| `MLRunDataframeDataset` | Save/load pandas DataFrames as MLRun artifacts |
| `MLRunResult` | Log scalar results/metrics with optional dict flattening |
| `MLRunAbstractDataset` | Base class for custom MLRun dataset implementations |

**Architecture:**
- `MLRunContextManager` - Singleton that manages MLRun project and context lifecycle
- All datasets share the same context, ensuring consistent artifact tracking within a pipeline run

### Example catalog configuration
```yaml
trained_model:
  type: kedro_datasets_experimental.mlrun.MLRunModel
  key: my_model
  framework: sklearn
  model_format: pkl

training_metrics:
  type: kedro_datasets_experimental.mlrun.MLRunResult
  key: metrics
  flatten: true

processed_data:
  type: kedro_datasets_experimental.mlrun.MLRunDataframeDataset
  key: processed_df
```

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -6,6 +6,7 @@
 - Added `autogen` mode to `LangfuseTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
 - `api.APIDataset` now stores the response received from a `PUT` or `POST` request via the `response_dataset` parameter.
 - Added `autogen` mode to `OpikTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
+- Experimental datasets `OpikTraceDataset` and `LangfuseTraceDataset` now receive `openai` credentials as `base_url` and `api_key`, instead of `openai_api_base` and `openai_api_key`.
 
 ## Community contributions
 
--- a/kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
@@ -44,7 +44,7 @@ class LangfuseTraceDataset(AbstractDataset):
             credentials={
                 "public_key": "pk_...",
                 "secret_key": "sk_...",  # pragma: allowlist secret
-                "openai": {"openai_api_key": "sk-..."},  # pragma: allowlist secret
+                "openai": {"api_key": "sk-..."},  # pragma: allowlist secret
             },
             mode="openai",
         )
@@ -55,7 +55,7 @@ class LangfuseTraceDataset(AbstractDataset):
                 "public_key": "pk_...",
                 "secret_key": "sk_...",  # pragma: allowlist secret
                 "host": "https://custom.langfuse.com",
-                "openai": {"openai_api_key": "sk-..."},  # pragma: allowlist secret
+                "openai": {"api_key": "sk-..."},  # pragma: allowlist secret
             },
             mode="openai",
         )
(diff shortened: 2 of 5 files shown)


kedro-datasets/kedro_datasets_experimental/mlrun/df_dataset.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
+
+from . import MLRunAbstractDataset
+
+
+class MLRunDataframeDataset(MLRunAbstractDataset):
```
merelcht, line 7: Could you add a short docstring here to explain what the dataset is for?

kedro-datasets/kedro_datasets_experimental/mlrun/abstract_artifact.py
```diff
@@ -0,0 +1,77 @@
...
+
+    DEFAULT_LOAD_ARGS: dict[str, Any] = {}
+    DEFAULT_SAVE_ARGS: dict[str, Any] = {}
+
+    def __init__(self,
+                 key: str | None = None,
+                 load_args: dict[str, Any] | None = None,
+                 save_args: dict[str, Any] | None = None, ) -> None:
+        self._ctx_manager = MLRunContextManager()
+        self._key = key or getattr(self, '_name', 'unnamed')
```
SajidAlamQB: When this runs, `_name` hasn't been set yet so this will always fall back to `unnamed` when key is not given. You'd need a different approach for example, doing it in `save()`/`load()` or requiring keys explicitly.

rokatyy: @SajidAlamQB good point! Changed it, another look?

SajidAlamQB: Thanks @rokatyy the property approach is much better! 👍

rokatyy:
@merelcht @SajidAlamQB I see that CI fails on smth unrelated, could you please take a look? 🙏

kedro-datasets/kedro_datasets_experimental/mlrun/result.py
```diff
@@ -0,0 +1,75 @@
...
+    def save(self, data: Any) -> None:
+        if self._flatten and isinstance(data, dict):
+            flat = self._flatten_dict(data)
+            for k, v in flat.items():
+                self._ctx_manager.context.log_result(k, v)
+        else:
+            self._ctx_manager.context.log_result(self.key, data)
+
+    def load(self) -> Any:
+        return self._ctx_manager.context.results.get(self.key)
```
SajidAlamQB, line 80: The save and load don't match up, looks like the save splits the dict into individual keys and discards the original key, but load looks for the original key.

rokatyy: @SajidAlamQB there is no way to restore the keys from data during loading. So, if user wants to load them, they should use a different approach. This is expected and by design, flatten options is optional

SajidAlamQB: Makes sense then just a small suggestions maybe we can add a comment or note there that load() won't return the flattened keys.

rokatyy: added

Issue #1277, opened 2026-01-06
`kedro-datasets`: Extend `OpikTraceDataset` to support `AutoGen` tracing


ElenaKhaustova:
## Description
We recently introduced `OpikTraceDataset`, a Kedro dataset that provides tracing integrations for different AI frameworks via a configurable mode:
  - `sdk`: Returns a simple namespace-like client exposing the `track` decorator for manual tracing.
  - `openai`: Returns an OpenAI client automatically wrapped for Opik tracing.
  - `langchain`: Returns an `OpikTracer` callback handler for LangChain integration.

We now want to extend this dataset to support `AutoGen` so that users can easily enable `Opik` tracing when working with `AutoGen` agents and conversations.

## Context
**Dataset Behavior**
- Configure Opik environment variables during initialization (consistent with existing modes)
- Return an object that can be directly plugged into AutoGen configuration
- Follow the same lifecycle semantics as other modes (no side effects on load beyond setup)
- Supports both single-agent and multi-agent tracing out of the box

**Current implementation:** https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py

**Changes can be tested with this project:** https://github.com/kedro-org/kedro-academy/pull/9

PR #1288, opened 2026-01-20
feat(datasets): Extend `LangfuseTraceDataset` to support AutoGen tracing

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-plugins/issues/1276

To test for QA use the kedro-academy example: https://github.com/kedro-org/kedro-academy/pull/104

Adds `autogen` mode to `LangfuseTraceDataset`, enabling OpenTelemetry based tracing for AutoGen agent pipelines via Langfuse's OTLP endpoint.

## Development notes

- Added `autogen` mode to `LangfuseTraceDataset` that returns a configured OpenTelemetry Tracer
- `_build_autogen_tracer()` sets up an OTLP exporter 

## Checklist

- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [ ] Updated `jsonschema/kedro-catalog-X.XX.json` if necessary
- [x] Added a description of this change in the relevant `RELEASE.md` file
- [x] Added tests to cover my changes
- [ ] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)

PR #1295, opened 2026-01-30
feat(datasets): Add autogen support for `OpiktraceDataset`

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-plugins/issues/1277

To test for QA use the kedro-academy example: https://github.com/kedro-org/kedro-academy/pull/105

Adds `autogen` mode to `OpikTraceDataset`, enabling OpenTelemetry based tracing for AutoGen agent pipelines via Opik's OTLP endpoint.

## Development notes

Added `autogen` mode to `OpikTraceDataset` that returns a configured OpenTelemetry Tracer
`_build_autogen_tracer()` sets up an OTLP exporter with Opik authentication headers.
`AutoGen` mode skips Opik SDK `configure()` uses OTLP directly.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
@@ -24,7 +24,7 @@ class OpikTraceDataset(AbstractDataset):
     - `sdk`: Returns a simple namespace-like client exposing the `track` decorator for manual tracing.
     - `openai`: Returns an OpenAI client automatically wrapped for Opik tracing.
     - `langchain`: Returns an `OpikTracer` callback handler for LangChain integration.
-    - `autogen`: Returns a configured ``Tracer`` for AutoGen integration via OTLP.
+    - `autogen`: Returns a configured `Tracer` for AutoGen integration via OTLP (OpenTelemetry Protocol).
 
     **Examples**
 


kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
```diff
@@ -178,6 +203,62 @@ def _build_openai_client_params(self) -> dict[str, str]:
...
+            "Comet-Workspace": self._credentials["workspace"],
+        }
+
+        # Add project name if specified
+        project_name = self._credentials.get("project_name")
+        if project_name:
+            headers["projectName"] = project_name
+
+        # Use Opik's OTLP endpoint (or custom url_override)
+        base_url = self._credentials.get("url_override", "https://www.comet.com")
```
ElenaKhaustova: I think here we have exactly the same issue as with the Langfuse: endpoint should be configurable

https://github.com/kedro-org/kedro-plugins/pull/1288#discussion_r2755610069

SajidAlamQB: Done.

kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
```diff
@@ -178,6 +209,66 @@ def _build_openai_client_params(self) -> dict[str, str]:
...
+        }
+
+        # Add project name if specified
+        project_name = self._credentials.get("project_name")
+        if project_name:
+            headers["projectName"] = project_name
+
+        # Use Opik's OTLP endpoint (url_override validated in _validate_opik_credentials)
+        base_url = self._credentials["url_override"]
+        endpoint = f"{base_url}/opik/api/v1/private/otel/v1/traces"
```
ElenaKhaustova: Here we have exactly the same issue with the hardcoded `endpoint` for `LangfuseTraceDataset`: https://github.com/kedro-org/kedro-plugins/pull/1288#discussion_r2764163781

One of the ways to do it is to create another constant, for example, REQUIRED_OPIK_CREDENTIALS_OTLP and do it in the `_validate_langfuse_credentials`, expecting that the endpoint is provided by the user as `{host}/api/public/otel/v1/traces` or the alternative is to make `host` required, plus the rest of the endpoint string.

ElenaKhaustova: And the same points on unit test - could you please add them too, depending on your solution?

We should add them to ensure some core cases work as expected, even though it's an experimental dataset.

SajidAlamQB: Done 👍

kedro-datasets/pyproject.toml
```diff
@@ -218,7 +218,8 @@ langfuse-langfusetracedataset = ["langfuse>=2.0.0"]
 langfuse = ["kedro-datasets[langfuse-langfusepromptdataset,langfuse-langfusetracedataset]", "openai>=2.3.0", "langchain>=0.2.0, <1.0"]
 opik-opikpromptdataset = ["opik>=1.8.0"]
 opik-opiktracedataset = ["opik>=1.8.0"]
-opik = ["kedro-datasets[opik-opikpromptdataset, opik-opiktracedataset]", "openai>=2.3.0", "langchain>=0.2.0"]
+opik-opiktracedataset-autogen = ["opik", "opentelemetry-sdk", "opentelemetry-exporter-otlp-proto-http"]
```
ElenaKhaustova: Shall we pin opik version here as well?

ElenaKhaustova (commented):
Apart from the nit comment it looks good to me, thanks!

Tested with https://github.com/kedro-org/kedro-academy/pull/105 and `python3.11`, `python3.12`

kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
```diff
@@ -23,6 +24,7 @@ class OpikTraceDataset(AbstractDataset):
     - `sdk`: Returns a simple namespace-like client exposing the `track` decorator for manual tracing.
     - `openai`: Returns an OpenAI client automatically wrapped for Opik tracing.
     - `langchain`: Returns an `OpikTracer` callback handler for LangChain integration.
+    - `autogen`: Returns a configured ``Tracer`` for AutoGen integration via OTLP.
```
merelcht: Might be helpful here to expand what OTLP stands for since it's referred to a lot in the subsequent code.

kedro-datasets/pyproject.toml
```diff
@@ -218,7 +218,8 @@ langfuse-langfusetracedataset = ["langfuse>=2.0.0"]
 langfuse = ["kedro-datasets[langfuse-langfusepromptdataset,langfuse-langfusetracedataset]", "openai>=2.3.0", "langchain>=0.2.0, <1.0"]
 opik-opikpromptdataset = ["opik>=1.8.0"]
 opik-opiktracedataset = ["opik>=1.8.0"]
-opik = ["kedro-datasets[opik-opikpromptdataset, opik-opiktracedataset]", "openai>=2.3.0", "langchain>=0.2.0"]
+opik-opiktracedataset-autogen = ["opik>=1.8.0", "opentelemetry-sdk", "opentelemetry-exporter-otlp-proto-http"]
+opik = ["kedro-datasets[opik-opikpromptdataset, opik-opiktracedataset, opik-opiktracedataset-autogen]", "openai>=2.3.0", "langchain>=0.2.0"]
```
merelcht, line 222: Not specifically about the changes you made, but I'm just wondering if the dependency splits here for opik and langfuse make sense. `opikpromptdataset` and `opiktracedataset` and `opiktracedataset-autogen` all specify `opik>=1.8.0` and then the generic `opik` one assembles them all and adds on openai and langchain as dependencies. 

If a user only installs `opik-opikpromptdataset` do they actually get all required dependencies? And is it necessary to assemble them all into the generic `opik` one, can't we instead do something like:

```
opik-opiktracedataset-autogen = ["opik>=1.8.0", "opentelemetry-sdk", "opentelemetry-exporter-otlp-proto-http"]
opik = ["kedro-datasets[opik-opiktracedataset-autogen]", "opik>=1.8.0", "openai>=2.3.0", "langchain>=0.2.0"]
```

SajidAlamQB: The split follows the same pattern used in kedro-datasets for langfuse, langchain etc. Each dataset gets its own extra so users can install only what they need. The generic opik extra is a convenience group that pulls everything together for users who want the full Opik integration. 

The individual extras like `opik-opikpromptdataset` does include all deps for that specific dataset.

PR #1297, opened 2026-02-03
fix(datasets): Remove `cachetools` deps 

rashidakanchwala:
## Description

This PR removes the `cachetools` dependency from `kedro-datasets`, where it was used in `PartitionedDataset` and `IncrementalDataset`. 

Breaking changes introduced in cachetools 7 (including stricter @cachedmethod behavior) caused failures when running pipelines with ParallelRunner in Kedro core, which motivated the removal of this dependency there and prompted the same change here (see [#5352](https://github.com/kedro-org/kedro/pull/5352)).

Since these datasets only cache a small amount of per-instance state—such as the list of discovered partitions and the incremental tracking state—and do not require features like eviction policies, TTLs, or multi-key caching, the use of cachetools is unnecessary. This change replaces cachetools-based caching with lightweight manual caching, simplifying the implementation.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -3,6 +3,7 @@
 ## Bug fixes and other changes
 
 - Fixed `ibis.TableDataset` `exists` method to account for `database` (i.e. the collection of tables, or schema).
+- Added `autogen` mode to `LangfuseTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
 - `api.APIDataset` now stores the response received from a `PUT` or `POST` request via the `response_dataset` parameter.
 - Added `autogen` mode to `OpikTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
 
--- a/kedro-datasets/docs/javascript/deindex-old-docs.js
+++ b/kedro-datasets/docs/javascript/deindex-old-docs.js
@@ -1,0 +1,35 @@
+if (document.querySelector("meta[name='readthedocs-addons-api-version']") === null) {
+    const meta = document.createElement("meta");
+    meta.name = "readthedocs-addons-api-version";
+    meta.content = "1";
+    document.head.appendChild(meta);
+}
+
+function ensureNoIndexMeta() {
+    if (!document.querySelector('meta[name="robots"]')) {
+        const meta = document.createElement("meta");
+        meta.name = "robots";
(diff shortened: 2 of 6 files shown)


kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
```diff
@@ -204,7 +202,7 @@ def __init__(  # noqa: PLR0913
         self._filename_suffix = filename_suffix
         self._overwrite = overwrite
         self._protocol = infer_storage_options(self._path)["protocol"]
-        self._partition_cache: Cache = Cache(maxsize=1)
```
merelcht, line 207: Do we need to preserve this `maxsize` functionality?

kedro-datasets/tests/partitions/test_partitioned_dataset.py
```diff
@@ -152,8 +152,8 @@ def test_save_invalidates_cache(self, local_csvs, mocker):
 
         first_load = pds.load()
         assert (
-            pds._partition_cache.currsize == 1
-        )  # _list_partitions caches its result in _partition_cache
+            pds._cached_partitions is not None
```
merelcht, line 155: Is `currsize` something we need to keep? It seems like now we wouldn't be able to check the cache size?

kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py
rashidakanchwala: maxsize here is just the number of cached entries. Since we only cache one entry in this code path, there’s no real maxsize behaviour to preserve. Cachetools is overkill here, so replacing it with pds._cached_partitions is equivalent.

In the other kedro PR, we cache two distinct values (save and load), which is why that change introduces two cached variables instead.

kedro-datasets/tests/partitions/test_partitioned_dataset.py
rashidakanchwala: With maxsize=1, currsize could only ever be 0 or 1, effectively answering the question “has this been cached or not?”. Now that we’ve removed cachetools, we assert the same thing directly, so currsize is no longer required.

ElenaKhaustova (approved):
Thank you, @rashidakanchwala, LGTM!

One minor suggestion is to manually test these datasets with the changes we've made on the kedro side in a few scenarios to double-check that everything works as expected.

PR #1305, opened 2026-02-11
refactor(datasets): Change openai credentials on langfuse/opik datasets

lrcouto:
## Description

This PR changes the OpenAI credential format in `LangfuseTraceDataset` and `OpikTraceDataset` to use :

```
openai:
    base_url: <openai-api-base>  # Optional, defaults to OpenAI default
    api_key: <openai-api-key>   # Optional if OPENAI_API_KEY is set
```

Instead of the way it was previously:

```
openai:
    openai_api_base: <openai-api-base>
    openai_api_key: <openai-api-key>
```

This aligns with the format used by ChatOpenAIDataset and OpenAI's own conventions, making it possible to reuse the same credential structure across all datasets that accept OpenAI credentials.

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -6,6 +6,7 @@
 - Added `autogen` mode to `LangfuseTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
 - `api.APIDataset` now stores the response received from a `PUT` or `POST` request via the `response_dataset` parameter.
 - Added `autogen` mode to `OpikTraceDataset` for tracing AutoGen agent conversations with OpenTelemetry integration.
+- Experimental datasets `OpikTraceDataset` and `LangfuseTraceDataset` now receive `openai` credentials as `base_url` and `api_key`, instead of `openai_api_base` and `openai_api_key`.
 
 ## Community contributions
 


kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -248,19 +248,19 @@ def _build_openai_client_params(self) -> dict[str, str]:
...
+            raise DatasetError("Missing required OpenAI credential: 'api_key'")
 
         # Validate that API key is not empty
-        if not openai_creds["openai_api_key"] or not openai_creds["openai_api_key"].strip():
+        if not openai_creds["api_key"] or not openai_creds["api_key"].strip():
             raise DatasetError("OpenAI API key cannot be empty")
 
         # Build validated client parameters
-        client_params = {"api_key": openai_creds["openai_api_key"]}
+        client_params = {"api_key": openai_creds["api_key"]}
```
ElenaKhaustova: Since we're no longer modifying the default `openai` credentials, we don't need a separate object. And the function itself should only validate, not build, so it doesn't need to return anything. And we need to rename it as well `_build_openai_client_params` -> `_validate_openai_client_params`

kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_trace_dataset.py
```diff
@@ -248,19 +248,19 @@ def _build_openai_client_params(self) -> dict[str, str]:
...
 
         # Add base_url if provided (optional)
-        if "openai_api_base" in openai_creds and openai_creds["openai_api_base"]:
-            client_params["base_url"] = openai_creds["openai_api_base"]
+        if "base_url" in openai_creds and openai_creds["base_url"]:
```
ElenaKhaustova: And we don't need to add base url as well. We can optionally validate it's not empty if it's provided.

kedro-datasets/kedro_datasets_experimental/opik/opik_trace_dataset.py
```diff
@@ -214,13 +214,13 @@ def _build_openai_client_params(self) -> dict[str, str]:
 
         openai_creds = self._credentials["openai"]
 
-        api_key = str(openai_creds.get("openai_api_key", "")).strip()
+        api_key = str(openai_creds.get("api_key", "")).strip()
```
ElenaKhaustova, line 221: Same ideas as with the langfuse one. We don't need to build credentials anymore, as keys are default for `openai`. Only validate.

ElenaKhaustova (changes requested):

Issue #1162, opened 2025-08-14
Write mode `upsert` not working in Databricks ManagedTableDataset


ravi-kumar-pilla:
## Description

While working on #1140 I tried using `upsert` as write_mode. I am not sure if it is related to permissions but `catalog.save()` throws the below error pointing to - https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py#L482

"[CONFIG_NOT_AVAILABLE] Configuration fullTableAddress is not available. SQLSTATE: 42K0I"

<img width="2328" height="180" alt="Image" src="https://github.com/user-attachments/assets/05e19930-7023-4fe9-94e8-9d613f38ff38" />

## Context

https://github.com/kedro-org/kedro-plugins/issues/1140#issuecomment-3112006382

## Steps to Reproduce

If you are familiar with Databricks, try creating a [ManagedTableDataset](https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-8.0.0/api/kedro_datasets/databricks.ManagedTableDataset/) in upsert mode and save it twice. 

If you are not familiar with Databricks, try below steps:

- Try [databricks free edition ](https://docs.databricks.com/aws/en/getting-started/free-edition) in case you do not have Platform McK Azure Analytics Service
- Follow the [databricks setup](https://github.com/kedro-org/kedro-academy/tree/main/kedro-databricks-bootcamp#3-databricks-setup)
- Follow the notebook for initial setup - https://github.com/kedro-org/kedro-academy/blob/main/kedro-databricks-bootcamp/02_databricks/02_0-Unity-Catalog-checks.ipynb
- Follow the notebook to create connection with Unity Catalog completing Exercise 1 - https://github.com/kedro-org/kedro-academy/blob/main/kedro-databricks-bootcamp/02_databricks/rocketfuel/notebooks/02_1-Kedro-on-Databricks-part-1.ipynb . Though you need to modify the companies catalog entry. 

Make sure your conf at `kedro-databricks-bootcamp/02_databricks/rocketfuel/conf/databricks/catalog.yml` is as below. Note that the companies entry contains a primary_key prop.

```yaml
_uc_catalog: <your catalog name>
_uc_schema: <your schema/database name>

companies_raw:
  type: spark.SparkDataset
  filepath: /Volumes/${_uc_catalog}/${_uc_schema}/bronze/companies.csv
  file_format: csv
  load_args:
[...]

PR #1323, opened 2026-02-26
fix(datasets): Fix ManagedTableDataset [CONFIG_NOT_AVAILABLE] error on Databricks

lrcouto:
## Description

Fix for https://github.com/kedro-org/kedro-plugins/issues/1162

### Why the bug happened

The `_save_upsert` method in `_base_table_dataset.py` used `spark.conf.set()` to store the table address and join condition as runtime config values, then referenced them via ${...} variable substitution in the MERGE SQL statement, as previously pointed out in [this comment](https://github.com/kedro-org/kedro-plugins/issues/1162#issuecomment-3385169576):

```py
get_spark().conf.set("fullTableAddress", self._table.full_table_location())
get_spark().conf.set("whereExpr", where_expr)
upsert_sql = """MERGE INTO ${fullTableAddress} base USING update ON ${whereExpr}
    WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *"""
get_spark().sql(upsert_sql)
```

This approach breaks on Databricks Spark Connect, which is the connection protocol used in Databricks Serverless compute and DBR 14+ runtimes. This happens because Spark Connect enforces a strict allowlist of configuration keys via `SparkConnectConfig.assertConfigAllowed()`. Attempting to set an arbitrary key like `fullTableAddress` raises:

`[CONFIG_NOT_AVAILABLE] Configuration fullTableAddress is not available.`

### How it was fixed

The `spark.conf.set()` calls and the ${...} SQL template were replaced with direct f-string interpolation:

```py
update_data.createOrReplaceTempView("update")
full_table = self._table.full_table_location()
upsert_sql = f"""MERGE INTO {full_table} base USING update ON {where_expr}
    WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *"""  # nosec B608
get_spark().sql(upsert_sql)
```

This eliminates the dependency on `spark.conf.set()`, and works both locally and on Databricks Spark Connect.

### Why # nosec B608 was used
Bandit's [B608](https://bandit.readthedocs.io/en/latest/plugins/b608_hardcoded_sql_expressions.html) rule flags any SQL string built with string interpolation as a potential SQL injection vector (CWE-89). In this case, the warning is a false positive.

`full_table` is derived from `self._table.full_table_location()`, which is composed of the catalog, database, and table fields. These are set by the pipeline developer in the Kedro catalog configuration YAML, so they are not runtime user input.
The value does not originate from an untrusted source at runtime. They are equivalent to a database connection string or table name in an ORM configuration.

If we want an alternative approach that doesn't require this suppression, what we can do is instead of f-string interpolation, we could use the Delta Python API `(DeltaTable.merge())` with structured Column expressions, which avoids SQL string construction altogether:

```py
from delta.tables import DeltaTable

if isinstance(self._table.primary_key, str):
    condition = functions.col(f"base.{self._table.primary_key}") == functions.col(
        f"update.{self._table.primary_key}"
    )
elif isinstance(self._table.primary_key, list):
    conditions = [
        functions.col(f"base.{col}") == functions.col(f"update.{col}")
        for col in self._table.primary_key
    ]
    condition = conditions[0]
    for c in conditions[1:]:
        condition = condition & c

(
    DeltaTable.forName(get_spark(), self._table.full_table_location())
    .alias("base")
    .merge(source=update_data.alias("update"), condition=condition)
    .whenMatchedUpdateAll()
    .whenNotMatchedInsertAll()
    .execute()
)
```

I chose to not use this approach right now because it would require for another dependency to be added to `databricks-managedtabledataset`.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,8 @@ fix-markdownlint:
 
 # Run test_tensorflow_model_dataset separately, because these tests are flaky when run as part of the full test-suite
 dataset-tests: dataset-doctests
-	cd kedro-datasets && pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile --ignore tests/tensorflow --ignore tests/databricks
+	cd kedro-datasets && pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile --ignore tests/tensorflow --ignore tests/databricks --ignore tests/spark --cov-fail-under=0
+	cd kedro-datasets && pytest tests/spark --cov-config pyproject.toml --cov-append
 	cd kedro-datasets && pytest tests/tensorflow/test_tensorflow_model_dataset.py --no-cov
 	cd kedro-datasets && pytest tests/databricks --no-cov
 


kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
```diff
@@ -561,10 +560,9 @@ def _save_upsert(self, update_data: DataFrame) -> None:
                 )
 
             update_data.createOrReplaceTempView("update")
-            get_spark().conf.set("fullTableAddress", self._table.full_table_location())
-            get_spark().conf.set("whereExpr", where_expr)
-            upsert_sql = """MERGE INTO ${fullTableAddress} base USING update ON ${whereExpr}
-                WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *"""
+            full_table = self._table.full_table_location()
+            upsert_sql = f"""MERGE INTO {full_table} base USING update ON {where_expr}
+                WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *"""  # nosec B608
```
ravi-kumar-pilla, line 565: The change looks good and as you mentioned in the description regarding nosec, the fstring seems to be the way to go. I could not find any other approach without installing other dependencies like you mentioned. If the team feels fstring is fine, we can go ahead with this. Thank you

kedro-datasets/RELEASE.md
```diff
@@ -6,6 +6,8 @@
 
 ## Bug fixes and other changes
 
+- Fixed `databricks.ManagedTableDataset` upsert write mode failing with `[CONFIG_NOT_AVAILABLE]` on Databricks Spark Connect runtimes by replacing `spark.conf.set` variable substitution with direct f-string interpolation in the MERGE SQL statement.
+- Fixed `ibis.TableDataset` `exists` method to account for `database` (i.e. the collection of tables, or schema).
```
ravi-kumar-pilla, line 10: is this release note intentional ?

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
```diff
@@ -561,10 +560,9 @@ def _save_upsert(self, update_data: DataFrame) -> None:
                 )
 
             update_data.createOrReplaceTempView("update")
-            get_spark().conf.set("fullTableAddress", self._table.full_table_location())
-            get_spark().conf.set("whereExpr", where_expr)
-            upsert_sql = """MERGE INTO ${fullTableAddress} base USING update ON ${whereExpr}
-                WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *"""
+            full_table = self._table.full_table_location()
```
ravi-kumar-pilla: [Nit] I see self._table.full_table_location() being used couple of times in `_save_upsert` , it might be worth reusing it and also making sure self and self._table exists before accessing full_table_location()

kedro-datasets/RELEASE.md
lrcouto: There was a merge conflict when I pushed this PR and I just kept all of the present notes. I can remove it if it's wrong.

kedro-datasets/kedro_datasets/databricks/_base_table_dataset.py
lrcouto: Valid point, pushed a fix for it now.

lrcouto:
Build failure is blocking this PR. (https://github.com/kedro-org/kedro-plugins/actions/runs/22810262655/job/66165804473)

Issue #1322, opened 2026-02-25
`OpikEvaluationDataset`: production implementation


ElenaKhaustova:
## Description
Parent https://github.com/kedro-org/kedro/issues/5036

Should be started after completion of:
- https://github.com/kedro-org/kedro-plugins/issues/1320
- https://github.com/kedro-org/kedro-plugins/issues/1321

The goal is to deliver a production-quality **experimental** `OpikEvaluationDataset` aligned with the finalised `Langfuse` design.

This includes:

- Aligning API surface with `LangfuseEvaluationDataset` (see https://github.com/kedro-org/kedro-plugins/issues/1321)
- Handling Opik-specific constraints
- Validating cross-platform abstraction `OpikEvaluationDataset` based on POC findings.

## Context
A POC `OpikEvaluationDataset` was created to validate dataset-based abstraction portability. Trade-offs and platform differences were documented there. This ticket converts the POC into a production-ready experimental dataset.

### 1. Implementation

- Clean, hardened dataset class
- Consistent interface with the `Langfuse` dataset
- Explicit sync/reproducibility behaviour

### 2. Lifecycle Strategy Alignment

Based on:

- Langfuse production decision
- `Opik` API maturity
- Trade-offs identified in POC

We must determine:

- Whether lifecycle abstraction is feasible
- Whether delegation to the native `Opik` dataset API is preferable

### 3. Cross-Platform Validation

Ensure:

- Same Kedro pipeline works unchanged
- Only the dataset class and nodes swap are required
- No orchestration-level differences

## Deliverables

- Production-ready experimental dataset implementation
- Documented lifecycle strategy
- Cross-platform comparison summary
- Updated E2E example in the academy repo (agent evaluation for intent detection or response generation). See [kedro-agentic-workflows](https://github.com/kedro-org/kedro-academy/tree/main/kedro-agentic-workflows) project
- Documentation for experimental datasets

PR #1364, opened 2026-03-31
feat(datasets): Add `OpikEvaluationDataset` to experimental datasets

lrcouto:
## Description

Introduces a production-ready `OpikEvaluationDataset` to the experimental datasets in `kedro-plugins`, closing issue [#1322](https://github.com/kedro-org/kedro-plugins/issues/1322). This implementation is the Opik counterpart to `LangfuseEvaluationDataset` ([#1347](https://github.com/kedro-org/kedro-plugins/pull/1347)) and follows the same design principles and API surface.

### Key Features

**Implementation changes:**
- Validates credentials (`api_key` required), sync policy, and filepath extension at initialization
- Explicit sync modes: `local` (default, local file as source of truth - upserts all items to remote on every `load()`) and `remote` (Opik as source of truth, no local file interaction)
- Items are deduplicated by content hash on the remote side (Opik SDK behavior); items with `id` fields are also deduplicated locally during `save()` merge
- Warnings issued for items without `id` fields that cannot be tracked across syncs

**Opik-specific behaviour:**
- Opik requires item IDs to be valid UUIDs. Human-readable IDs from local files are stripped before uploaded, as Opik auto-generates UUIDs. Deduplication is content hash-based and is not affected by the `id` field.
- `metadata` is accepted as a constructor param so the API is the same as in `LangfuseEvaluationDataset`, but Opik's `create_dataset()` does not accept a metadata argument. The value is stored locally and returned by `_describe()` but is not propagated to the remote dataset.
- No snapshot versioning: Opik does not support pinning `load()` to a historical snapshot (unlike Langfuse's ISO 8601 `version` param).

**Design philosophy:**

Lifecycle operations (delete dataset, delete items, clear) are delegated to the native Opik SDK rather than exposed through `OpikEvaluationDataset`. This matches what `LangfuseEvaluationDataset` does and keeps the dataset class focused on the local/remote sync. Item deletion requires Opik's internal UUIDs, which are not tracked in the local file, making a thin wrapper of limited value.

### Testing & Documentation

- Unit tests included in `kedro_datasets_experimental/tests/opik` (mirroring `LangfuseEvaluationDataset` test structure)
- Class docstring with sync policy explanation, item format, catalog YAML examples, and Python API example
- `opik/README.md` extended with an `OpikEvaluationDataset` section and a Langfuse to Opik migration guide covering catalog changes, credential key differences, experiment runner API differences, scorer/task signature differences, and known limitations.

### How to test

- Clone the [`opik-evaluation` branch of kedro-academy](https://github.com/kedro-org/kedro-academy/tree/opik-evaluation)
- Set up credentials in `conf/local/credentials.yml`:

```yaml
opik_credentials:
  api_key: "opik-api-key"
  workspace: "workspace-name"

openai:
  openai_api_key: "openai-api-key"
```

Then we can run the pipeline and check the result on the opik dashboard:

```bash
kedro run --pipeline intent_detection_evaluation_opik --params model_name=gpt-4o
```

Expected in the Opik dashboard:
- A new dataset named `evaluations/intent_agent_evaluation` appears under Datasets
- Items are visible with auto-generated UUIDs (human-readable IDs are stripped, this is expected)
- An experiment run appears under Experiments with scoring results per item

image

Run the pipeline a second time without changing the local file. Content hash deduplication means no new items are created in Opik, the item count stays the same.

Expected: dataset item count unchanged; a new experiment run is recorded.

image

Then, add a new item to `data/intent_detection/evaluation/intent_evaluation.json` and rerun. The new item should appear in the remote dataset after load.

image

**A note on versioning:**

In the Langfuse evaluation pipeline, experiment names are derived from a stable integer version number pinned in the catalog (`load_args.version`). This makes experiment runs reproducible and traceable to a specific prompt version.

The Opik equivalent is not yet available. `OpikPromptDataset` currently loads the latest prompt commit by default and exposes its commit hash at runtime, so the kedro-academy pipeline derives the experiment name from that hash (e.g. `intent_eval_a3f2c1b0_model_gpt-4o`). This works in practice but has the drawback of now having verson pinning on the catalog. There is no way to declaratively pin `OpikPromptDataset` to a specific prompt version the way Langfuse does with `load_args.version: 1`. Swapping prompt versions requires editing the local file and re-syncing, rather than a catalog change.

This works to test versioning directly using the Opik SDK, but to get something equivalent to what's being done with `LangfuseEvaluationDataset` we'd have to implement version pinning for `OpikPromptDataset`. There's an issue open on it already (kedro-org/kedro-plugins#1348).

## Checklist


kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,460 @@
...
+    run experiments. Supports an optional local JSON/YAML file as the
+    authoring surface for evaluation items.
+
+    **On load / save behaviour:**
+
+    - **On load:** Creates the remote dataset if it does not exist,
+      synchronises based on ``sync_policy``, and returns an ``opik.Dataset``.
+    - **On save:** Upserts all items to the remote dataset. In ``local`` mode,
+      items are also merged into the local file (new items take precedence).
+      In ``remote`` mode, only the remote upload occurs.
```
ravi-kumar-pilla: [Nit]  In ``remote`` mode, only the remote upsert occurs. ??

ravi-kumar-pilla:
Hi @lrcouto ,

The code looks good. Few things observed as discussed during testing the functionality -

1. Sync mode`local`: On initial run when there is no dataset in remote, I bumped into **`DatasetError` - Failed while loading data from dataset, Connection refused** which I am unable to reproduce now, even after deleting datasets from remote 

2. Sync mode `remote`: I deleted remote dataset and tried with remote sync. Based on the design, I thought it will throw an error as it cannot fetch remote ds. But instead it did not throw any error. It created a ds in remote without items. I do not remember the Langfuse behavior but we should have similar behavior here for consistency. Irrespective of this, the remote sync mode is not creating items if we use `remote` sync mode to start with. It actually make sense as remote mode does not interact with local file, but may be we should throw the error when remote ds is missing.

Thank you

lrcouto:
> Hi @lrcouto ,
> 
> The code looks good. Few things observed as discussed during testing the functionality -
> 
>     1. Sync mode`local`: On initial run when there is no dataset in remote, I bumped into **`DatasetError` - Failed while loading data from dataset, Connection refused** which I am unable to reproduce now, even after deleting datasets from remote
> 
>     2. Sync mode `remote`: I deleted remote dataset and tried with remote sync. Based on the design, I thought it will throw an error as it cannot fetch remote ds. But instead it did not throw any error. It created a ds in remote without items. I do not remember the Langfuse behavior but we should have similar behavior here for consistency. Irrespective of this, the remote sync mode is not creating items if we use `remote` sync mode to start with. It actually make sense as remote mode does not interact with local file, but may be we should throw the error when remote ds is missing.
> 
> 
> Thank you

Hey Ravi, thanks for the review!

Re: creating a dataset on remote without items, and then on the second run working normally. That seems to be a known behavior from the Opik SDK. dataset.insert() is async, so it queues items in a background thread that pushes them to the server. When evaluate() runs right after, the items haven't been committed yet, so they don't appear on the dashboard. I'm seeing if I can make it stop doing that.

Re: LangfuseEvaluationDataset's implementation, it also has this "get or create" behavior:

https://github.com/kedro-org/kedro-plugins/blob/b211f045df8bd0bda147354b59614ad337a2a741/kedro-datasets/kedro_datasets_experimental/langfuse/langfuse_evaluation_dataset.py#L290

So maybe we can maintain it like this on OpikEvaluationDataset too to make it predictable.

And re: the ERRCONREFUSED you saw, I couldn't reproduce it here either. Maybe it was actually a networking hiccup on the server side? 🤔 

I'll make the error a bit more explicit regardless.

lrcouto:
> Hi @lrcouto ,
> 
> Thank you for addressing the issue. I do not see any change with `remote` sync mode. The `local` sync mode works as expected. For `remote` sync mode, on initial run when there are no remote datasets and when the catalog entry is like below -
> 
> ```yaml
> opik_intent_evaluation_data:
>   type: kedro_agentic_workflows.datasets.opik_evaluation_dataset.OpikEvaluationDataset
>   dataset_name: evaluations/intent_agent_evaluation
>   filepath: data/intent_detection/evaluation/intent_evaluation.json
>   sync_policy: remote
>   credentials: opik_credentials
> ```
> 
> This creates an empty dataset on opik without any items. If this is the intended behavior then it looks good.
> 
> > When evaluate() runs right after, the items haven't been committed yet
> 
> When you run in `remote` sync mode any number of times without running `local` mode atleast one time, there are no items pushed to remote. I think this might be the right behavior but users should be informed of this. They should atleast run `local` sync mode one time incase they do not create a dataset via Opik UI.

Sounds fair, I'll modify the docstring to add this.

ravi-kumar-pilla (approved):
Thank you @lrcouto

kedro-datasets/RELEASE.md
```diff
@@ -15,6 +15,7 @@
 | Type                                 | Description                                          | Location                               |
 | ------------------------------------ | ---------------------------------------------------- | -------------------------------------- |
 | `langfuse.LangfuseEvaluationDataset` | A dataset for managing Langfuse evaluation datasets. | `kedro_datasets_experimental.langfuse` |
+| `opik.OpikEvaluationDataset`         | A dataset for managing Opik evaluation datasets.     | `kedro_datasets_experimental.opik`     |
```
ElenaKhaustova: This should be in the upcoming release

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+        "input": {"text": "cancel my order"},
+        "expected_output": "cancel_order",
+        "metadata": {"source": "production"}
+      }
+    ]
+    ```
+
+    **Sync policies:**
+
+    - **local** (default): The local file is the source of truth. On
```
ElenaKhaustova, line 86: The wording suggests Langfuse-style ID upserts on every load/save. The implementation uses `dataset.insert()` and (currently) removes id for all items. That combination does not guarantee `replace the same remote row when I change the local item but keep the same string id.`

**Suggestion**: Rephrase to describe what actually happens: re-insert from local on each sync, optional Opik-side dedup for unchanged content, and (if you adopt the UUID proposal below) stable remote rows only when a valid UUID id is sent. Avoid promising generic `upsert by local id.`

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+      items are also merged into the local file (new items take precedence).
+      In ``remote`` mode, only the remote upsert occurs.
+
+    **Item format:**
+
+    The local file and ``save()`` data must be a list of dicts. Each item
+    accepts the following keys:
+
+    - ``input`` (**required**) — the evaluation input payload.
+    - ``id`` — stable identifier used for local deduplication. Note: Opik
```
ElenaKhaustova: **Proposal:**
Clarify two layers: 

- local file: any id can be used for _merge_items / authoring;
- remote: only IDs Opik accepts should be sent.

Then document explicitly: Send id to Opik only when it is a valid UUID; otherwise omit id on upload so Opik can assign one. Update this bullet after `_upload_items` matches that rule.

See comment on `_upload_items` below.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+        Raises:
+            DatasetError: If any item is missing the ``input`` key.
+        """
+        for i, item in enumerate(items):
+            if "input" not in item:
+                raise DatasetError(
+                    f"Dataset item at index {i} is missing required 'input' key."
+                )
+
+    def _upload_items(self, dataset: Dataset, items: list[dict[str, Any]]) -> None:
```
ElenaKhaustova, line 328: Shall we consider the following implementation as it feels like we should not drop ids in case they are valid?

For each item, if id is present and parses as a valid UUID, include it in the payload passed to insert. If id is missing or not a valid UUID (e.g. human-readable strings), omit id on upload so Opik auto-generates and the API is not fed invalid IDs.

If we go ahead, refresh the docstring as well so it no longer says all IDs are stripped or that dedup is always independent of id, if UUIDs are preserved.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+        Callers are responsible for validating items before calling this method.
+        """
+        items_to_insert = [
+            {k: v for k, v in item.items() if k != "id"}
+            for item in items
+        ]
+        dataset.insert(items_to_insert)
+
+    def _sync_local_to_remote(self, dataset: Dataset) -> Dataset:
+        """Upsert all local items to the remote dataset.
```
ElenaKhaustova: The name/doc says “upsert,” which may overload users’ expectations vs Langfuse. Shall we replace it with something like: “insert items from local; remote identity depends on whether UUID id is sent”)?

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+        if not self._filepath or not self._filepath.exists():
+            return dataset
+
+        local_items = self.file_dataset.load()
+        self._validate_items(local_items)
+
+        if not local_items:
+            return dataset
+
+        items_without_id = [item for item in local_items if "id" not in item]
```
ElenaKhaustova: Shall we also warn when `id` is `None` or empty string, so users get feedback before they get surprising duplicate remote rows?

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+            dataset = self._sync_local_to_remote(dataset)
+
+        logger.info(
+            "Loaded dataset '%s' (sync_policy='%s').",
+            self._dataset_name,
+            self._sync_policy,
+        )
+        return dataset
+
+    def save(self, data: list[dict[str, Any]]) -> None:
```
ElenaKhaustova, line 523: `save()` / `load()` docstrings mention `DatasetError` for failures, and `_get_or_create_remote_dataset()` wraps errors, but insert / flush may still raise raw SDK exceptions.
Shall we wrap Opik API errors in `DatasetError` here too for a single, predictable error type for pipeline code?

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,477 @@
...
+        self._upload_items(dataset, data)
+        self._client.flush()
+
+        if self._sync_policy == "local" and self._filepath:
+            existing: list[dict] = []
+            if self._filepath.exists():
+                existing = self.file_dataset.load()
+            self.file_dataset.save(self._merge_items(existing, data))
+
+    def _exists(self) -> bool:
```
ElenaKhaustova, line 559: `_exists()` handles `ApiError` specially but may let other failures bubble. `_get_or_create_remote_dataset()` maps more failures to `DatasetError`.
We may use the same pattern in `_exists()` so “can’t reach Opik” behaves consistently whether the user calls `load()` or `_exists()`.

kedro-datasets/kedro_datasets_experimental/tests/opik/test_opik_evaluation_dataset.py
```diff
@@ -0,0 +1,640 @@
+import json
```
ElenaKhaustova, line 2: If comment `_upload_items` change is implemented, add or adjust tests so that:

- a valid UUID in id is still present in the payload passed to insert;
- a non-UUID string id is omitted from that payload;
- optionally: same UUID, changed content, two uploads — still passes the same id (guards regressions).
- Update tests that currently assume every id is always stripped.

ElenaKhaustova (commented):
Thank you, @lrcouto!

I made a first pass and left a few suggestions: 

- tighten docs to match insert semantics;
- propose keeping only valid UUID id values on upload (omit others)
- align error handling with DatasetError
- extend warnings/tests
- refresh README migration notes accordingly.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
lrcouto: Sounds good. I'll add a couple of tests for that as well.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,569 @@
...
+      string) are stripped before upload — Opik auto-generates a UUID v7 in
+      those cases. Items without an ``id`` are uploaded without one and Opik
+      assigns a UUID v7 automatically.
+    - ``expected_output`` — ground-truth value for scoring.
+    - ``metadata`` — arbitrary metadata dict attached to the item.
+
+    ```json
+    [
+      {
+        "id": "q1",
```
ElenaKhaustova, line 75: nit: Since "q1" is a human-readable string, it will be stripped before upload and won't give a stable remote row. Worth adding a one-line note under the example, e.g.: "q1" is used for local deduplication only — it is not a UUID v7 and will be stripped on upload. Same applies to the Python example at line 137.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,569 @@
...
+      }
+    ]
+    ```
+
+    **Sync policies:**
+
+    - **local** (default): The local file is the source of truth. On
+      ``load()``, all local items are re-inserted to remote on every sync.
+      The Opik SDK deduplicates by content hash — items whose content has
+      not changed since the last sync are ignored. If a local item's content
```
ElenaKhaustova: This sentence is accurate when the id is non-UUID (stripped before upload), but may be misleading when the id is a valid UUID v7. In that case the code forwards the same id on every insert — does Opik's `dataset.insert()` create a second row with the same UUID, or does it update the existing one? The answer changes what users should expect.

Could you verify against Opik's API and split this into two cases?

- Non-UUID id: stripped → changed content creates a new remote row.
- UUID v7 id: forwarded → [update existing row / create duplicate — whichever is correct].

Same concern applies to `load()` docstring at line 466 and `_upload_items` docstring at line 313.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,569 @@
...
+        Raises:
+            DatasetError: If the Opik API returns an error or the server is
+                unreachable during insert.
+        """
+        items_to_insert = []
+        for item in items:
+            if "id" not in item:
+                items_to_insert.append(item)
+            elif not item["id"]:
+                items_to_insert.append({k: v for k, v in item.items() if k != "id"})
```
ElenaKhaustova: nit: This dict-comprehension-to-strip-id pattern appears three times. Consider extracting a small helper like:

```python
def _strip_id(item):
    return {k: v for k, v in item.items() if k != "id"}
```

kedro-datasets/kedro_datasets_experimental/tests/opik/test_opik_evaluation_dataset.py
```diff
@@ -0,0 +1,791 @@
+import datetime
+import json
+from unittest.mock import Mock, call, patch
```
ElenaKhaustova: nit: call appears unused in this file

kedro-datasets/kedro_datasets_experimental/tests/opik/test_opik_evaluation_dataset.py
```diff
@@ -0,0 +1,791 @@
...
+
+    def test_non_id_fields_are_preserved(self, dataset_local, mock_remote_dataset, eval_items):
+        """input and expected_output fields are passed through unchanged."""
+        dataset_local._upload_items(mock_remote_dataset, eval_items)
+
+        inserted = mock_remote_dataset.insert.call_args[0][0]
+        assert inserted[0]["input"] == eval_items[0]["input"]
+        assert inserted[0]["expected_output"] == eval_items[0]["expected_output"]
+
+    def test_valid_uuid_ids_are_preserved(self, dataset_local, mock_remote_dataset, eval_items_uuid):
```
ElenaKhaustova: nit: This test only uses UUID v7 IDs (which is correct per `REQUIRED_UUID_VERSION`). Consider renaming to `test_valid_uuid_v7_ids_are_preserved`, so the name doesn't imply v4 or other versions would also be preserved.

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -434,3 +434,322 @@ DatasetError: Remote sync policy specified but no remote prompt exists in Opik
...
+### Core Features
+
+#### Item Format
+
+The local file and `save()` data must be a list of dicts:
+
+```json
+[
+  {
+    "id": "q1",
```
ElenaKhaustova, line 486: Same as the code docstring: "q1" is local-only and will be stripped on upload. A short note would help readers connect this example to the UUID v7 rule described in the table.

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -434,3 +434,322 @@ DatasetError: Remote sync policy specified but no remote prompt exists in Opik
...
+
+**Solution:** Populate the remote dataset first using one of:
+- Run the pipeline once with `sync_policy="local"` to seed from your local file, then switch back to `"remote"`.
+- Create and populate the dataset directly via the [Opik UI](https://www.comet.com/docs/opik/).
+
+---
+
+## Migrating from LangfuseEvaluationDataset to OpikEvaluationDataset
+
+`OpikEvaluationDataset` and `LangfuseEvaluationDataset` share the same constructor signature and local file format. Migrating is a catalog swap plus evaluation pipeline node changes.
```
ElenaKhaustova, line 634: This is a good summary, but tne behavioral difference worth calling out here: Langfuse forwards any string id to the API for upsert, while Opik only forwards UUID v7 id values (all others are stripped). So item identity/update semantics differ even though the file format is the same. This is mentioned already, but a short heads-up here would prevent someone from skipping to the migration steps and missing it.

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -434,3 +434,322 @@ DatasetError: Remote sync policy specified but no remote prompt exists in Opik
...
+| `input` | Yes | The evaluation input payload |
+| `id` | No | Used for local deduplication. If `id` is a valid **UUID v7** it is forwarded to Opik, giving the remote row a stable identity. Opik requires UUID v7 for item IDs — all other values (human-readable strings, UUIDs of other versions, `None`, or empty string) are stripped before upload and Opik auto-generates a UUID v7. Items without a stable UUID v7 `id` create new remote rows on every sync. |
+| `expected_output` | No | Ground-truth value for scoring |
+| `metadata` | No | Arbitrary metadata dict attached to the item |
+
+#### Sync Policies
+
+| Policy | Local File | Remote (Opik) | Use Case |
+|--------|------------|---------------|----------|
+| **`local`** (default) | Source of truth | Re-inserted from local on every `load()`. Unchanged items are deduplicated by content hash; changed items create new remote rows. | Authoring, development, initial seeding |
```
ElenaKhaustova: This blanket statement may not hold when a UUID v7 id is forwarded — depending on Opik's insert behavior, it might update the existing row instead of creating a new one. Could we qualify this: "changed items without a stable UUID v7 id create new remote rows"?

ElenaKhaustova (commented):
Thank you for addressing the comment! This looks good overall.

I left a few new comments, mostly around documentation precision. The main one: the docs claim both that UUID v7 gives "stable remote identity" and that changed content always creates new remote rows — those may contradict each other depending on what Opik's insert does with an existing UUID v7. Worth a quick check against their API docs. The rest are small wording tweaks and test nits.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
lrcouto: Checked the SDK here and for UUID v7 it does update the existing row instead of creating a duplicate, I'll make that clear.

lrcouto:
@ElenaKhaustova Addressed the latest comments.

ElenaKhaustova:
Thank you for addressing all the comments, @lrcouto! The changes look great overall — the UUID v7 split, `_strip_id` helper, error handling, test cleanup, and migration callout are all well done.

One thing I noticed: there's an internal doc inconsistency that crept in while splitting the UUID v7 vs non-UUID cases.

The `_sync_local_to_remote` docstring says:

> The Opik SDK deduplicates by content hash, so re-inserting unchanged items is a no-op.

But the class docstring, `_upload_items` docstring, and README all say the opposite for non-UUID items:

> Opik auto-generates a new UUID v7 on every sync, so a **new remote row is created on every sync** regardless of whether the content has changed.

These contradict each other, and your own PR testing confirms content-hash dedup works ("Run the pipeline a second time without changing the local file. Content hash deduplication means no new items are created in Opik, the item count stays the same.").

The actual behavior is:

| Scenario | UUID v7 `id` | No UUID (stripped) |
|---|---|---|
| Content **unchanged** | No-op (upsert, nothing to update) | No-op (content-hash dedup) |
| Content **changed** | Update existing row in-place | New row created (old row remains) |

So the key advantage of UUID v7 isn't about unchanged content (both paths are no-ops) — it's about **changed content**: UUID v7 updates the row in-place, while non-UUID creates a new row and the old one persists, leading to row accumulation over time.

**Places to fix** (replace "regardless of whether content has changed" / "even for unchanged content" with something like "when content changes, a new remote row is created and the previous one remains"):

- Class docstring (lines 62–64)
- `_upload_items` docstring (lines 333–335)
- `_sync_local_to_remote` docstring (lines 374–375) — this one is closer to correct but should mention the UUID v7 vs non-UUID distinction for changed content
- `load()` docstring (lines 489–490)
- README item format table (`id` field notes, line 498)
- README sync policy table (`local` row, line 506)

Once that's aligned, this is good to go from my side!

lrcouto:
@ElenaKhaustova see if it looks OK now. I tried to make it extremely explicit that a non-UUIDv7 id will create a new row and leave the old one in place, while a UUIDv7 will not create a new row and update the existing one.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,599 @@
...
+    - ``id`` — identifier used for local deduplication. The upload
+      behaviour depends on whether ``id`` is a valid UUID v7:
+
+      - **Valid UUID v7**: forwarded to Opik. Opik's API upserts by item
+        ID — the first sync creates the remote row; subsequent syncs
+        update that same row in-place if the content has changed.
+        The remote row keeps the same UUID across all syncs. Whenever
+        content changes, the existing remote row is updated in-place,
+        while no new row is created.
+      - **All other values** (human-readable strings, UUIDs of other
```
ElenaKhaustova:
```suggestion
     - **All other values** (human-readable strings, UUIDs of other
        versions, ``None``, empty string, or no ``id`` key): stripped
        before upload. Opik auto-generates a new UUID v7. Unchanged
        content is deduplicated by content hash (no-op), but changed
        content creates a **new remote row** while the previous one
        remains, leading to row accumulation over time.
```

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,599 @@
...
+
+    - **local** (default): The local file is the source of truth. On
+      ``load()``, all local items are re-inserted to remote on every sync.
+      Opik's API upserts by item ID, so the outcome depends on whether
+      each item carries a UUID v7 ``id``:
+
+      - Items with a UUID v7 ``id`` are updated in-place on the remote —
+        content changes replace the existing row; unchanged items are
+        a no-op.
+      - Items without a UUID v7 ``id`` (non-UUID values are stripped)
```
ElenaKhaustova, line 94:
```suggestion
      - Items without a UUID v7 ``id`` (non-UUID values are stripped)
        are deduplicated by content hash — unchanged content is a no-op,
        but changed content creates a **new remote row** (the previous
        row remains), leading to row accumulation over time.
```

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,599 @@
...
+        """Insert items into the remote Opik dataset.
+
+        Upload behaviour depends on whether an item carries a UUID v7 ``id``:
+
+        - **Valid UUID v7**: forwarded to Opik. Opik's REST API calls
+          ``create_or_update`` by item ID — the first call creates the
+          remote row; subsequent calls update that same row in-place if
+          the content has changed. Whenever content changes, the existing
+          remote row is updated in-place, while no new row is created.
+        - **All other values** (human-readable strings, UUIDs of other
```
ElenaKhaustova, line 338:
```suggestion
        - **All other values** (human-readable strings, UUIDs of other
          versions, ``None``, empty string, or no ``id`` key): stripped
          before upload. Opik auto-generates a new UUID v7. Unchanged
          content is deduplicated by content hash (no-op), but changed
          content creates a **new remote row** while the previous one
          remains.
```

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,599 @@
...
+        return merged
+
+    def load(self) -> Dataset:
+        """Load the Opik dataset, syncing local items to remote if sync_policy is ``local``.
+
+        Creates the remote dataset if it does not exist. In ``local`` mode, all
+        local items are re-inserted to remote on every load via Opik's
+        ``create_or_update`` API (upsert by item ID). On items with a valid UUID v7
+        ``id``, update the existing remote row in-place, and no new row is created.
+        On items where the ``id`` is not a valid UUID v7 (including missing, ``None``, or empty),
```
ElenaKhaustova, line 499:
```suggestion
        On items where the ``id`` is not a valid UUID v7 (including missing, ``None``, or empty),
        the ``id`` is stripped before upload and Opik auto-generates a new UUID v7.
        Unchanged content is deduplicated (no-op), but changed content creates a
        new remote row while the previous one remains.
```

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -434,3 +434,325 @@ DatasetError: Remote sync policy specified but no remote prompt exists in Opik
...
+    "metadata": {"source": "production"}
+  }
+]
+```
+("q1" is used for local deduplication only, as it is not a UUID v7 and will be stripped on upload)
+
+| Field | Required | Notes |
+|-------|----------|-------|
+| `input` | Yes | The evaluation input payload |
+| `id` | No | Used for local deduplication. Upload behaviour depends on the value: **valid UUID v7** — forwarded to Opik; Opik's API upserts by item ID, so the first sync creates the remote row and subsequent syncs update it in-place (when the content is changed, the already existing row is updated). **All other values** (human-readable strings, other UUID versions, `None`, empty string, or absent) — stripped before upload; Opik auto-generates a new UUID v7 every sync, so a new remote row is created on every sync regardless of content, while the already existing row remains in place. |
```
ElenaKhaustova:
```suggestion
| `id` | No | Used for local deduplication. Upload behaviour depends on the value: **valid UUID v7** — forwarded to Opik; Opik's API upserts by item ID, so the first sync creates the remote row and subsequent syncs update it in-place (when the content is changed, the already existing row is updated). **All other values** (human-readable strings, other UUID versions, `None`, empty string, or absent) — stripped before upload; Opik auto-generates a new UUID v7 every sync. Unchanged content is deduplicated by content hash (no-op), but changed content creates a new remote row while the previous one remains, leading to row accumulation over time. |
```

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -434,3 +434,325 @@ DatasetError: Remote sync policy specified but no remote prompt exists in Opik
...
+| `expected_output` | No | Ground-truth value for scoring |
+| `metadata` | No | Arbitrary metadata dict attached to the item |
+
+#### Sync Policies
+
+| Policy | Local File | Remote (Opik) | Use Case |
+|--------|------------|---------------|----------|
+| **`local`** (default) | Source of truth | All local items are re-inserted on every `load()` via Opik's upsert-by-ID API. Items with a UUID v7 `id` update the existing remote row in-place, without creating a new row; items without a UUID v7 `id` receive a new auto-generated UUID on every sync and create a new remote row each time, while the previous row still remains. | Authoring, development, initial seeding |
```
ElenaKhaustova:
```suggestion
| **`local`** (default) | Source of truth | All local items are re-inserted on every `load()` via Opik's upsert-by-ID API. Items with a UUID v7 `id` update the existing remote row in-place, without creating a new row; items without a UUID v7 `id` are deduplicated by content hash — unchanged content is a no-op, but changed content creates a new remote row (the previous row remains), leading to row accumulation over time. | Authoring, development, initial seeding |
```

ElenaKhaustova (commented):
Thanks, @lrcouto!

There are a few places that still look confusing to me (mostly because of how Opik works). I've suggested the concrete changes. Let me know what you think.

lrcouto:
> Thanks, @lrcouto!
> 
> There are a few places that still look confusing to me (mostly because of how Opik works). I've suggested the concrete changes. Let me know what you think.

Looks good to me. Can't get any more explicit than that, I think.

ElenaKhaustova:
@lrcouto it looks like you've applied the suggestions but didn't delete the old text and now we have duplicate paragraphs in three places. The new text is there, but the old (incorrect) text remains right below it for class docstring, `_upload_items` and `load` docstrings.

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,604 @@
...
+    - ``id`` — identifier used for local deduplication. The upload
+      behaviour depends on whether ``id`` is a valid UUID v7:
+
+      - **Valid UUID v7**: forwarded to Opik. Opik's API upserts by item
+        ID — the first sync creates the remote row; subsequent syncs
+        update that same row in-place if the content has changed.
+        The remote row keeps the same UUID across all syncs. Whenever
+        content changes, the existing remote row is updated in-place,
+        while no new row is created.
+     - **All other values** (human-readable strings, UUIDs of other
```
ElenaKhaustova:
```suggestion
     - **All other values:** (human-readable strings, UUIDs of other
```

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,604 @@
...
+    **Item format:**
+
+    The local file and ``save()`` data must be a list of dicts. Each item
+    accepts the following keys:
+
+    - ``input`` (**required**) — the evaluation input payload.
+    - ``id`` — identifier used for local deduplication. The upload
+      behaviour depends on whether ``id`` is a valid UUID v7:
+
+      - **Valid UUID v7**: forwarded to Opik. Opik's API upserts by item
```
ElenaKhaustova: Can we please fix indentation here:

Image

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
```diff
@@ -0,0 +1,604 @@
...
+        "metadata": {"source": "production"}
+      }
+    ]
+    ```
+    ("q1" is used for local deduplication only, as it is not a UUID v7 and will be stripped on upload)
+
+    **Sync policies:**
+
+    - **local** (default): The local file is the source of truth. On
+      ``load()``, all local items are re-inserted to remote on every sync.
```
ElenaKhaustova, line 87: Same problem with the indentation

Image

ElenaKhaustova (commented):
Thank you, @lrcouto, everything looks good apart from a few doc rendering issues!

kedro-datasets/kedro_datasets_experimental/opik/opik_evaluation_dataset.py
lrcouto: This one is not missing. The `:` is after the closing parenthesis.

lrcouto:
Should be better now.

image
image

ElenaKhaustova:
This part still looks a bit broken

Screenshot 2026-04-16 at 20 26 01

lrcouto:
Ah ok, I see.

image

ElenaKhaustova (approved):
🚀

Issue #1298, opened 2026-02-04
bug(datasets): ibis.FileDataset does not work with abfss (or other cloud protocols?)


mark-druffel:
## Context
I think FileDataset is butchering cloud filepaths. When I set a filepath to `abfss://`, FileDataset (pathlib) drops the second forward slash making it `abfss:/` [here](https://github.com/kedro-org/kedro-plugins/blob/fcf96d8594d9a526d948b135371af1bd77fae8e6/kedro-datasets/kedro_datasets/ibis/file_dataset.py#L130-L132) and says the the path does not exist. 

The effect is I can read a file using ibis like so, but it will fail in kedro. This works:
```python
import ibis
ispark = ibis.pyspark.connect(session = spark)
my_path = 'abfss://container@account.dfs.core.windows.net/folder/parquet_directory'
df = ispark.read_parquet_dir(my_path)
```
This fails with `[PATH_NOT_FOUND](https://learn.microsoft.com/azure/databricks/error-messages/error-classes#path_not_found)] Path does not exist: dbfs:/Workspace/Users/me@company.com/notebooks/abfss:/container@account.dfs.core.windows.net/folder/parquet_directory. SQLSTATE: 42K03`:
```yaml
my_data:
  type: ibis.FileDataset
  filepath: abfss://container@account.dfs.core.windows.net/folder/parquet_directory
  file_format: parquet_dir
  table_name: my_table
  connection: ${globals:my_backend}
```
## Steps to Reproduce
This is challenging to create. I tried to use an open source dataset on Google, but my databricks environment (work) blocked it so that was kind of a dead end. Given that, I'm just going to provide the code with abstracted paths for now and I'll update my issue if I'm able to get a better reproducible example or debug further on my side.

I setup a catalog entry:
```yaml
my_data:
  type: ibis.FileDataset
  filepath: abfss://container@account.dfs.core.windows.net/folder/parquet_directory
  file_format: parquet_dir
  table_name: my_table
  connection: ${globals:my_backend}
```

I install the project and dependencies in my databricks cluster, instantiate kedro, and call `catalog.load()`:
```python
%pip install -e ../.
dbutils.library.restartPython()
```
```python
import os
os.environ["KEDRO_ENV"] = "dev"
%load_ext kedro.ipython
```
```python
catalog.load("my_data")
```
[...]

PR #1373, opened 2026-04-05
feat: Add file based Hugging Face datasets to `kedro-datasets`

iwhalen:
## Description

Adds datasets for interacting with Hugging Face datasets on a file system.

## Development notes

Added docs, tests, ran in a fresh pipeline.

Iterable and in-memory versions have both been tested as well. 

> [!Note]
> I couldn't figure out a good way to save an `IterableDataset` without looping through it entirely first.
> 
> Maybe there's a better way someone knows about.

Updated `jsonschema/kedro-catalog.1.00.json`.

## Checklist


kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
     def list_datasets():
         api = HfApi()
         return list(api.list_datasets())
+
+
+class LocalHFDataset(AbstractVersionedDataset[DatasetLike, DatasetLike]):
```
deepyaman: Nit: I think most datasets are one-per-module, but some are not, and I don't have any strong aversion to this (although having them in one file hardly saves any duplication).

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+
+    _SUPPORTED_FORMATS = {"arrow", "parquet", "json", "csv", "lance", "hdf5"}
+    _FORMAT_EXTENSIONS = {
+        "arrow": ".arrow",
+        "parquet": ".parquet",
+        "json": ".json",
+        "csv": ".csv",
+        "lance": ".lance",
+        "hdf5": ".h5",
+    }
```
deepyaman:
```suggestion
    _FORMAT_EXTENSIONS = {
        "arrow": ".arrow",
        "parquet": ".parquet",
        "json": ".json",
        "csv": ".csv",
        "lance": ".lance",
        "hdf5": ".h5",
    }
```

This is duplicative; you could just as easily infer supported formats by checking if it's in the dictionary.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+        load_args: dict[str, Any] | None = None,
+        save_args: dict[str, Any] | None = None,
+        credentials: dict[str, Any] | None = None,
+        fs_args: dict[str, Any] | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
+        if file_format not in self._SUPPORTED_FORMATS:
+            msg = (
+                f"Unsupported file_format '{file_format}'. "
+                f"Must be one of {sorted(self._SUPPORTED_FORMATS)}."
```
deepyaman: Nit: I know it's not significant, but it feels unnecessary to perform a sort, when you can just as easily keep the dict keys sorted yourself.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+        self._storage_options = {**_credentials, **_fs_args} or None
+
+        super().__init__(
+            filepath=PurePosixPath(resolved_path),
+            version=version,
+            exists_function=self._fs.exists,
+            glob_function=self._fs.glob,
+        )
+
+    def _load(self) -> DatasetLike:
```
deepyaman:
```suggestion
    def load(self) -> DatasetLike:
```
The modern way.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
deepyaman: Is "Local" a misnomer? E.g. https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_from_disk can also read from a storage bucket (more questions on that later). You've also implemented the fsspec stuff, so I don't think it's accurate.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+        _fs_args = deepcopy(fs_args) or {}
+        _credentials = deepcopy(credentials) or {}
+
+        protocol, resolved_path = get_protocol_and_path(path, version)
+        self._protocol = protocol
+
+        if protocol == "file":
+            _fs_args.setdefault("auto_mkdir", True)
+
+        self._fs = fsspec.filesystem(self._protocol, **_credentials, **_fs_args)
```
deepyaman: The problem with this approach is that Hugging Face also supports remote URIs natively (e.g. https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_from_disk). I think we don't want to use fsspec in those cases, because Hugging Face very well could have a more efficient native path.

This is a hard problem to solve; we have similar concerns on the Ibis side (e.g. https://github.com/kedro-org/kedro-plugins/issues/1298), but intuitively I'd err on the side of leaving it up to the Hugging Face APIs.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+                storage_options=self._storage_options,
+                **self._load_args,
+            )
+
+        ext = self._FORMAT_EXTENSIONS[self._file_format]
+        loader = getattr(Dataset, f"from_{self._file_format}")
+
+        if self._fs.isdir(load_path):
+            paths = {
+                PurePosixPath(p).stem: p for p in self._fs.glob(f"{load_path}/*{ext}")
```
deepyaman: Does it have to have the extension, or is this too narrow? Will people come saying they have HDF5 files with .hdf5 extension instead of .h5?

deepyaman: Actually, look at the C4 example under https://huggingface.co/docs/datasets/en/loading#hugging-face-hub; there's an example with `.json.gz` extension.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+            }
+            return DatasetDict(
+                {
+                    split: loader(path, **self._load_args)
```
deepyaman: It seems you can specify splits in the `data_files` argument of `load_dataset` (e.g. https://huggingface.co/docs/datasets/loading#json); would this be preferable to constructing the `DatasetDict` with multiple `Dataset.from_*` calls? I haven't looked into the implementation, but I'd be curious if there's a good reason to not use the higher-level API.

Referencing the same C4 example from above (https://huggingface.co/docs/datasets/en/loading#hugging-face-hub), it seems you can also pass the wildcard directly under `data_files`.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
deepyaman: How do you get an `IterableDataset`, for example? At the bottom of https://huggingface.co/docs/datasets/en/filesystems, it seems like you could do `IterableDataset.from_dict()`, but it's not clear you're ever doing that, right?

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -65,3 +85,255 @@ def _describe(self) -> dict[str, Any]:
...
+            raise DatasetError(msg)
+
+        if isinstance(data, IterableDatasetDict):
+            data = DatasetDict({k: Dataset.from_list(list(v)) for k, v in data.items()})
+        elif isinstance(data, IterableDataset):
+            data = Dataset.from_list(list(data))
+
+        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+
+        if self._file_format == "arrow":
```
deepyaman: Not a huge fan of the `if isinstance()/elif isinstance()/else` logic; I could be wrong, but it makes me feel like it's multiple datasets in one, since you actually have some distinct logic for each.

Does everything you save/reload roundtrip cleanly?

kedro-datasets/RELEASE.md
```diff
@@ -7,6 +7,17 @@
 
 ## Major features and improvements
 
+- Added `huggingface.LocalHFDataset` to handle saving and loading from Hugging Face datasets on a filesystem.
+
+## Bug fixes and other changes
+## Community contributions
+
+[iwhalen](https://github.com/iwhalen)
```
```suggestion
Many thanks to the following Kedroids for contributing PRs to this release:

- [iwhalen](https://github.com/iwhalen)
```

deepyaman (changes requested):
High-level concerns:

1. Use of fsspec.
2. Bespoke split/partition/multi-file handling (whatever you want to call it).
3. Handling too many types in one.

I think I'd personally it rather be a lightweight wrapper that delegates more to underlying Hugging Face APIs.

Unless you're convinced, or if you disagree, it's probably worth getting a second opinion/review on the above.

iwhalen:
> High-level concerns:
> 
>     1. Use of fsspec.
> 
>     2. Bespoke split/partition/multi-file handling (whatever you want to call it).
> 
>     3. Handling too many types in one.
> 
> 
> I think I'd personally it rather be a lightweight wrapper that delegates more to underlying Hugging Face APIs.
> 
> Unless you're convinced, or if you disagree, it's probably worth getting a second opinion/review on the above.

Thanks for the review @deepyaman! This was a little sloppy you're right.

My first thought was also to try to get the Hugging Face API to handle everything. I'll try again on this and update.

Otherwise, I'll just make a separate dataset for each format HF supports.

Thanks again!

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
iwhalen: Yeah, `FilesystemHFDataset` might be more appropriate.

After some noodling, I think I'm going to split them all into separate datasets by file format. So, this naming will go away regardless.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
iwhalen: Agreed! Implementing this.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
iwhalen: There's two things happening here:

- Loading a dataset from the Hub (handled by `HFDataset` now - its not perfect, but I'm ignoring it in this PR).
- Loading / saving from a filesystem (what I'm trying to implement here).

## Loading

The `load_dataset` can handle all our needed types with a call like:

``` python
load_dataset("csv", data_files="path/to/data.csv")
```

Same goes for parquet, lance, hdf5, json.

Arrow datasets don't play nice like this though and use `load_from_disk("path/to/arrow")`.

## Saving

This procedure works for everything but Arrow:

``` python
data = Dataset.from_dict({"a": [1,2], "b": [3,4]})
data.to_parquet("data.parquet")
load_dataset("parquet", data_files="data.parquet")
```

I couldn't find a way to save a `DatasetDict` to a particular format.

This loop is what happens in the `DatasetDict` itself here: 

https://github.com/huggingface/datasets/blob/4775eeba2d5e73349790f7575182d71d5cd8e1bf/src/datasets/dataset_dict.py#L1376-L1383

Which eventually makes it down to [this](https://github.com/huggingface/datasets/blob/main/src/datasets/arrow_dataset.py#L1942) private method that _only_ will save to Arrow format.

This same goes for the individual `Dataset` objects. You have to use the `to_<format>` methods there's not a single method to rule them all like there is with loading. Lance and HDF5 don't actually have save methods handled by Hugging Face :(

Agree with your point on file name format strictness though.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
iwhalen: See comment below on needing a filesystem. 

In order to save a `DatasetDict` we have to be able to access it.

Hugging Face uses fsspec under the hood anyway 🤷

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
iwhalen: Hmmm... maybe I'll just glob everything in the directory then?

iwhalen:
Ok changes are all up. I think this is a bit better.

- Each dataset has its own file.
- Saving an loading a little simpler.

Unchanged:

- Looping over `DatasetDict` objects to save (have to, see comment).
- Awkward `isinstance` checks to control behavior for `DatasetDict` and `Iterable*` types.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,203 @@
...
+        _fs_args = deepcopy(fs_args) or {}
+        _credentials = deepcopy(credentials) or {}
+
+        protocol, resolved_path = get_protocol_and_path(path, version)
+        self._protocol = protocol
+
+        if protocol == "file":
+            _fs_args.setdefault("auto_mkdir", True)
+
+        self._fs = fsspec.filesystem(self._protocol, **_credentials, **_fs_args)
```
ElenaKhaustova, line 79: The original review flagged the use of fsspec, and I think that concern still applies here. The base class creates its own `fsspec.filesystem()` and then uses `self._fs.isdir()`, `self._fs.glob()`, `self._fs.mkdirs()` to manage files — but HF's `load_dataset()` already accepts `storage_options` and handles remote filesystems internally.

For the non-Arrow formats, I think we should lean harder on HF's APIs and avoid maintaining a parallel filesystem layer. Concretely:
- For loading: `load_dataset("parquet", data_files=..., storage_options=...)` already handles remote paths. We shouldn't need `self._fs.isdir()` / `self._fs.glob()` to discover split files.
- For saving: `to_parquet()` / `to_csv()` / `to_json()` accept `storage_options` directly.
- The fsspec filesystem object should mainly be used for `_exists()` and cache invalidation, not for orchestrating load/save.

This simplifies the code and avoids potential behavior divergence between our fsspec layer and HF's internal one (e.g. around auth, caching, or remote URI schemes).

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,203 @@
...
+
+        if isinstance(data, DatasetDict):
+            self._save_dataset_dict(data, save_path)
+        else:
+            self._save_dataset(data, save_path)
+
+        self._invalidate_cache()
+
+    def _load_dataset(self, load_path: str) -> DatasetLike:
+        if self._fs.isdir(load_path):
```
ElenaKhaustova: `storage_options` is not passed in the directory branch. The single-file branch passes `storage_options=self._storage_options`, but this directory branch doesn't. This means loading a `DatasetDict` from a remote filesystem (S3, GCS) will silently drop credentials.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,203 @@
...
+            ext = self.EXTENSION
+            data_files = {
```
ElenaKhaustova: This glob assumes each split is a single file named `{split}{ext}` (e.g. `train.parquet`, `test.parquet`). That's fragile — HF datasets in the wild often shard splits into multiple files like `train-00000-of-00002.parquet`, which this glob would pick up with the wrong split name. HF's `load_dataset` already handles this internally when given a directory. Could we delegate to it instead?
```python
return load_dataset(
    self.BUILDER,
    data_dir=load_path,
    storage_options=self._storage_options,
    **self._load_args,
)

File: kedro-datasets/kedro_datasets/huggingface/_base.py [hunk 13]
```diff
@@ -0,0 +1,203 @@
...
+            msg = (
+                f"{type(self).__name__} only supports `datasets.Dataset`, "
+                "`datasets.DatasetDict`, "
+                "`datasets.IterableDataset`, and "
+                "`datasets.IterableDatasetDict` instances. "
+                f"Got {type(data)}"
+            )
+            raise DatasetError(msg)
+
+        if isinstance(data, IterableDatasetDict):
```
ElenaKhaustova: This silently materializes the entire iterable into memory via `list()`. The whole point of `IterableDataset` is to handle data that doesn't fit in RAM — silently defeating that guarantee will cause hard-to-debug OOM crashes for users.
I'd suggest either:
- Raise an error telling users to materialize explicitly before saving (e.g. `Dataset.from_list(list(iterable_ds))`) so they make a conscious choice, or
- At minimum, emit a `warnings.warn()` explaining what's happening and the memory implications.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,203 @@
...
+            data_files=load_path,
+            storage_options=self._storage_options,
+            **self._load_args,
+        )
+
+        # load_dataset wraps a single file in a DatasetDict with one
+        # split (typically "train"). When the caller didn't ask for a
+        # specific split, unwrap it so a single file round-trips as a
+        # Dataset, not a DatasetDict.
+        if (
```
ElenaKhaustova: This means `save(dataset_dict_with_one_split)` → `load()` returns a `Dataset`, not a `DatasetDict`. The round-trip changes the type, which is surprising. Users who write generic pipeline code expecting `DatasetDict` back will hit `AttributeError`s.

I'd remove this auto-unwrap. If someone passes a single file, `load_dataset` wrapping it in a `DatasetDict` with a "train" split is standard HF behavior, and users can always pass `load_args: {split: train}` in the catalog to unwrap explicitly.

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -1,13 +1,21 @@
...
+    Dataset,
+    DatasetDict,
+    IterableDataset,
+    IterableDatasetDict,
+    load_dataset,
+)
 from huggingface_hub import HfApi
 from kedro.io import AbstractDataset
 
+DatasetLike: TypeAlias = Dataset | DatasetDict | IterableDataset | IterableDatasetDict
```
ElenaKhaustova, line 15: `DatasetLike` is now defined identically in both `_base.py` and here. One of them should import from the other.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,203 @@
...
+            save_path,
+            storage_options=self._storage_options,
+            **self._save_args,
+        )
+
+    def _save_dataset_dict(self, data: DatasetDict, save_path: str) -> None:
+        self._fs.mkdirs(save_path, exist_ok=True)
+        ext = self.EXTENSION
+        saver = f"to_{self.BUILDER}"
+        for split, split_ds in data.items():
```
ElenaKhaustova, line 148: Worth adding a brief comment explaining _why_ this manual loop exists — HF only provides `DatasetDict.save_to_disk()` for Arrow format; there's no equivalent for parquet/csv/json, so we have to iterate and call `to_<format>()` per split.

ElenaKhaustova (commented):
Thanks for the contribution and for being so responsive to feedback! The split into separate dataset classes makes sense, and it matches how the rest of `kedro-datasets` is organized (e.g. `pandas.CSVDataset`, `pandas.ParquetDataset`). Also, Arrow genuinely behaves differently from the other formats.

A few high-level suggestions before we iterate on the details:

**Scope this PR to the four round-trip formats.** I'd suggest removing `HDF5Dataset` and `LanceDataset` (and their tests/docs) from this PR. Focus on Arrow, Parquet, CSV, and JSON — these all support full save + load round-trips and share the same base class, so they belong together. The read-only formats are a separate concern that deserves a small follow-up PR where we can get the `save()` error handling right (e.g. overriding `save()` directly so the error is raised immediately, rather than letting the base class do type checking, iterable materialization, and path resolution before the error surfaces in `_save_dataset`).

**Deduplicate the tests.** The test files for CSV, JSON, and Parquet are near-identical (differing only in class name and extension). Consider a parametrized shared test instead.

iwhalen:
@ElenaKhaustova Ok! I think I addressed most of the changes. There's still a couple things I'm not in love with.

I see my tests are failing, so I'll get to those. Just wanted to send a note on some higher level things.

## Checking for directory in non-Arrow datasets

In the `FilesystemDataset` we have to make an assumption somewhere on whether or not we're trying to load a `DatasetDict`.

I thought it was reasonable that, if the provided `path` looks like a directory (or _is_ an existing directory), we assume its a directory.

Then, if the user doesn't tell us what we're looking for in the directory, we throw an error.

That's what's happening here in `_validate_load_paths`.

In other words, we can't call `load_dataset("json", data_files="my/directory/")` we have to call `load_dataset("json", data_files={"data": "my/directory/data.json", "labels": "my/directory/labels.json")`. 

I'm not sure if there's a smarter way to do this. Let me know what you think.

## `data_files` convenience processing for non-Arrow datasets

As I say above, the right way to load from a directory is: 

``` python
load_dataset(
    "json", data_files={"data": "my/directory/data.json", "labels": "my/directory/labels.json"
)
```

Where the values in the `data_files` dictionary are full paths to your data.

If we were 100% strict, our `yaml` entries for this dataset would have to look like:

``` yaml
reviews:
  type: huggingface.JSONDataset
  path: data/01_raw/reviews
  load_args:
    data_files:
      labels: data/01_raw/reviews/labels.json
      data: data/01_raw/reviews/data.json
```

Instead, I introduce the helper `_build_data_files` so that the `yaml` entries can look like:

``` yaml
reviews:
  type: huggingface.JSONDataset
  path: data/01_raw/reviews
  load_args:
    data_files:
      labels: labels.json
      data: data.json
```

Maybe this is overstepping! Happy to hear thoughts on this as well.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        return self._load_dataset(load_path)
+
+    def save(self, data: DatasetLike) -> None:
+        if isinstance(data, IterableDataset | IterableDatasetDict):
+            msg = (
+                f"{type(self).__name__} got iterable dataset. "
+                "Before saving an iterable dataset "
+                "you must materialize it into a `Dataset` or `DatasetDict`."
+            )
+            raise RuntimeError(msg)
```
ElenaKhaustova: This should be `DatasetError` instead of `RuntimeError` for consistency — that's the standard exception type for dataset operations throughout kedro-datasets. The type check right below already uses `DatasetError`.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,213 @@
...
+            filepath=PurePosixPath(resolved_path),
+            version=version,
+            exists_function=self._fs.exists,
+            glob_function=self._fs.glob,
+        )
+
+        # For non-Arrow datasets, we have to validate that, if we were given
+        # a directory, the user also provided ``data_files`` in the load_args.
+        filepath_str = get_filepath_str(self._get_load_path(), self._protocol)
+        self._path_is_dir = not PurePosixPath(filepath_str).suffix or self._fs.isdir(
```
ElenaKhaustova: Two concerns with this running in `__init__`:

1. **`self._fs.isdir()` makes a filesystem call during construction.** For remote filesystems (S3, GCS), this is a network round-trip every time the dataset is instantiated — including when just building the catalog. It will also return `False` if the path doesn't exist yet.

2. **`_validate_load_paths` blocks save-first workflows.** You can't construct `CSVDataset(path="data/reviews")` without providing `data_files` — even if the intent is to save a `DatasetDict` first and load later.

Could we defer both checks to `load()` time instead? That way construction always succeeds, and the error only surfaces when the user actually tries to load without the required config. You could determine "is this a directory?" based on whether the user provided `data_files` in `load_args` — that's the real signal for "I'm working with a directory of splits," and it avoids the filesystem call entirely.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+
+        if isinstance(data, DatasetDict):
+            self._save_dataset_dict(data, save_path)
+        else:
+            self._save_dataset(data, save_path)
+
+        self._invalidate_cache()
+
+    def _build_data_files(self) -> str | dict[str, str]:
```
ElenaKhaustova, line 142: The relative path resolution is nice. One thing to watch: `_save_dataset_dict` creates files as `{split}{ext}`, so the filenames are derived from the `DatasetDict` split names. If the keys in `data_files` don't match the split names in the saved `DatasetDict`, load will fail with a confusing "file not found" error.

Worth either:
- Adding a note in the docstring that `data_files` keys must match the `DatasetDict` split names, or
- Validating at save time that the saved split names match `data_files` keys (if configured).

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
ElenaKhaustova: Still defined in both `_base.py` and here.

kedro-datasets/kedro_datasets/huggingface/arrow_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+from ._base import DatasetLike, FilesystemDataset
+
+
+class ArrowDataset(FilesystemDataset):
+    """``ArrowDataset`` loads/saves Hugging Face ``Dataset`` and
+    ``DatasetDict`` objects to/from disk in
+    `Arrow <https://huggingface.co/docs/datasets/about_arrow>`_ format
+    using ``save_to_disk`` / ``load_from_disk``.
+
+    Iterable variants (``IterableDataset``, ``IterableDatasetDict``)
```
ElenaKhaustova: The docstring still says "Iterable variants (`IterableDataset`, `IterableDatasetDict`) are materialised before saving." — but the behavior was changed to raise an error instead. Should be updated to say that iterable variants must be materialized by the user before saving.

ElenaKhaustova (commented):
Thanks for the updates @iwhalen — this is much improved!

On your two questions:

**Directory loading with `data_files`:** You're right — my earlier suggestion to use `data_dir` was too optimistic about what HF handles automatically for non-Arrow formats. Your approach of requiring explicit `data_files` is the correct middle ground: it removes the fragile glob-based discovery while still delegating the actual loading to `load_dataset`. 

**`_build_data_files` helper:** — this is a good UX call.

A few remaining items in inline comments below.

iwhalen:
@ElenaKhaustova seems like we're getting closer!

I addressed the changes you had above and fixed my doctest issues.

Now I'm seeing some failures in CI that seem to be unrelated to the changes in this branch.

Any advice?

kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py
```diff
@@ -1,13 +1,21 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TypeAlias
 
-from datasets import load_dataset
+from datasets import (
```
ElenaKhaustova: Small cleanup: now that `DatasetLike` is imported from `_base`, the `Dataset`, `DatasetDict`, `IterableDataset`, `IterableDatasetDict` imports from `datasets` are no longer used here, and `TypeAlias` is no longer needed either. Can be reduced to:

```python

from typing import Any

from datasets import load_dataset
from huggingface_hub import HfApi
from kedro.io import AbstractDataset

from ._base import DatasetLike

**Review by ElenaKhaustova — COMMENTED** [review]
Thanks @iwhalen, this looks great now! All the feedback from the previous round has been addressed cleanly:

- `__init__` is now free of filesystem calls and validation logic, matching the convention used elsewhere in `kedro-datasets`
- `_save_dataset_dict` now validates that `data_files` keys match the `DatasetDict` split names — much better than a silent "file not found" at load time
- `RuntimeError` → `DatasetError` for consistency
- `DatasetLike` is now defined once in `_base.py`
- The Arrow docstring (and the others) correctly describe the iterable behavior now

The CI failure is actually from this PR — coverage dropped to 99.88% and the project requires 100%. The uncovered lines are:

- `_base.py:199-200` — the `except DatasetError: return False` branch in `_exists()`
- `arrow_dataset.py:78-79` — the same branch in Arrow's `_exists()` override

This branch fires when `_get_load_path()` can't resolve a path — i.e. a versioned dataset with no saved versions yet. Adding a test like this to both `test_filesystem_datasets.py` and `test_arrow_dataset.py` should cover it:

```python
def test_exists_no_versions(self, kedro_dataset_cls, path_file):
    """`exists()` returns False (not raises) when no versions are saved yet."""
    ds = kedro_dataset_cls(path=path_file, version=Version(None, None))
    assert ds.exists() is False

File: kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py (hunk 21 above)
Comment (iwhalen) [reply kedro-datasets/kedro_datasets/huggingface/hugging_face_dataset.py]: I'm surprised my linter didn't catch this! Thanks.

**Comment by iwhalen** [comment]
@ElenaKhaustova Fixed some more issues, including a weird one with chromadb dependencies on python 3.14.

It seems like this was known: https://github.com/chroma-core/chroma/issues/2571

(Failure was [here](https://github.com/kedro-org/kedro-plugins/actions/runs/25700176899/job/75458249064)).

**Comment by ElenaKhaustova** [comment]
@iwhalen, thank you for addressing the remaining comments!

It looks like the failure isn't from your HF changes. All `unit-tests` jobs (where the HF datasets live) are passing on every platform/Python combination, including Windows + 3.14.


The only failing job is `unit-tests-experimental (windows-latest, 3.14)`, which runs the tests under `kedro_datasets_experimental/`. Looking at the PR diff, your branch includes two unrelated files that look like they came in via a rebase:
- `kedro_datasets_experimental/opik/opik_evaluation_dataset.py`
- `kedro_datasets_experimental/tests/opik/test_opik_evaluation_dataset.py`


That Opik test suite is failing specifically on Windows + Python 3.14 (it passes on every other matrix entry). That's a pre-existing `main`-branch issue, not something introduced here.

Could you rebase fresh on `main` to drop those unrelated Opik changes from your diff? And remove all changes unrelated to your PR. If the Windows + 3.14 experimental job still fails after that, we'll look into it separately on our side in parallel.

**Comment by ankatiyar** [comment]
The failures on experimental datasets will not block merging PRs btw, so no worries about that.

**Comment by iwhalen** [comment]
@ElenaKhaustova Opik changes removed! Not sure how they got in there...

Thanks again for all the help.

File: kedro-datasets/pyproject.toml [hunk 22]
```diff
@@ -210,7 +210,10 @@ yaml-yamldataset = ["kedro-datasets[pandas-base]", "PyYAML>=4.2, <7.0"]
 yaml = ["kedro-datasets[yaml-yamldataset]"]
 
 # Experimental Datasets
-chromadb-chromadbdataset = ["chromadb>=1.0.0"]
+chromadb-chromadbdataset = [
```
ElenaKhaustova: A few small things here:

1. **The new HF datasets are missing per-dataset extras.** Right now you've left the `huggingface-*` block unchanged, so `ArrowDataset`/`CSVDataset`/`JSONDataset`/`ParquetDataset` have no installable extras of their own. CI still passes because the existing `huggingface-hfdataset` extra already declares `datasets`, but this breaks the convention used by every other dataset family in this file (compare `pandas-csvdataset`, `pandas-parquetdataset`, etc.).

Could you add:
```yaml
huggingface-arrowdataset = ["datasets"]
huggingface-csvdataset = ["datasets"]
huggingface-jsondataset = ["datasets"]
huggingface-parquetdataset = ["datasets"]
```
And update the `huggingface = [...]` extra to include them? That way users can do `pip install "kedro-datasets[huggingface-csvdataset]"` and pull in only what they need, without dragging in `huggingface_hub` (which is only required for the Hub-backed `HFDataset`).

2. **The `opentelemetry-exporter-otlp-proto-grpc` additions are a `chromadb` fix for Python 3.14**, unrelated to this PR's HF work. Could you please revert them? It'll be much faster for us to do this in a separate PR.

kedro-datasets/kedro_datasets/huggingface/_base.py
```diff
@@ -0,0 +1,210 @@
...
+                f"{type(self).__name__} only supports `datasets.Dataset`, "
+                "`datasets.DatasetDict`, "
+                f"Got {type(data)}"
+            )
+            raise DatasetError(msg)
+
+        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+
+        if isinstance(data, DatasetDict):
+            self._save_dataset_dict(data, save_path)
```
ankatiyar, line 136: QQ: Do you need a `data_files` dict here to specify filepaths?

ankatiyar: Sorry I see below it's being read from `_load_args`. Should this be read from `save_args` instead?

iwhalen: This is a really good point actually... it should be read from `save_args`.

However, the way its implemented right now would require a user to specify `data_files` in both the load and save args.

Concretely, it would have to look like this in yaml:


``` yaml
reviews:
  type: huggingface.CSVDataset
  path: data/01_raw/reviews
  load_args:
    data_files:
      labels: labels.csv
      data: data.csv
  save_args:
    data_files:
      labels: labels.csv
      data: data.csv
```

Which isn't the most user friendly... hmm...

I guess there are two options: swap the saving operation to read from `save_args` and require it to be specified twice or allow `data_files` as a top-level argument to the dataset. 

In this second option,

``` yaml
reviews:
  type: huggingface.CSVDataset
  path: data/01_raw/reviews
  data_files:
    labels: labels.csv
    data: data.csv
```

would work and so would specifying `data_files` in both the save and load args, but something like:

``` yaml
reviews:
  type: huggingface.CSVDataset
  path: data/01_raw/reviews
  data_files:
    labels: labels.csv
    data: data.csv
  save_args:
    data_files:
      labels: labels.csv
      data: data.csv
```

would throw an error (regardless of if filenames match or not). 

What do you think @ankatiyar? Or maybe there's another option you'd prefer.

ankatiyar: If this information is used for both saving and loading, I don't mind if it's a top level key - you're right having to specify it for both load and save args might be tedious but maybe templating (with `OmegaConfigLoader`) could also help for that option and this approach can be documented -
```
_data_files:
  labels: labels.csv
  data: data.csv

reviews:
  type: huggingface.CSVDataset
  path: data/01_raw/reviews
  load_args:
    data_files: ${_data_files}
  save_args:
    data_files: ${_data_files}
```
I'm okay with either approach, just using `load_args` for saving stuff feels not quite right. WDYT? @ElenaKhaustova

ElenaKhaustova: +1 to @ankatiyar that using `load_args` for save behaviour feels off. I'd go with **Option B (top-level `data_files`)** rather than duplicating it across `load_args`/`save_args`:

- `data_files` describes the directory layout, which both load and save need to agree on — same shape as `path`, which is also top-level.
- Single source of truth, no risk of `load_args` and `save_args` drifting apart.
- Surfaces it in the constructor signature instead of hiding it inside an opaque `load_args` dict.
- OmegaConf templating works, but relies on users knowing the trick and still produces visually duplicated YAML.

While you're in this code, there's also a related bug worth fixing: `_save_dataset_dict` currently *ignores* the filename values in `data_files` on save — it synthesises `f"{save_path}/{split}{ext}"` regardless. So:

```yaml
load_args:
  data_files:
    labels: my_labels.csv    # silently ignored on save
    data:   my_data.csv
```

…will save to `labels.csv` / `data.csv` but then load looks for `my_labels.csv` / `my_data.csv` → round-trip fails. The current validation only checks keys, not values, and the test fixtures all use `{split}{ext}` filenames so this never gets exercised.

So my suggestion:
1. Promote `data_files` to a top-level constructor arg.
2. Raise a clear error if it also appears in `load_args` or `save_args`.
3. Make the save loop use `f"{save_path}/{data_files[split]}"` so filenames round-trip.
4. Add a test with a custom filename (e.g. `{"train": "my_train.csv"}`) to lock the behaviour in.

iwhalen: Great! Agreed. I'll do this.

iwhalen:
@ElenaKhaustova @ankatiyar thanks again for all the feedback!

There's a seemingly unrelated segfault happening now, but the changes are up.

The only place I had to make an arbitrary decision was in the case that `load_args["data_files"] != save_args["data_files"]`. In this case, I just do nothing and let the user save and load to different files if they want to. 

Its a weird edge case, but I think we want people to use the top level `data_files` argument anyway 🤷

kedro-datasets/kedro_datasets/huggingface/__init__.py
```diff
@@ -11,6 +11,26 @@
     # https://github.com/pylint-dev/pylint/issues/4300#issuecomment-1043601901
```
deepyaman, line 11: I know this is dumb, but can we include this in each block, in the spirit of blind consistency with how other datasets are. 😅

kedro-datasets/kedro_datasets/huggingface/parquet_dataset.py
```diff
@@ -0,0 +1,116 @@
...
+        ...         "labels": "labels.parquet",
+        ...         "data": "data.parquet",
+        ...     },
+        ... )
+        >>> dataset.save(dataset_dict)
+        >>> loaded = dataset.load()
+    """
+
+    BUILDER: ClassVar[str] = "parquet"
+    EXTENSION: ClassVar[str] = ".parquet"
```
deepyaman, line 116: Just to confirm, this extension is only used if you don't specify `save_data_files`? Does this also work for loading e.g. `.parquet.snappy` or some other sorts of extensions?

I think so, but want to be sure.

deepyaman (approved):
I just took a quick pass, but LGTM overall; one nit comment, and then agree with some feedback you already are going to address.

ankatiyar:
@iwhalen how about we only allow the top level `data_files` and not in `load_args` or `save_args`. That way it's less ambiguous behaviour. Otherwise the PR LGTM! Since this is a core dataset contribution, I'll get the TSC to start voting on it 👍🏾

merelcht (approved):

kedro-datasets/kedro_datasets/huggingface/parquet_dataset.py
iwhalen: Yep! This works fine. 

Example code if you want to try it:

``` python
from kedro_datasets.huggingface import ParquetDataset
from datasets import Dataset, DatasetDict

dataset = Dataset.from_dict({"a": [1,2]})
dataset_dict = DatasetDict({"data": dataset, "labels": dataset})
ds_snappy = ParquetDataset(
    path="data_snappy/", data_files={"data": "data.parquet.snappy", "labels": "labels.parquet.snappy"}
)
ds_snappy.save(dataset_dict)
ds_snappy.load()
```

Will save and load to the current working directory and have the expected files:

```
data_snappy/
    data.parquet.snappy
    labels.parquet.snappy
```

iwhalen:
> @iwhalen how about we only allow the top level `data_files` and not in `load_args` or `save_args`. That way it's less ambiguous behaviour. Otherwise the PR LGTM! Since this is a core dataset contribution, I'll get the TSC to start voting on it 👍🏾

This new change is up. A `DatasetError` will be thrown now if `data_files` is in the load or save args.

ankatiyar (approved):
Thank you @iwhalen! 💯

ElenaKhaustova (approved):
Thank you, @iwhalen 🚀🚀🚀

rashidakanchwala (commented):
thank you @iwhalen !

rashidakanchwala (approved):

yetudada (approved):

SajidAlamQB (approved):

iwhalen:
Thanks all!

lrcouto (approved):

Galileo-Galilei (approved):

ravi-kumar-pilla (approved):

jitu5 (approved):

Issue #1378, opened 2026-04-11
ibis.FileDataset fails to load a saved parquet file


jstammers:
## Description
I've encountered an error when trying to load a parquet file using the `polars` backend for an `ibis.FileDataset` object. This appears to be due to the fact that the `PurePosixPath` file path is not correctly handled by `ibis.polars.backends.read_parquet`

## Context
I am unable to use this class to correctly save and load parquet files using the polars backend without using the following patch to convert the file paths to strings

```python
class PosixSafeFileDataset(FileDataset):
    """FileDataset variant that coerces resolved paths to strings.

    The upstream ``kedro_datasets.ibis.FileDataset`` currently passes a
    ``PurePosixPath`` into Ibis backend file readers. The Polars backend's
    ``read_parquet`` path handling does not accept that type reliably, so we
    normalize both load and save paths to plain strings before delegating to the
    backend.
    """

    def load(self) -> ir.Table:
        load_path = str(self._get_load_path())
        reader = getattr(self.connection, f"read_{self._file_format}")
        return reader(load_path, table_name=self._table_name, **self._load_args)

    def save(self, data: ir.Table) -> None:
        save_path = str(self._get_save_path())
        Path(save_path).parent.mkdir(parents=True, exist_ok=True)
        writer = getattr(self.connection, f"to_{self._file_format}")
        writer(data, save_path, **self._save_args)
```

## Steps to Reproduce
The following script reproduces the issue
```python
"""Minimal repro for kedro_datasets.ibis.FileDataset parquet round-trip."""

from __future__ import annotations

import traceback
from pathlib import Path
from tempfile import TemporaryDirectory

import ibis
import pandas as pd
from kedro_datasets.ibis import FileDataset


def main() -> None:
    backend = ibis.polars.connect()
    source_frame = pd.DataFrame(
        {
            "id": [1, 2],
            "a": ["foo", "bar"],
        }
    )
    source_table = backend.read_pandas(source_frame, table_name="source_table")

    with TemporaryDirectory() as tmp_dir:
        parquet_path = Path(tmp_dir) / "output.parquet"
        dataset = FileDataset(
            filepath=str(parquet_path),
            file_format="parquet",
            connection={"backend": "polars"},
            table_name="output",
        )

        print(f"dataset filepath type: {type(dataset._filepath)!r}")
        print(f"dataset filepath value: {dataset._filepath!r}")

        dataset.save(source_table)
[...]

PR #1380, opened 2026-04-17
fix(datasets): Explicitly convert paths to str in ibis.FileDataset

lrcouto:
## Description

Fix for https://github.com/kedro-org/kedro-plugins/issues/1378.

`ibis.FileDataset` passes PurePosixPath objects directly to ibis backend reader/writer methods. The Polars backend's `read_parquet` internally calls `len()` on the path argument (to distinguish a single path from a list of paths), which fails because PurePosixPath has no such method:

```py
kedro.io.core.DatasetError: Failed while loading data from dataset kedro_datasets.ibis.file_dataset.FileDataset(filepath=PurePosixPath('/tmp/tmpwkw44st1/output.parquet'), file_format='parquet', table_name='output', backend='polars', load_args={}, save_args={}).
object of type 'PurePosixPath' has no len()
```

Explicitly converted load and save paths to str before passing them to the backend. This is the lowest-common-denominator type accepted by all ibis backends and aligns with the workaround suggested by the community member who reported the issue.

## Checklist


deepyaman (dismissed):
~I don't think this is the right fix.~ I'll comment on the original issue.

**Update**: It's too harsh to say this isn't the right fix; it's probably a valid fix given the current issues with `ibis.FileDataset`. I still think converting to `Path` instead of `str` should also work and be more aligned with the existing code, but really anything is a stopgap until resolve #1298.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -154,15 +154,15 @@ def connection(self) -> BaseBackend:
         return self._connection
 
     def load(self) -> ir.Table:
-        load_path = self._get_load_path()
+        load_path = str(self._get_load_path())
```
deepyaman, line 157:
```suggestion
        load_path = Path(self._get_load_path())
```

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -154,15 +154,15 @@ def connection(self) -> BaseBackend:
...
         reader = getattr(self.connection, f"read_{self._file_format}")
         return reader(load_path, table_name=self._table_name, **self._load_args)
 
     def save(self, data: ir.Table) -> None:
         save_path = self._get_save_path()
         Path(save_path).parent.mkdir(parents=True, exist_ok=True)
         writer = getattr(self.connection, f"to_{self._file_format}")
-        writer(data, save_path, **self._save_args)
+        writer(data, str(save_path), **self._save_args)
```
deepyaman, line 165:
```suggestion
        writer(data, save_path, **self._save_args)
```

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -154,15 +154,15 @@ def connection(self) -> BaseBackend:
...
 
     def load(self) -> ir.Table:
-        load_path = self._get_load_path()
+        load_path = str(self._get_load_path())
         reader = getattr(self.connection, f"read_{self._file_format}")
         return reader(load_path, table_name=self._table_name, **self._load_args)
 
     def save(self, data: ir.Table) -> None:
         save_path = self._get_save_path()
         Path(save_path).parent.mkdir(parents=True, exist_ok=True)
```
deepyaman, line 163:
```suggestion
        save_path = Path(self._get_save_path())
        save_path.parent.mkdir(parents=True, exist_ok=True)
```
For a temporary workaround, maybe the set of proposed suggestions is also fine?

deepyaman (commented):
Updated suggestions, with my updated understanding.

P.S. Can we also add a test that covers this?

lrcouto:
`Path` does not work here, it crashes in the same way as it did before since it has no `__len__` method.

I'll add an unit test for the current state.

ankatiyar (approved):

ravi-kumar-pilla (approved):

PR #1393, opened 2026-05-05
fix(datasets): Add path validation to IncrementalDataset

lrcouto:
## Description

`IncrementalDataset._parse_checkpoint_config()` accepted any user-supplied `checkpoint.filepath` without validation. Via Kedro's runtime parameter interpolation (`${runtime_params:...}`), it would be possible to point the checkpoint to a file outside the dataset directory and `confirm()` would write there.

With the changes made on this PR, this is now validated against the dataset's base directory in two modes:
- Absolute paths (e.g. /tmp/file.txt): checked directly, must start with the dataset directory
- Relative paths (e.g. ../../etc/passwd): delegated to the existing validate_sub_path() utility, which normalizes `..` components and rejects escapes

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
+++ b/kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
@@ -9,6 +9,7 @@ new partitions past the checkpoint.It also uses `fsspec` for filesystem level op
 from __future__ import annotations
 
 import operator
+import posixpath
 from collections.abc import Callable
 from copy import deepcopy
 from typing import Any
@@ -22,6 +23,8 @@ from kedro.io.core import (
     parse_dataset_definition,
 )
 from kedro.utils import load_obj
+
+from kedro_datasets._utils import validate_sub_path
 
 from .partitioned_dataset import (
     KEY_PROPAGATION_WARNING,
@@ -175,7 +178,32 @@ class IncrementalDataset(PartitionedDataset):
                 {"keys": CREDENTIALS_KEY, "target": "checkpoint"},
             )
 
-        return {**default_config, **checkpoint_config}
+        merged = {**default_config, **checkpoint_config}
+
+        if self._filepath_arg in checkpoint_config:
+            user_filepath = str(checkpoint_config[self._filepath_arg])
+            dir_path = self._filesystem._strip_protocol(self._normalized_path).rstrip(
+                self._sep
+            )
+            stripped = self._filesystem._strip_protocol(user_filepath).replace(
+                "\\", "/"
+            )
+            if posixpath.isabs(stripped):
+                normalized = posixpath.normpath(stripped)
+                normalized_base = posixpath.normpath(dir_path)
+                if not (
+                    normalized == normalized_base
+                    or normalized.startswith(normalized_base + "/")
+                ):
+                    raise DatasetError(
+                        f"Checkpoint filepath '{user_filepath}' resolves to "
+                        f"'{normalized}' which is outside the dataset "
+                        f"directory '{dir_path}'."
+                    )
+            else:
+                validate_sub_path(stripped.lstrip("/"), dir_path)
+
+        return merged
 
     def _list_partitions(self) -> list[str]:
         if self._cached_partitions is not None:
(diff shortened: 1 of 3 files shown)


kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
```diff
@@ -9,6 +9,8 @@
 from __future__ import annotations
 
 import operator
+import os
+import posixpath
```
merelcht, line 13: Looks like this is unused when I check out the code.

kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
```diff
@@ -175,7 +179,38 @@ def _parse_checkpoint_config(
...
+            else:
+                # Cloud filesystem (S3, GCS, Azure, …): strip protocol and
+                # use validate_sub_path which handles cloud path traversal.
+                dir_path = self._filesystem._strip_protocol(
+                    self._normalized_path
+                ).rstrip(self._sep)
+                fp_stripped = self._filesystem._strip_protocol(user_filepath).replace(
+                    "\\", "/"
+                )
+                validate_sub_path(fp_stripped.lstrip("/"), dir_path)
```
merelcht: I used my review agent and it flagged this:

Cloud branch argument semantics are incorrect. `_strip_protocol` on an S3 URL like `s3://bucket/dir/../../../bad` returns `bucket/dir/../../../bad`. After `.lstrip("/")` this is still a bucket-prefixed absolute-style path, not a path relative to
    dir_path. When passed as the first argument to `validate_sub_path`, traversal via a same-bucket `..` attack may be silently allowed because the normalised result can still share the `dir_path` prefix. A cross-bucket attack (s3://other-bucket/evil → other-bucket/evil) is also not obviously handled. 

Simplest correct fix: strip protocol from both paths, then do a direct `posixpath.normpath` + prefix check — the same pattern as the local branch — and raise DatasetError directly, removing the `validate_sub_path` delegation for the cloud path.

Tests that verify this indeed isn't handled:

```python
def test_s3_same_bucket_traversal_blocked(self, mocked_csvs_in_s3):
        traversal_path = f"s3://{BUCKET_NAME}/csvs/../outside"
        with pytest.raises(DatasetError, match="outside the dataset directory"):
            IncrementalDataset(
                path=mocked_csvs_in_s3,
                dataset=DATASET,
                checkpoint={"filepath": traversal_path},
            )

    def test_s3_cross_bucket_blocked(self, mocked_csvs_in_s3):
        cross_bucket_path = "s3://other-bucket/evil"
        with pytest.raises(DatasetError, match="outside the dataset directory"):
            IncrementalDataset(
                path=mocked_csvs_in_s3,
                dataset=DATASET,
                checkpoint={"filepath": cross_bucket_path},
            )
```

jitu5: There is a similar issue for local relative paths:
```
path: data/01_raw/events
checkpoint:
  filepath: conf/base/catalog.yaml
```
The validator treats this as `data/01_raw/events/conf/base/catalog.yaml`, but the merged checkpoint config still has the raw `conf/base/catalog.yaml`, so `TextDataset` writes to `<cwd>/conf/base/catalog.yaml`. In a normal Kedro run <cwd> is the project root, so this overwrites the project's own `conf/base/catalog.yaml`, which is the same end state as the original issue.

kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
lrcouto: Oh, that's a leftover =X

kedro-datasets/kedro_datasets/partitions/incremental_dataset.py
lrcouto: Thank you, I'll address it this afternoon.

lrcouto: Applied the suggested fix, and added the tests as well.

merelcht (approved):
Thanks @lrcouto ! I checked it manually with the case that Jitendra mentioned and it's working as expected now 👍

jitu5 (approved):

Issue #1285, opened 2026-01-13
Migrate kedro-airflow DAG template to Airflow 3.x TaskFlow API


DimedS:
The `kedro-airflow` plugin still generates DAGs in the **Airflow 1.x style** (custom operators, `with DAG(...)` blocks, legacy imports such as `apply_defaults`).

Airflow introduced the decorator-based **TaskFlow API** in **Airflow2.0**. We discussed migrating to this approach in #25 but decided not to proceed at the time.

With the release of **Airflow 3.x**, the current template has now become incompatible, and the decorator-based approach has clearly become the standard for all new Airflow DAGs.

At the same time, we want all Kedro plugins to support **Python 3.13**(#940). That requires `kedro-airflow` to be compatible with **Airflow 3.1+**, which means the existing DAG template must be modernised.

### Goal

Revise the DAG template generated by `kedro-airflow` so that it aligns with the standards introduced in **Airflow 2.x and 3.x**.

At minimum:

-   Use **TaskFlow decorators** (`@dag`, `@task`) instead of custom operators and legacy DAG context managers.
-   Remove all deprecated or removed APIs (for example `apply_defaults`, legacy imports, `schedule_interval`).
-   Adopt the **Airflow SDK import paths** (`airflow.sdk`) for DAG authoring
-   Ensure the generated DAG parses and runs correctly on **Airflow 3.1+  / Python 3.13**.

Issue #25, opened 2022-04-29
[spike] Investigate support for TaskFlowAPI in kedro-airflow


noklam:
## Description
Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

Since Airflow 2.0, a simpler TaskFlowAPI for DAGs is released as an alternative with the `Operator` API. At the moment `kedro-airflow` supports the `Operator`, but it's good to keep an eye on it.
https://airflow.apache.org/docs/apache-airflow/stable/tutorial_taskflow_api.html

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

astrojuanlu:
Apparently both approaches can coexist, so TaskFlow doesn't aim to be an alternative https://www.astronomer.io/blog/apache-airflow-taskflow-api-vs-traditional-operators/

astrojuanlu:
Outcome: Example of how this API would look like, plus some investigation on what's the sentiment of the community towards this new API (with respect to more modern Python operators, such as the virtual env operator, name TBC).

DimedS:
I attempted to update the spaceflights-pandas DAG from its current form to the TaskFlow API, as seen in this commit: https://github.com/DimedS/kedro-taskFlowAPI/commit/98b69f575a80332bc67868f8d1796346a90860e9
It works, but the result looks not so beautiful. I believe we should redesign some aspects. With the TaskFlow API, it seems we no longer need the KedroOperator, and we should also improve something in the way how we describe the execution order with TaskFlowAPI.

ankatiyar:
# TaskFlow API
[TaskFlow API](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html) was introduced in Apache Airflow v2 and onwards. It's an alternative way to define your DAG files for orchestration by Airflow.

## Basic Structure
<details>
<summary> Example TaskFlow DAG </summary>

```python
from airflow.decorators import dag, task

@dag(
    schedule=None,
    start_date=pendulum.datetime(2021, 1, 1, tz="UTC"),
    catchup=False,
    tags=["example"],
)
def tutorial_taskflow_api():
    """
    ### TaskFlow API Tutorial Documentation
    This is a simple data pipeline example which demonstrates the use of
[...]

PR #1395, opened 2026-05-07
feat(airflow): Migrate kedro-airflow DAG template to Airflow 3.x

lrcouto:
## Description

https://github.com/kedro-org/kedro-plugins/issues/1285

Migrates the default Airflow DAG template from the Airflow 1.x/2.x BaseOperator pattern to the Airflow 3.x TaskFlow API (https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html).

**This is a breaking change. Airflow 2.x is no longer supported.**

What changed:

  - Template rewritten to use `@dag` / `@task` decorators from airflow.sdk — no more KedroOperator subclass
  - schedule_interval config key in airflow.yml is preserved for backwards compatibility; it maps to the schedule
  parameter in the generated Python
  - Airflow test dependency updated from <3.0 to >=3.2.0,<4.0 (3.2.0 is the minimum release with Python 3.14
  support)
  - Python 3.13 and 3.14 added to the CI test matrix ~(3.14 pending Kedro's 3.14 release)~ (Kedro was released now, so this also adds support to Python 3.14).
  - Black bumped from 22.x to 24.x: ast.Str was removed in Python 3.13. Black 22.x uses it internally, which crashes
   local linting (AttributeError: module 'ast' has no attribute 'Str') on any Python 3.13+ dev environment. Black
  24.1.0 was the release that added Python 3.13 support.
  - New test for namespace grouping that verifies that session.run(pipeline_name, namespaces=["..."]) is rendered for
  namespace-typed groups
  - README updated with an Airflow 2.x to 3.x migration guide

Users still on Airflow 2.x should pin to the previous kedro-airflow release. Custom templates that subclass or reference KedroOperator will need to be rewritten using @task-decorated closures inside a `@dag` function.

## Checklist


lrcouto:
Note that this should also work for Python 3.14, but we should wait for the next Kedro release to bump the version so CI doesn't fail because of version compatibility warnings.

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,45 @@ package_name = "{{ package_name }}"
...
 
-# Using a DAG context manager, you don't have to specify the dag property of each task
-with DAG(
+@dag(
     dag_id="{{ dag_name | safe | slugify }}",
     start_date=datetime({{ start_date | default([2023, 1, 1]) | join(",")}}),
     max_active_runs={{ max_active_runs | default(3) }},
-    # https://airflow.apache.org/docs/stable/scheduler.html#dag-runs
-    schedule_interval="{{ schedule_interval | default('@once') }}",
+    schedule="{{ schedule_interval | default('@once') }}",
```
ankatiyar: If we're making breaking changes, might as well get rid of this backwards compatibility and make the variable `schedule` OR support both

kedro-airflow/pyproject.toml
ankatiyar, line 14: Need to update dependency here - I believe taskflow syntax came with airflow 2.0 so should work with airlfow>=2.0.0 but please check once

kedro-airflow/RELEASE.md
```diff
@@ -1,4 +1,7 @@
 # Upcoming Release
+* Migrated default DAG template to Airflow 3.x TaskFlow API (`@dag`/`@task` from `airflow.sdk`).
+* Dropped support for Airflow 2.x. Minimum test dependency is now `apache-airflow>=3.2.0`.
```
ankatiyar: Again I think Taskflow was added in 2.x so this might not be accurate, worth verifying and testing - maybe we're just dropping support for 1.x 
Also no need to mention the minimum test dependency

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -3,50 +3,11 @@ from __future__ import annotations
...
-                session.run(self.pipeline_name, namespaces=[self.namespace])
-            else:
-                # For memory grouping or individual nodes, use node_names
-                if isinstance(self.node_name, str):
-                    self.node_name = [self.node_name]
-                session.run(self.pipeline_name, node_names=self.node_name)
-
 # Kedro settings required to run your pipeline
 env = "{{ env }}"
 pipeline_name = "{{ pipeline_name }}"
```
ankatiyar, line 13: Also a good time to move from `pipeline_name`(deprecated) to `pipeline_names`. If this is a breaking release we can just support `pipeline_names` but `session` still works with `pipeline_name` too so can support both as well

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,45 @@ package_name = "{{ package_name }}"
...
+{%- for group in node_objs %}
+
+    @task(task_id="{{ group.name | slugify }}")
+    def {{ group.name | slugify | replace("-", "_") }}():
+        configure_project(package_name)
+        with KedroSession.create(project_path, env=env, conf_source=conf_source) as session:
+{%- if group.type == "namespace" %}
+            session.run(pipeline_name, namespaces=["{{ group.name }}"])
+{%- else %}
+            session.run(pipeline_name, node_names=[{% for node in group.nodes %}"{{ node }}"{% if not loop.last %}, {% endif %}{% endfor %}])
```
ankatiyar: re: above comment about `pipeline_name`/`pipeline_names`. Would be better to specify the argument name `pipeline_name=` just so this doesn't stop functioning when we drop the deprecated one

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,45 @@ package_name = "{{ package_name }}"
...
-) as dag:
+    ),
+)
+def {{ dag_name | safe | slugify | replace("-", "_") }}():
+{%- for group in node_objs %}
+
+    @task(task_id="{{ group.name | slugify }}")
+    def {{ group.name | slugify | replace("-", "_") }}():
+        configure_project(package_name)
+        with KedroSession.create(project_path, env=env, conf_source=conf_source) as session:
```
ankatiyar: Suggest doing `project_path=project_path` within create as well - will help with making `session` interchangeable in the future as `KedroServiceSession` as slightly different args

kedro-airflow/kedro_airflow/airflow_dag_template.j2
ankatiyar: Actually this bit - configure_project + session create + session run can be abstracted away in a function and then just parametrise the call to the function - https://github.com/kedro-org/kedro-plugins/issues/25#issuecomment-2107299300 for inspiration

ankatiyar (commented):
Left some comments, over all looking good! I want to see if we can support custom `SESSION_CLASS` via settings same framework but I think that can be a follow up thing. 
I did some research on the TaskFlow API before (https://github.com/kedro-org/kedro-plugins/issues/25#issuecomment-2107299300)
I would like to revisit our airflow deployment strategy guidance once the service session work is complete - the new session with data injection should allow for not having to save intermediate datasets etc Let me know if you have any thoughts based on the earlier research and current work (again out of scope for this PR, just worth noting)

kedro-airflow/pyproject.toml
ankatiyar: SOrry this comment is irrelevant, i misread kedro for airflow

ankatiyar (commented):
Tested this new DAG template and looks better! Some suggestions:
- use `pipeline_names` which passes a list of names to align with `--pipelines` in session.
- Deprecate/remove `--pipeline` option from the plugin.
- Still check if airflow 2.x version is compatible or not

ankatiyar:
Maybe the `--pipeline` vs `--pipelines` change can be a separate PR before or after this one. But before we do a breaking release for sure.

lrcouto:
> Maybe the `--pipeline` vs `--pipelines` change can be a separate PR before or after this one. But before we do a breaking release for sure.

We're gonna need to do another PR anyway to bump the CI to run Python 3.14 before release, it's just waiting for Kedro to go first. But might as well do it here since that --pipelines change is already live. I'm gonna be working on the pipeline selection task this sprint, so I can take a look at this again once that one is done.

And yeah Airflow 2.x will not be compatible anymore.

kedro-airflow/README.md
```diff
@@ -194,6 +194,48 @@ For more information about namespaces in Kedro, see the [namespaces documentatio
 It is possible to use [Airflow task groups](https://docs.astronomer.io/learn/task-groups) by changing the template.
 See ["What if I want to use a different Jinja2 template?"](#what-if-i-want-to-use-a-different-jinja2-template) for instructions on using custom templates.
 
+## Migrating from Airflow 2.x
+
+`kedro-airflow` requires **Apache Airflow 3.x** (`>=3.2.0`). The default DAG template was rewritten to use the [TaskFlow API](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html) and is not compatible with Airflow 2.x.
```
ravi-kumar-pilla: [nit] should we mention the kedro-airflow version which requires airflow >3 ? or mention which kedro-airflow version supports airflow <3 ?

ravi-kumar-pilla:
Hi @lrcouto ,

Overall the code looks good and I left a minor comment. Would you mind sharing a quick test snippet, so I can replicate on my end to test this e2e. I will do an e2e test tomorrow. Thank you

lrcouto:
Thanks for the review @ravi-kumar-pilla ! I've added a couple extra tests to check for the airflow.sdk import. You can also check that it only runs on Airflow>=3.x by running this:

```py

from pathlib import Path
import importlib.metadata

airflow_version = importlib.metadata.version("apache-airflow")

print(f"Airflow {airflow_version} — OK")

import jinja2
from slugify import slugify
from kedro.pipeline import GroupedNodes

jinja_file = Path(__file__).parent.parent / "kedro_airflow" / "airflow_dag_template.j2"
loader = jinja2.FileSystemLoader(jinja_file.parent)
jinja_env = jinja2.Environment(autoescape=True, loader=loader, lstrip_blocks=True)
jinja_env.filters["slugify"] = slugify

node_objs = [
    GroupedNodes(name="split", type="nodes", nodes=["split_data"], dependencies=[]),
    GroupedNodes(name="train", type="nodes", nodes=["train_model"], dependencies=["split"]),
]
dag_code = jinja_env.get_template(jinja_file.name).render(
    dag_name="smoke_test_project",
    node_objs=node_objs,
    env="local",
    pipeline_name="__default__",
    package_name="smoke_test_project",
    conf_source="",
)
print("Template rendered")

assert "from airflow.sdk import dag, task" in dag_code, \
    "FAIL: generated DAG does not import from airflow.sdk"
assert "def _run_kedro_node(" in dag_code, \
    "FAIL: _run_kedro_node helper missing from generated DAG"

try:
    compile(dag_code, "<generated_dag>", "exec")
except SyntaxError as exc:
    raise AssertionError(f"generated DAG has a syntax error: {exc}") from exc
print("Generated DAG compiles without syntax errors")

try:
    from airflow.sdk import dag, task  # noqa: F401
except ImportError as exc:
    raise AssertionError(
        f"airflow.sdk not available — Airflow >=3.x required. {exc}"
    ) from exc
print("airflow.sdk importable — Airflow 3.x confirmed")

print("\nAll checks passed.")
```

It should fail on a version under 3, but work on 3 and over.

ravi-kumar-pilla (approved):
Works well. Thank you @lrcouto

ankatiyar (commented):
The DAG generated looks good but `--pipelines` only accepts one pipeline at a time. For example, `kedro airflow create --pipelines=data_processing,data_science` wouldn't work. The pipeline argument here in this plugin uses this to check which files to convert, so I think we can split the input of the pipelines then create individual dags for these pipelines.

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -26,7 +26,7 @@
 
 PIPELINE_ARG_HELP = """Name of the registered pipeline to convert.
 If not set, the '__default__' pipeline is used. This argument supports
-passing multiple values using `--pipeline [p1] --pipeline [p2]`.
```
ankatiyar, line 29: Accept a list for piplines instead of allowing multiple values like for Kedro repo

kedro-airflow/kedro_airflow/plugin.py
```diff
@@ -87,7 +87,6 @@ def _get_pipeline_config(config_airflow: dict, params: dict, pipeline_name: str)
 @airflow_commands.command()
 @click.option(
     "-p",
-    "--pipeline",
     "--pipelines",
     "pipeline_names",
     multiple=True,
```
ankatiyar: Disable multiple

ankatiyar: And split the string into a list of pipelines

kedro-airflow/tests/test_plugin.py
```diff
@@ -432,15 +460,13 @@ def test_namespace_grouping_uses_namespace_parameter(cli_runner, metadata):
 
     dag_content = dag_file.read_text()
 
-    # Verify that the execute method contains namespaces parameter logic
-    assert 'if self.group_type == "namespace":' in dag_content
-    assert "session.run(self.pipeline_name, namespaces=[self.namespace])" in dag_content
-
-    # Verify that group_type parameter is being passed
-    assert "group_type=" in dag_content
+    # Verify TaskFlow @task decorator is used
```
rashidakanchwala, line 463: **Suggestion:** The test name says `uses_namespace_parameter` but it only exercises the `node_names` fallback path (the fixture pipeline has no real namespaced nodes, as the comment here explains). Consider renaming to something like `test_namespace_groupby_on_non_namespaced_pipeline` — the actual namespace branch is covered by the new `test_namespace_grouping_renders_namespaces_call`.

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,52 @@ package_name = "{{ package_name }}"
 conf_source = "{{ conf_source }}" or Path.cwd() / "conf"
 
 
-# Using a DAG context manager, you don't have to specify the dag property of each task
-with DAG(
+def _run_kedro_node(node_names=None, namespaces=None):
```
rashidakanchwala, line 19: **Suggestion:** `_run_kedro_node` has no docstring. The old `KedroOperator.execute()` had inline comments explaining the `namespace` vs `node_names` branching — worth preserving as a short docstring since this is generated code that users read and debug in production Airflow environments.

kedro-airflow/tests/test_plugin.py
```diff
@@ -457,11 +483,83 @@ def test_memory_grouping_uses_node_names_parameter(cli_runner, metadata):
...
+
+    dag_file = metadata.project_path / "airflow_dags" / "fake_project_dag.py"
+    dag_content = dag_file.read_text()
+
+    assert 'schedule="@daily"' in dag_content
+    file_name.unlink()
+    dag_file.unlink()
+
+
+def test_namespace_grouping_renders_namespaces_call():
```
rashidakanchwala, line 535: **Suggestion:** This test only covers an all-namespace pipeline (every group is `type="namespace"`). In real projects it's common to have a mixed pipeline — some nodes in a namespace, others not — which would exercise both branches of `{%- if group.type == "namespace" %}` in the same render. Adding a mixed-type `node_objs` list here (or in a separate test) would give higher confidence that both paths render correctly together.

rashidakanchwala (commented):
## Kedro PR Review
> Generated with `review-kedro-pr` skill.

### Overview
- **Critical:** 0 | **Suggestions:** 3

### Critical (must fix before merge)
No issues found.

### Suggestions (consider improving)
- [ ] `kedro-airflow/tests/test_plugin.py:L460` — Rename `test_namespace_grouping_uses_namespace_parameter`: the test only exercises the `node_names` fallback, not the namespace path. Something like `test_namespace_groupby_on_non_namespaced_pipeline` is clearer.
- [ ] `kedro-airflow/kedro_airflow/airflow_dag_template.j2:L19` — Add a short docstring to `_run_kedro_node` explaining what `node_names` vs `namespaces` each mean. Useful for anyone reading the generated DAG file in production.
- [ ] `kedro-airflow/tests/test_plugin.py:L535` — `test_namespace_grouping_renders_namespaces_call` only covers an all-namespace pipeline. Consider adding a mixed-type case (`type="namespace"` + `type="nodes"` in the same DAG) to fully exercise the template's `{% if group.type == "namespace" %}` branching.

### Notes
- **General review:** Clean, well-scoped PR. `KedroOperator` fully removed, `@dag`/`@task` adopted correctly, `schedule_interval` backwards-compat handled. The new `_load_template()` helper in tests is a nice addition enabling template unit tests independent of the CLI.
- **PR checklist:** `RELEASE.md` entry and README migration guide are both present and accurate.
- **API usage:** `session.run(pipeline_name=...)` uses the deprecated singular arg — already raised in existing comments, not repeating.

lrcouto:
Addressed the suggestion to change the test and add the docstring.

kedro-airflow/README.md
```diff
@@ -31,7 +31,7 @@ kedro airflow create
 ```
 
 This command will generate an Airflow DAG file located in the `airflow_dags/` directory in your project.
-You can pass a `--pipeline` flag to generate the DAG file for a specific Kedro pipeline and an `--env` flag to generate the DAG file for a specific Kedro environment.
+You can pass a `--pipelines` flag to generate the DAG file for a specific Kedro pipeline and an `--env` flag to generate the DAG file for a specific Kedro environment.
```
merelcht:
```suggestion
You can pass the `--pipelines` flag to generate the DAG file for a specific Kedro pipeline and the `--env` flag to generate the DAG file for a specific Kedro environment.
```

merelcht (commented):
This mostly looks good, but to @ankatiyar 's point, right now passing multiple pipelines doesn't work, it will just generate the DAG for the last mentioned pipelines when doing e.g.: `kedro airflow create --pipelines data_science,reporting`.

This is confusing especially now `pipelines` is plural.

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,53 @@ package_name = "{{ package_name }}"
...
+    """Run a Kedro pipeline slice. Pass node_names for regular node groups or namespaces for namespace groups."""
+    configure_project(package_name)
+    with KedroSession.create(project_path=project_path, env=env, conf_source=conf_source) as session:
+        if namespaces is not None:
+            session.run(pipeline_name=pipeline_name, namespaces=namespaces)
```
ankatiyar:
```suggestion
            session.run(pipeline_names=[pipeline_name], namespaces=namespaces)
```

kedro-airflow/kedro_airflow/airflow_dag_template.j2
```diff
@@ -55,42 +16,53 @@ package_name = "{{ package_name }}"
...
+        else:
+            session.run(pipeline_name=pipeline_name, node_names=node_names)
```
ankatiyar:
```suggestion
            session.run(pipeline_names=[pipeline_name], node_names=node_names)
```

ankatiyar (approved):
The DAG generated looks good. Small suggestion to avoid the deprecation warning when using `pipeline_name` in the session. 
We probably need to think about how do we want to handle multiple pipelines, generate separate DAGs or one for all pipelines listed (Now that session allows for multiple pipelines) in a separate issue.

lrcouto:
Changed the template to use pipeline_names=[pipeline_name] instead of pipeline_name=pipeline_name for now, which should avoid the deprecation warning. And yeah, we should think about handling multiple pipelines in the future, since this is a semi-recent change.

PR #1398, opened 2026-05-09
feat(datasets): add send_individually option

### LINKED ISSUE #1383: APIDataset: option to send list items as individual JSON objects instead of arrays
## Description

`APIDataset` has no way to send list items as individual JSON objects. When the input to `save()` is a list, the request body is always a JSON array — even when `chunk_size=1` the body is `[{...}]` rather than `{...}`. I'm always frustrated when integrating with REST endpoints that expect a single record per POST/PUT, because there's no catalog-level option for it and I end up subclassing the dataset.

## Context

Many REST APIs expect one JSON object per request, not a single-element array. Today, users of `APIDataset` have two workarounds, neither of which is discoverable from catalog config:

1. Subclass `APIDataset` and override `_execute_save_with_chunks` to iterate and send each element unwrapped.
2. Split the data upstream and call `save()` per record, which loses the catalog abstraction.

The source of the wrapping is in `_execute_sais a list slice(`json_data[i*chunk_size:(i+1)*chunk_size]`), so even at `chunk_size=1` the body remains an array. `requests` itself is not responsible; it faithfully serializes whaes.

Adding a first-class option would benefit anyPIs that don't accept batched arrays (CRMsystems, webhooks, logging/ingestion endpoints, many third-party SaaS APIs).

## Possible Implementation

Add a boolean `save_args` option (working name: `send_individually`). When `True` and the input is a list, each element is sent as a standalone JSON object instead o

```yaml
my_api:
  type: api.APIDataset
  method: POST
  url: https://example.com/records
  save_args:
    send_individually: true
```

In `_execute_save_with_chunks`:

```python
if self._send_individually:
    response = None
    for record in json_data:
        response = self._execute_save_request(json_data=record)
    return response
```

Default behavior is unchanged, so this is fully backwards compatible. Happy to open a PR if maintainers are on board
with the flag name and semantics.

## Possible Alternatives

- **Treat `chunk_size=1` as "unwrap"**: break anyone relying on `[{...}]` bodies with oneelement per request.
- **Accept `chunk_size=0`**: currently raisessing it would be non-obvious and easy tomisuse.

PragnyaKhandelwal:
## Description
This PR fixes #1383 by adding a `send_individually` option to `APIDataset` so list payloads can be sent as individual JSON objects instead of a single array request. The default behavior remains unchanged, so existing chunked saves continue to work unless the new option is enabled.

## Development notes
Added `send_individually` to the `APIDataset` save arguments and updated the save path to branch between chunked requests and per-item requests. Added tests covering the new behavior, including the default case, single-item input, empty lists, dict input, precedence over chunk size, and response dataset integration. Validated with `pytest` on the API dataset test suite, which passed with 80 tests.

We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a Signed-off-by line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

## Checklist

* [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
* [ ] Updated the documentation to reflect the code changes
* [ ] Updated jsonschema/kedro-catalog-X.XX.json if necessary
* [ ] Added a description of this change in the relevant RELEASE.md file
* [x] Added tests to cover my changes
* [ ] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)


PragnyaKhandelwal:
Thanks for the confirmation, @nikanton. This PR implements the `send_individually` option for `APIDataset`, keeps the default chunked behavior unchanged, and fixes #1383. Ready for review when you have time, @SajidAlamQB and @merelcht

PragnyaKhandelwal:
Hi team! It looks like this PR is ready, but the CI is currently failing on Python 3.11 due to what looks like an unrelated upstream test failure. Could someone re-trigger the workflow when you have a moment?

PragnyaKhandelwal:
Hi team! Friendly check-in on this feature. The branch is rebased with the latest main, and all checks are fully passing green. Looking forward to your feedback when bandwidth allows!

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -291,7 +299,15 @@ def load(self) -> Any:
...
         self,
         json_data: list[dict[str, Any]],
-    ) -> requests.Response:
+    ) -> requests.Response | None:
+        # If send_individually is True, send each item as a separate request
+        if self._send_individually:
+            response: requests.Response | None = None
+            for record in json_data:
+                response = self._execute_save_request(json_data=record)
+            return response
```
rashidakanchwala: **Suggestion:** When `json_data` is an empty list, `response` stays `None` and `save()` silently returns `None`. Callers that do `response.status_code` will get an `AttributeError`. Consider raising early instead:

```python
if self._send_individually:
    if not json_data:
        raise DatasetError("Cannot save an empty list with send_individually=True.")
    response = None
    for record in json_data:
        response = self._execute_save_request(json_data=record)
    return response
```

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -180,6 +187,7 @@ def __init__(  # noqa: PLR0913
             if save_args is not None:
                 self._params.update(save_args)
             self._chunk_size = self._params.pop("chunk_size", 1)
+            self._send_individually = self._params.pop("send_individually", False)
```
rashidakanchwala, line 194: **Suggestion:** `_send_individually` is only initialised inside the `PUT`/`POST` branch. It is never accessed on a GET instance so nothing breaks today, but adding `self._send_individually = False` before the `if/elif` would make the class more robust.

kedro-datasets/tests/api/test_api_dataset_send_individually.py
```diff
@@ -0,0 +1,295 @@
...
+import pytest
+import requests
+
+from kedro_datasets.api import APIDataset
+
+TEST_URL = "http://example.com/api/test"
+SAVE_METHODS = ["POST", "PUT"]
+
+
+class TestAPIDatasetSendIndividually:
```
rashidakanchwala: **Suggestion:** A few things worth tidying up here:

1. **File location** — `TestAPIDatasetResponseDataset` lives as a class inside `test_api_dataset.py`. `TestAPIDatasetSendIndividually` should follow the same pattern rather than being a standalone file.
2. **Repeated boilerplate** — the `request_history = []` + inline `json_callback` block appears verbatim in 6 of 7 tests. A class-level `@pytest.fixture` would remove the duplication without needing a `conftest.py`.
3. **Duplicated constants** — `TEST_URL` and `SAVE_METHODS` are already defined in `test_api_dataset.py` with identical values.
4. **Minor** — `import json as json_module` alias is not needed; `import json` matches the rest of the file.

rashidakanchwala (commented):
Thank you for your contribution, @PragnyaKhandelwal! This is a clean and well-scoped addition. One critical item to fix before merge, plus a few suggestions below.

## Kedro PR Review
> Generated with `review-kedro-pr` skill.

### Overview
- **Critical:** 1 | **Suggestions:** 4

### Critical (must fix before merge)
- [ ] `kedro-datasets/RELEASE.md` — no entry for `send_individually` under `# Upcoming Release`. The `## Major features and improvements` section is currently empty. Please add a bullet, e.g. `* Added send_individually option to APIDataset to send list items as individual requests instead of batched arrays.`

### Suggestions (consider improving)
- [ ] `kedro-datasets/kedro_datasets/api/api_dataset.py:308` — `save([])` with `send_individually=True` silently returns `None`; callers doing `response.status_code` will get `AttributeError`. Consider raising `DatasetError` for empty input instead.
- [ ] `kedro-datasets/kedro_datasets/api/api_dataset.py:190` — `_send_individually` is only set inside the `PUT`/`POST` branch. It is never accessed on a GET instance so it does not break today, but initialising it to `False` unconditionally would make the class safer.
- [ ] `kedro-datasets/kedro_datasets/api/api_dataset.py:154` — The `save_args` docstring still only mentions `chunk_size`. Worth adding a line about `send_individually` there too.
- [ ] `kedro-datasets/tests/api/test_api_dataset_send_individually.py` — see inline comment.

### Notes
- **General review:** Logic is correct for the non-empty list path. The only real correctness concern is the silent `None` return on empty list.
- **PR checklist:** RELEASE.md entry is missing (Critical). The class-level docstring covers `send_individually` well so no separate docs page is needed.
- **API usage:** `AbstractDataset` subclass correctly implemented. No backward compatibility issues for the normal path.

rashidakanchwala (commented):
Submitted via skill

rashidakanchwala (changes requested):

PragnyaKhandelwal:
Thanks for the thorough review and the great suggestions, @rashidakanchwala! I've added the missing `RELEASE.md` entry, implemented the `DatasetError` for empty lists, unconditionally initialized the `_send_individually` flag to False, updated the docstring, and refactored the tests into `test_api_dataset.py` using a fixture to remove the boilerplate. Ready for another look!

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -12,6 +13,7 @@
 from kedro.io.core import AbstractDataset, DatasetError, parse_dataset_definition
 from requests import Session, sessions
 from requests.auth import AuthBase
+from requests_mock import response
```
rashidakanchwala: I think this was added accidentally. Can we pls remove it?

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -291,7 +304,20 @@ def load(self) -> Any:
     def _execute_save_with_chunks(
         self,
         json_data: list[dict[str, Any]],
-    ) -> requests.Response:
+    ) -> requests.Response | None:
```
rashidakanchwala: Since the empty-list path now raises `DatasetError`,`_execute_save_with_chunks` can never actually return `None`. Three small cleanups follow from that:

Line 307: change -> requests.Response | None to -> requests.Response
Line 348: change response: requests.Response | None = ... to response: requests.Response = ...
Lines 354–355: remove the if response is None: return None guard — it's unreachable
And save() itself (line 345) can drop the | None from its return type too, since the only way it returned None was via that guard.

PragnyaKhandelwal:
Thanks for catching those, @rashidakanchwala! You're totally right. I've removed the accidental `requests_mock` import, stripped the `| None` type hints from the signatures, and deleted the unreachable `response is None` guard. Changes are pushed and all checks are green!

PragnyaKhandelwal:
Hi @rashidakanchwala, just a gentle ping! I've addressed the final type-hint cleanups and the accidental import you caught. The branch is rebased, and all CI checks are green. Could you take a quick look and re-approve when you have a moment? Thanks again for the detailed review!

kedro-datasets/kedro_datasets/api/api_dataset.py
```diff
@@ -292,14 +304,27 @@ def _execute_save_with_chunks(
...
+                raise DatasetError(
+                    "Cannot save an empty list with send_individually=True."
+                )
+
+            response = None
+            for record in json_data:
+                response = self._execute_save_request(json_data=record)
+            return response  # type: ignore[return-value]
+
+        # Otherwise, use chunked sending
```
rashidakanchwala, line 319: Can we also have a check for empty list here.


```suggestion
        # Otherwise, use chunked sending
        if not json_data:
             raise DatasetError("Cannot save an empty list.")
```

jitu5 (approved):

kedro-datasets/tests/api/test_api_dataset.py
```diff
@@ -701,3 +701,236 @@ def json_callback(request: requests.Request, context: Any) -> dict:
...
+        )
+
+        # Save list of items
+        api_dataset.save.__wrapped__(api_dataset, items)
+
+        # Verify response is from the last chunk
+        assert mock_json_echo[-1] == [items[2]]
+
+        # Verify items were sent as chunks
+        assert len(mock_json_echo) == len(items) - 1
```
rashidakanchwala:
```suggestion
        assert len(mock_json_echo) == 2
```

PragnyaKhandelwal:
Hi @rashidakanchwala, I applied both of your suggestions locally to ensure my DCO sign-off remained intact. I added the empty list check for chunked sending, updated the test assertion, and also added a new test case to ensure the `api_dataset.py` maintains its 100% test coverage. Everything is pushed and the CI is fully green. Thank you again for the thorough review!

rashidakanchwala (approved):
Amazing, thank you so much @PragnyaKhandelwal !

PR #1400, opened 2026-05-12
fix(datasets): make PolarsDatabaseDataset usable end-to-end

### LINKED ISSUE #1384: PolarsDatabaseDataset: inverted validation in __init__ makes the dataset unusable (docstring example fails)
## Description

`kedro_datasets_experimental.polars.PolarsDatabaseDataset` raises `DatasetError` on instantiation for any valid configuration, including the exact example shown in its own docstring. The validation condition in `__init__` is inverted — it raises whenever `sql` or `filepath` **is** provided, which is the opposite of the intended check, making the dataset effectively unusable.

## Context

I wanted to load data from a SQL database as a Polars DataFrame directly inside a Kedro catalog entry. `PolarsDatabaseDataset` is the natural fit, but it cannot be instantiated with any combination of arguments. After losing a significant amount of time I fell back to `pandas.SQLQueryDataset` and convert to Polars downstream, which defeats the point of using Polars natively.

## Steps to Reproduce

Reproduced in a fresh `uv` project:

```bash
mkdir polars_db_repro && cd polars_db_repro
uv init --python 3.12 .
uv add kedro kedro-datasets polars sqlalchemy
```

Save as `repro.py` (copied verbatim from the `PolarsDatabaseDataset` docstring):

```python
from pathlib import Path
import polars as pl
from kedro_datasets_experimental.polars import PolarsDatabaseDataset

data = pl.DataFrame({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
sql = "SELECT * FROM table_a"
tmp_path = Path.cwd() / "tmp"
tmp_path.mkdir(parents=True, exist_ok=True)
credentials = {"con": f"sqlite:///{tmp_path / 'test.db'}"}

dataset = PolarsDatabaseDataset(
    sql=sql,
    credentials=credentials,
    table_name="table_a",
)

dataset.save(data)
reloaded = dataset.load()
print(reloaded)
```

Then:

```bash
uv run python repro.py
```

## Expected Result

The dataset should instantiate successfully when `table_name` and one of `sql`/`filepath` are provided (as the docstring example demonstrates), so that `.save(data)` and `.load()` run without error.

## Actual Result

Instantiation always fails with a `DatasetError`, even though all required arguments are provided.

```
Traceback (most recent call last):
  File "/tmp/polars_db_repro/repro.py", line 11, in <module>
    dataset = PolarsDatabaseDataset(
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/polars_db_repro/.venv/lib/python3.12/site-packages/kedro/io/core.py", line 331, in new_init
    init_func(self, *args, **kwargs)
  File "/tmp/polars_db_repro/.venv/lib/python3.12/site-packages/kedro_datasets_experimental/polars/polars_database_dataset.py", line 160, in __init__
    raise DatasetError(
kedro.io.core.DatasetError: Either 'table_name' or one of 'sql' or 'filepath' arguments cannot both be empty.Please provide a sql query or path to a sql query file.
```

**Root cause** in `kedro_datasets_experimental/polars/polars_database_dataset.py`:

```python
if not table_name or (sql or filepath):
    raise DatasetError(
        "Either 'table_name' or one of 'sql' or 'filepath' arguments cannot both be empty."
        "Please provide a sql query or path to a sql query file."
    )
```

The condition is inverted — it raises whenever `sql` or `filepath` is provided. The intended guard is the opposite: raise when `table_name` is missing, or when both `sql` and `filepath` are missing. For example:

```python
if not table_name or not (sql or filepath):
    raise DatasetError(...)
```

The first `if sql and filepath:` check already handles mutual exclusion, so this second check only needs to enforce that `table_name` is provided together with at least one of `sql`/`filepath`.

Minor aside: the error message also has a missing space between the two sentences (`"...empty."` + `"Please..."`).

## Environment

* Kedro version used (`kedro -V`): 1.3.1
* Kedro plugin and kedro plugin version used (`pip show kedro-datasets`): kedro-datasets 9.3.0 (also affects the docstring example shipped on `main`)
* Polars / SQLAlchemy: polars 1.40.0, SQLAlchemy 2.0.49
* Python version used (`python -V`): 3.12.13
* Operating system and version: Oracle Linux Server 9.5 (kernel 5.15.0-303.171.5.2.1.el9uek.x86_64)

nikanton:
## Summary

Closes #1384.

`PolarsDatabaseDataset` (in `kedro_datasets_experimental`) was effectively broken end-to-end and needed a proper repair, not a one-line fix. The class shipped with **zero tests** — no test file existed for it on `main` — was missing basic construction validation, didn't actually support `table_name`-only usage despite advertising it, carried a copy-pasted method from the pandas dataset that would have raised `TypeError` if it ever fired, and shipped without an installable extras combination. This PR fixes the dataset, removes the dead code, and adds a real test suite from scratch.

### Fixes in `polars_database_dataset.py`
- Validate construction args: raise `DatasetError` when both `sql` and `filepath` are given, when none of `sql`/`filepath`/`table_name` is provided, or when `con` is missing/empty.
- Make `table_name`-only mode actually work: `load()` builds `SELECT * FROM <table_name>` when neither `sql` nor `filepath` is configured.
- Clarify load-source precedence in the docstring: `sql` / `filepath` (mutually exclusive) take priority over `table_name`; `save()` always writes to `table_name`.
- Remove a dead, copy-pasted-from-pandas `adapt_mssql_date_params` method. It mutated `self._load_args[\"params\"]`, which the next `load()` would unpack into `pl.read_database(..., params=...)` — but `pl.read_database` has no `params` kwarg, so it would have raised `TypeError`. SQLAlchemy already binds native `datetime` objects via `execute_options={\"parameters\": ...}`; the pandas/pyodbc string-conversion hack does not apply on the polars path.

### Tests (new file — none existed on `main`)
- New `kedro_datasets_experimental/tests/polars/test_polars_database_dataset.py` (29 tests).
- Real SQLite round-trips via a `seed_table` factory fixture and three distinct frame fixtures (`numeric_frame`, `string_frame`, `mixed_frame`) — not mocks.
- Cover construction validation, load precedence (sql / filepath / table_name), `save` modes (`replace` default, `append`), fsspec-backed query files, and `__str__` redaction of credentials.
- Cover the class-level `Engine` cache: same `con` reuses one `Engine`, different `con`s get distinct engines, and non-`con` credential keys (e.g. `pool_size`) are forwarded to `sqlalchemy.create_engine`.

### Packaging
- Add `polars>=1.0` and `SQLAlchemy>=1.4` to the `experimental` and `experimental_test` extras in `pyproject.toml` so the dataset is actually installable from the extras.
- `RELEASE.md` entry.

## Test plan
- [x] `pytest kedro_datasets_experimental/tests/polars/test_polars_database_dataset.py` — 29 passed (suite did not exist before this PR)
- [x] Verified by direct execution that `pl.read_database` rejects a `params` kwarg, confirming the removed mssql adapter was dead code
- [ ] CI green


kedro-datasets/RELEASE.md
```diff
@@ -33,12 +33,14 @@
...
 - Added `checkpoint.filepath` validation for IncrementalDataset.
+- Repaired `polars.PolarsDatabaseDataset` end-to-end (it was effectively broken and previously had no tests): corrected the inverted argument validation in `__init__` that made the dataset uninstantiable for any valid configuration; renamed the default `save_args` key from `if_exists` to `if_table_exists` for compatibility with current `polars.DataFrame.write_database`; switched `load()` to pass the cached SQLAlchemy `engine` (instead of a URI string) to `pl.read_database`; added support for table-name-only catalog entries (loads via `SELECT * FROM <table_name>` when no `sql` or `filepath` is provided); added construction-time validation that raises `DatasetError` for conflicting/missing arguments; removed a dead `adapt_mssql_date_params` method that had been copy-pasted from the pandas SQL dataset and was incompatible with `polars.read_database`; added `polars` and `SQLAlchemy` to the `experimental` and `experimental_test` extras so the dataset is installable; and added a full test suite covering load/save round-trips, source precedence, engine caching, and credentials forwarding.
```
ravi-kumar-pilla: Can we keep the release notes compact and explain the details in the PR description as we always do ?

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -273,26 +348,3 @@ def save(self, data: pl.DataFrame) -> NoReturn:
             connection=self._connection_str,
```
ravi-kumar-pilla: load() uses the cached SQLAlchemy Engine (connection=self.engine), but save() still passes connection=self._connection_str. Polars documents write_database as accepting a URI or an Engine. Using the string bypasses the class-level engine cache for writes, opens a separate client for each write path, and is risky for sqlite:///:memory: (each new connection is a different empty database). Consider connection=self.engine in write_database for parity with read_database, pooling, and in-memory SQLite.

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
ravi-kumar-pilla: create_connection still describes a “singleton connection”; the implementation caches an Engine. A one-line docstring tweak would match behavior.

kedro-datasets/kedro_datasets_experimental/polars/polars_database_dataset.py
```diff
@@ -96,12 +95,48 @@ class PolarsDatabaseDataset(AbstractDataset[None, pl.DataFrame]):
...
-        sql: "select shuttle, shuttle_id from spaceflights.shuttles;"
+        sql: "SELECT shuttle, shuttle_id FROM spaceflights.shuttles;"
+        credentials: db_credentials
+    ```
+
+    Pass extra arguments to the underlying polars methods via ``load_args`` and
+    ``save_args``:
+
+    ```yaml
+    shuttles_table:
```
ravi-kumar-pilla: L111 vs L130 — Two YAML snippets use the same catalog key shuttles_table:. They are separate blocks, but renaming this to (e.g. shuttles_configured:) would avoid implying one catalog merges all three.

ravi-kumar-pilla:
Hi @nikanton ,

Thank you for the contribution. I left few comments which are not major or blocking issues. I find the overall code pretty good. You might need to fix CI failures. This might be because of the `numpy==2.2.6` not yet supporting `python=3.14`. Can you try adding numpy `2.3.4` conditionally for python >= 3.14 and see if this works.

nikanton:
@ravi-kumar-pilla Thank you for your feedback. All changes have been implemented.

The conditional NumPy 2.3 dependency for Python 3.14 did not work because other datasets depend on numpy <2.3.

I am fixing this in #1408. Let’s merge that PR first, and then merge this one.

ravi-kumar-pilla:
Hi @nikanton ,

Now that #1408 is merged, Can you please fix the CI issues ?

Thank you

jitu5 (approved):
Thanks for the contributions. Could you please fix the CI issues and resolve the conflicts with main before we can merge this.

merelcht:
@ravi-kumar-pilla @jitu5 the builds are now passing, can you have a final look and if everything is fine, merge this?

ravi-kumar-pilla (approved):
Thank you @nikanton . LGTM 💯

PR #1387, opened 2026-04-23
feat(datasets): Add support for Python 3.14

lrcouto:
## Description

Adds Python 3.14 support to Kedro-Datasets.

A couple dependencies of Kedro-Datasets do not have themselves support for Python 3.14, which makes the specific datasets that depend on them require a previous version. Those dependencies are Tensorflow (required by tensorflow.TensorFlowModelDataset) and Fiona (required by geopandas.GenericDataset).

Here are the Github issues for both of those, discussing the version support:
https://github.com/tensorflow/tensorflow/issues/102890
https://github.com/Toblerity/Fiona/issues/1504

Because of that, tests for those two datasets are also not being run on Python 3.14.

A small change had to be made on the `_grandparent()` auxiliary function on PartitionedDataset because passing additional positional arguments to `PurePath.relative_to()` [is deprecated](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) on Python 3.14.

Other than that, it required version bumps for cloudpickle, pillow, pytest-cov and lxml.

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Updated `jsonschema/kedro-catalog-X.XX.json` if necessary
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ ] Added tests to cover my changes
- [ ] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)

PR #1408, opened 2026-05-20
fix(datasets): bump opencv-python to ~=4.13.0.92 for Python 3.14

nikanton:
## Summary
- `unit-tests-experimental (windows-latest, 3.14)` has been failing because `opencv-python~=4.12.0.88` (transitive dep of the video dataset, included in the `experimental_test` extras) pins `numpy>=2,<2.3.0`. NumPy <2.3.0 has no cp314 Windows wheel, so `uv` falls back to a MINGW-W64 source build of NumPy that aborts mid-build.
- This matrix entry has never run green: Python 3.14 was added to the experimental matrix in #1387, but the workflow's `paths:` filter means it only fires for PRs touching `kedro-datasets/kedro_datasets_experimental/**`. #1400 was the first such PR after #1387 and is what surfaced the latent failure.
- `opencv-python==4.13.0.92` is the current latest release. It drops the `numpy<2.3.0` cap (now only requires `numpy>=2` on Python ≥3.9) and ships abi3 wheels including `win_amd64`, so it installs cleanly on Python 3.14 with `numpy>=2.3.x`.
- Bumps the pin in all three places it appears: the standalone `video-videodataset` extra and both `experimental` / `experimental_test` extras.

## Test plan
- [x] CI: `unit-tests-experimental (windows-latest, 3.14)` resolves and goes green.
- [x] CI: the rest of the experimental matrix (ubuntu/windows × 3.10–3.14) stays green.
- [x] CI: `unit-tests` (non-experimental, which also reads the bumped `video-videodataset` pin) stays green.

### DIFF
--- a/.github/workflows/kedro-datasets-experimental.yml
+++ b/.github/workflows/kedro-datasets-experimental.yml
@@ -14,6 +14,7 @@ on:
       - main
     paths:
       - "kedro-datasets/kedro_datasets_experimental/**"
+      - ".github/workflows/kedro-datasets-experimental.yml"
 
 jobs:
   unit-tests-experimental:
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -34,6 +34,7 @@
 - Refactored shared validation and utility logic from the three Langfuse experimental datasets (`PromptDataset`, `EvaluationDataset`, `TraceDataset`) into a common `langfuse._common` module.
 - Added `os.PathLike` support for `plotly` datasets.
 - Added `checkpoint.filepath` validation for IncrementalDataset.
+- Bumped `opencv-python` to `~=4.13.0.92` so `experimental_test` resolves on Python 3.14 (the old `~=4.12.0.88` capped `numpy<2.3.0`, which has no Windows cp314 wheel).
 
 ## Community contributions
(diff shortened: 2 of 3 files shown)


ElenaKhaustova (commented):
Thank you, @nikanton, this looks correct!

**One thing though:** the failing job (`unit-tests-experimental (windows-latest, 3.14)`) doesn't run on this PR, because `.github/workflows/kedro-datasets-experimental.yml`` is path-filtered to kedro-datasets/kedro_datasets_experimental/**` only — and this PR only touches `pyproject.toml` and `RELEASE.md`. That's also why the failure stayed latent until #1400.

Could you extend the paths: filter in this same PR to also include `kedro-datasets/pyproject.toml` and `.github/workflows/kedro-datasets-experimental.yml`? That way the experimental matrix will actually run on this PR and prove the fix, and future dependency-only PRs won't fall into the same hole.

**And just to confirm:** did you also reproduce the green run locally / in your fork on Windows + py3.14?

ankatiyar (approved):
Thanks @nikanton 💯 
@ElenaKhaustova I've added the path to the workflow to trigger the tests and they seem to be passing!

ElenaKhaustova (approved):

SajidAlamQB (approved):

PR #1416, opened 2026-06-02
fix(datasets): Conform OpikDatasets to Tolerate extra credentials

### LINKED ISSUE #1417: kedro-datasets: filter credentials before splatting to Opik()
## Description

While verifying the Opik environment in #169, every `kedro run --env opik` that touched `opik.PromptDataset` or `opik.EvaluationDataset` crashed with:

TypeError: Opik.init() got an unexpected keyword argument 'endpoint'

Root cause is in `kedro-datasets-experimental`:

- `opik/prompt_dataset.py` calls `Opik(**credentials, **opik_kwargs)`
- `opik/evaluation_dataset.py` calls `Opik(**credentials)`

The `Opik` client doesn't accept arbitrary keys. The autogen-mode `TraceDataset` legitimately needs an `endpoint` in credentials (for the OTLP exporter), and that key then breaks the other Opik datasets that share `opik_credentials`.

This is inconsistent with sibling datasets:
- `langfuse.PromptDataset` does explicit key picking (`Langfuse(public_key=..., secret_key=...)`).
- `opik.TraceDataset` already uses explicit `configure(api_key=..., workspace=..., url=...)`.

The two Prompt/Evaluation datasets are the outliers.

SajidAlamQB:
## Description

`opik.PromptDataset` and `opik.EvaluationDataset` splat the entire `credentials` dict into `Opik(**credentials)`. 

Any key the Opik client doesn't recognise raises:
```TypeError: Opik.__init__() got an unexpected keyword argument 'endpoint'```

This effects users who want a single `opik_credentials` block to serve all three datasets: `opik.TraceDataset` in `mode: autogen` requires `endpoint` in credentials (for the OTLP exporter), but adding it then breaks the Prompt and Evaluation datasets sharing the same block.

`langfuse.PromptDataset` already avoids this by picking specific keys
(`public_key`, `secret_key`) explicitly. This PR mirrors that pattern for Opik.

## Development notes

A new helper in `opik/_common.py` filters credentials down to the keys `Opik()` actually accepts:

```
    OPIK_CLIENT_KEYS = frozenset({
        "api_key", "workspace", "host", "url_override", "project_name",
    })
```

Both `PromptDataset` and `EvaluationDataset` now route credentials through it. `TraceDataset` needs no change as it already picks keys explicitly via `configure(api_key=..., workspace=..., url=credentials.get("url_override"))` and reads `endpoint` directly only inside `_build_autogen_tracer`.

- Credentials containing only valid keys: unchanged (happy path).
- Credentials containing extras like `endpoint` or an `openai` sub-block:
  previously raised `TypeError`; now silently ignored by Prompt/Evaluation.
- Forward-compatible: if the Opik client gains a kwarg, add it to
  `OPIK_CLIENT_KEYS`.
- Added `test_init_filters_unknown_credentials` to both `test_prompt_dataset.py` and `test_evaluation_dataset.py`, asserting that extra credential keys are filtered out of the `Opik(...)` call.

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/opik/_common.py
+++ b/kedro-datasets/kedro_datasets_experimental/opik/_common.py
@@ -28,9 +28,9 @@ OPIK_CLIENT_KEYS = frozenset({
 def build_opik_client_kwargs(credentials: dict[str, Any]) -> dict[str, Any]:
     """Pick only credential keys accepted by `Opik()`.
 
-    Filters out keys used elsewhere in the same credentials block — e.g.
+    Filters out keys used elsewhere in the same credentials block e.g.
     `endpoint` (consumed by `TraceDataset` autogen mode) or an `openai`
-    sub-block (consumed by `TraceDataset` openai mode) — so they don't
+    sub-block (consumed by `TraceDataset` openai mode) so they don't
     reach the Opik client constructor and raise `TypeError`. This lets a
     single `opik_credentials` block serve all three datasets.
 
--- a/kedro-datasets/kedro_datasets_experimental/opik/_common.py
+++ b/kedro-datasets/kedro_datasets_experimental/opik/_common.py
@@ -26,19 +26,19 @@ OPIK_CLIENT_KEYS = frozenset({
 
 
 def build_opik_client_kwargs(credentials: dict[str, Any]) -> dict[str, Any]:
-    """Pick only credential keys accepted by ``Opik()``.
+    """Pick only credential keys accepted by `Opik()`.
 
     Filters out keys used elsewhere in the same credentials block — e.g.
-    ``endpoint`` (consumed by ``TraceDataset`` autogen mode) or an ``openai``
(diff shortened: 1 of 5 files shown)


merelcht (approved):
I didn't test it, but the code looks good to me 👍 

One thing I didn't fully follow is how `build_opik_client_kwargs` is used for `TraceDataset`? It's mentioned in the doc string, but I don't see it being called in trace dataset.

SajidAlamQB:
> I didn't test it, but the code looks good to me 👍
> 
> One thing I didn't fully follow is how `build_opik_client_kwargs` is used for `TraceDataset`? It's mentioned in the doc string, but I don't see it being called in trace dataset.

it's deliberately not called in TraceDataset that dataset never splats credentials into `Opik()`, it already picks keys explicitly through `configure(...)` and reads endpoint only in autogen mode. The docstring mentions TraceDataset just to explain where the extra keys in a shared credentials block come from. I'll reword it so that more clear.

ElenaKhaustova (approved):
👍

PR #1422, opened 2026-06-09
fix(datasets): Forward project_name in Opik langchain-mode tracer

### LINKED ISSUE #1418: kedro-datasets: forward project_name in Opik langchain-mode tracer
## Description

While verifying scenario B (LangGraph + Opik) in #169, every trace landed in Opik's "Default Project" no matter what `project_name` was set in `opik_credentials`. Confirmed: setting `OPIK_PROJECT_NAME` via shell env var didn't work either; passing `project_name:` as a catalog kwarg on `intent_tracer` did.

Root cause in `kedro_datasets_experimental.opik.TraceDataset`:

1. `_configure_opik` sets `OPIK_PROJECT_NAME` env var from credentials.
2. A subsequent `configure(force=True)` writes `~/.opik.config` without `project_name`, and that takes precedence.
3. `_load_langchain_tracer` calls `OpikTracer(**self._trace_kwargs)` without forwarding `project_name` from credentials.

Net effect: `opik_credentials.project_name` never reaches `OpikTracer` in langchain mode. (autogen mode is unaffected. It uses `start_as_current_span` directly and routing works.)

## Suggested fix

Mirror what `_load_openai_client` already does in the same file i.e. pull `project_name` from credentials, pass it to `OpikTracer`:

```
def _load_langchain_tracer(self) -> Any:
    from opik.integrations.langchain import OpikTracer
    project_name = self._credentials.get("project_name")
    kwargs = {**self._trace_kwargs}
    if project_name and "project_name" not in kwargs:
        kwargs["project_name"] = project_name
    return OpikTracer(**kwargs)
```

SajidAlamQB:
## Description

In `langchain` mode, `opik.TraceDataset` ignores `credentials.project_name`, so every LangChain trace lands in Opik's **"Default Project"** regardless of what's configured. This PR forwards `project_name` from credentials to `OpikTracer`, mirroring the existing `_load_openai_client` pattern in the same file.

Sibling of #1416 same theme of "make the Opik datasets consistent in how they handle credentials." Together, both PRs unblock removal of two workarounds in `kedro-academy/kedro-academy#168`

## Development notes
In `_load_langchain_tracer`, pull `project_name` from credentials when not explicitly overridden in catalog `trace_kwargs`:

```
project_name = self._credentials.get("project_name")
kwargs = {**self._trace_kwargs}
if project_name and "project_name" not in kwargs:
    kwargs["project_name"] = project_name

return OpikTracer(**kwargs)
```

## Checklist

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/opik/trace_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/opik/trace_dataset.py
@@ -13,6 +13,12 @@ REQUIRED_OPIK_CREDENTIALS = {"api_key", "workspace"}
 REQUIRED_OPIK_CREDENTIALS_AUTOGEN = {"endpoint"}
 OPTIONAL_OPIK_CREDENTIALS = {"project_name", "url_override"}
 
+# Opik configuration is global per process: the first project configured wins
+# once a client has been created. This tracks the project configured so far (a
+# mutable holder, to avoid a module-level `global` rebind) so we can warn if a
+# later TraceDataset configures a different one.
+_session_project_state: dict[str, str | None] = {"project_name": None}
+
 
 class TraceDataset(AbstractDataset):
     """Kedro dataset for managing Opik tracing clients and callbacks.
@@ -179,15 +185,36 @@ class TraceDataset(AbstractDataset):
         Note: Opik configuration is global within a process, so the project
         cannot be changed once a client has been created. Using multiple
         `TraceDataset` instances with different projects in the same session
-        will log all traces to the first configured project.
+        will log all traces to the first configured project; a warning is
+        emitted in that case.
         """
+        project_name = self._credentials.get("project_name")
+        configured_project = _session_project_state["project_name"]
+
+        if (
+            configured_project is not None
+            and project_name is not None
+            and project_name != configured_project
+        ):
+            logger.warning(
+                "Opik is already configured for project '%s' in this session. "
+                "Opik configuration is global per process, so configuring project "
+                "'%s' may not take effect for all traces — run pipelines targeting "
+                "different projects in separate processes.",
+                configured_project,
+                project_name,
+            )
+
         configure(
             api_key=self._credentials["api_key"],
             workspace=self._credentials["workspace"],
             url=self._credentials.get("url_override"),
-            project_name=self._credentials.get("project_name"),
+            project_name=project_name,
             force=True,
         )
+
+        if configured_project is None and project_name is not None:
+            _session_project_state["project_name"] = project_name
 
     def _validate_openai_client_params(self) -> None:
         """Validate OpenAI credentials in the 'openai' section.
--- a/kedro-datasets/kedro_datasets_experimental/opik/trace_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/opik/trace_dataset.py
@@ -167,35 +167,25 @@ class TraceDataset(AbstractDataset):
                     )
 
     def _configure_opik(self) -> None:
-        """Initialize Opik global configuration with awareness of project switching.
+        """Configure the Opik SDK from the provided credentials.
 
-        This function ensures that the Opik SDK is configured using the provided credentials.
-        If an existing configuration is detected (from a prior dataset instance),
-        a warning is emitted since the active project cannot be changed dynamically.
+        `project_name` is passed to `configure()` so it is persisted to
+        Opik's session configuration and picked up by the auto-created client.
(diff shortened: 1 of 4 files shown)


kedro-datasets/kedro_datasets_experimental/opik/trace_dataset.py
```diff
@@ -356,7 +356,17 @@ def _load_langchain_tracer(self) -> Any:
         except ImportError as e:
             raise DatasetError("Opik LangChain integration not available.") from e
 
-        return OpikTracer(**self._trace_kwargs)
+        # Forward project_name from credentials when not explicitly overridden
+        # in catalog trace_kwargs. Without this, OpikTracer logs to the
+        # "Default Project" regardless of credentials.project_name, because
+        # _configure_opik calls configure(force=True), which writes
+        # ~/.opik.config without project_name and overrides OPIK_PROJECT_NAME.
```
merelcht: So with this are you saying that 
```
 elif project_name:
            os.environ["OPIK_PROJECT_NAME"] = project_name
```

in `_configure_project` is not actually doing anything? It looks like this new code just uses the project name from credentials directly, completely ignoring the env var.

SajidAlamQB: Yes, effectively for routing, that env-var line is dead code. `configure(force=True)` (called right after) writes a project_name into opik's session config via `update_session_config(...)`, and `OpikConfig` resolves in the order init → session config → env vars → file → defaults so the session write permanently hides the env var. Since we don't pass `project_name` to `configure()`, it writes the suggested/default project, which is why everything landed in "Default Project".

So the fix bypasses the broken chain and passes the value explicitly to `OpikTracer` which is same pattern as `_load_openai_client`.

merelcht: Okay I think I get it. But then shouldn't we just remove those lines of code?

merelcht: And then do 
```
configure(
    ...,
    force=True,
    project_name=project_name,
)
```

instead? (https://github.com/comet-ml/opik/pull/6093)

SajidAlamQB: Really good catch! I switched to the `configure(project_name=...)` approach and dropped the dead env-var block. It fixes sdk mode too, and an explicit `catalog project_name: kwarg` still wins for the langchain tracer. 

But this needs `opik>=1.11.0` (where `configure()` gained the param), so I bumped the `opik-tracedataset extras`.

kedro-datasets/kedro_datasets_experimental/tests/opik/test_trace_dataset.py
```diff
@@ -42,22 +42,22 @@ def test_empty_optional_credential_raises(base_credentials):
...
+    client (and any tracer derived from it) logs to the right project."""
     creds = base_credentials | {"project_name": "test-proj"}
     TraceDataset(creds)
-    assert os.getenv("OPIK_PROJECT_NAME") == "test-proj"
     configure_mock.assert_called_once()
+    assert configure_mock.call_args.kwargs["project_name"] == "test-proj"
 
 
 @patch("kedro_datasets_experimental.opik.trace_dataset.configure")
-def test_configure_opik_warns_on_project_switch(configure_mock, base_credentials, caplog):
```
merelcht, line 54: Was this actually the desired behaviour? It looks like we defined that warning before in Kedro (not on the Opik side), so I'm wondering if actually we'd still like to explicitly warn when someone tries to switch project?

merelcht (commented):
I think this is nearly there, just double checking if we should still inform the user about it not being possible to change the active project in the same session?

kedro-datasets/kedro_datasets_experimental/tests/opik/test_trace_dataset.py
SajidAlamQB: Makes sense I re-added the warning. I originally dropped it only because the old one keyed off `OPIK_PROJECT_NAME`, which this PR no longer sets. This version doesn't depend on the env var and avoids false positives from `configure(force=True)` persisting project_name to ~/.opik.config. Added a test for it too.

merelcht (approved):
Thanks @SajidAlamQB ! This now looks like the right behaviour to me 👍

ElenaKhaustova (approved):
Thank you, @SajidAlamQB, LGTM!

Issue #1157, opened 2025-08-08
Implement a Vector DB Dataset for Kedro


lrcouto:
This feature request proposes the creation of a Kedro dataset for interacting with vector databases. This was one of the suggested "quick wins" for [this spike](https://github.com/kedro-org/kedro/issues/5010). 

It could be an extension of or based on the [DeeplakeVectorStoreDataset](https://github.com/kedro-org/kedro-academy/blob/main/kedro-rag-chatbot/src/kedro_rag_chatbot/datasets/vector_store_dataset.py) written by Elena for her [Kedro RAG chatbot demo](https://github.com/kedro-org/kedro-academy/tree/main/kedro-rag-chatbot).

### Why?

Having a dataset that interacts with vector databases would help Kedro users more easily build retrieval-augmented generation (RAG) workflows, LLM pipelines, semantic search systems, and other AI-related projects.

### Desirable features

A backend-agnostic design would be desirable, with a base implementation of a AbstractVectorStoreDataset defining a minimal interface (save, load, query, etc.). Could also have concrete implementations for some commonly used backends, like DeepLake, Chroma, Pinecone, Weaviate, etc.

Support for both reading and writing. Could be used for storing new embeddings or documents, retrieving entries via vector similarity search, etc.

A nice-to-have would be an in-memory/mock backend so the user could run test pipelines without having to provide a real database.

### Challenges

Abstraction complexity: Vector DBs differ in their query APIs, metadata handling, and indexing strategies. Defining a common minimal API that feels natural is likely to be difficult.

Dependency management: If we do build specific datasets with backend-specific dependencies, this can potentially bring in a lot of new dependencies to us, that we will then have to keep up with.

Library specific quirks: Cloud-hosted vector DBs (e.g., Pinecone, Weaviate Cloud, Milvus Cloud) introduce network latency and potential rate limits, whereas in-process libraries (e.g., FAISS, Chroma in local mode) can be extremely fast but require enough local memory and disk space.

Indexing strategies: Some backends provide real-time indexing (new vectors are searchable immediately), while others need explicit index refresh or batch processing. The dataset API should not abstract this away entirely, or users might bump into inconsistent search results.

PR #1430, opened 2026-06-24
feat(datasets): Add VectorStore abstractions

### LINKED ISSUE #1429: Implement VectorStore core abstractions
Part of https://github.com/kedro-org/kedro-plugins/issues/1351

Define core abstraction:
    `AbstractVectorStoreDataset`
    `VectorStoreHandle` (minimal interface)

lrcouto:
## Description

This PR introduces the foundational abstractions for vector store support in `kedro-datasets`, following the design decisions made in https://github.com/kedro-org/kedro-plugins/issues/1157#issuecomment-4032483936. It adds two abstract base classes with no external dependencies.

First part of https://github.com/kedro-org/kedro-plugins/issues/1351

#### AbstractVectorStoreDataset

The `AbstractVectorStoreDataset` is a minimal implementation of an `AbstractDataset` subclass with only `save`, `_load` and `describe` methods.

**Why `save` and not `_save` like it was suggested in the original issue?**

Because of `AbstractDataset` itself. It has a `__init_subclass__` hook that runs automatically every time a new subclass is defined. This hook aliases `_load` to `load` and `_save` to `save`, then wraps both with logging and DatasetError handling. This is the Kedro convention that lets dataset authors implement the private `_load`/`_save` and get the public interface automatically.

```py
        if hasattr(cls, "_load") and not cls._load.__qualname__.startswith("Abstract"):
            cls.load = cls._load  # type: ignore[method-assign]

        if hasattr(cls, "_save") and not cls._save.__qualname__.startswith("Abstract"):
            cls.save = cls._save  # type: ignore[method-assign]
```

However, `AbstractVectorStoreDataset` does start with "Abstract". It would be skipped by this rule.

A `save` method still falls here:

```py
        if hasattr(cls, "save") and not cls.save.__qualname__.startswith("Abstract"):
            cls.save = cls._save_wrapper(  # type: ignore[assignment]
                cls.save
                if not getattr(cls.save, "__savewrapped__", False)
                else cls.save.__wrapped__  # type: ignore[attr-defined]
            )
```

So this means we miss out on `AbstractDataset`'s `_save_wrapper` which sets some logging and error handling for this method. Instead we replace the `save` method with a new one, which is also barebones.

For the load method it does not matter because, by the way it was decided that these classes will by design just return a `VectorStoreHandle` object, and the wrapping will still happen when whichever subclass this is used for implements the load behavior.

#### VectorStoreHandle

Also a minimal interface, with `add`, `search`, `delete`, `close` and `describe` methods plus a `raw_client` property to expose the underlying backend directly, letting the user do something that the interface might not cover.

## Checklist


kedro-datasets/kedro_datasets_experimental/vectorstore_base/abstract_vector_store_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+
+    The returned handle owns a backend connection and must be closed after
+    use, either explicitly via ``handle.close()`` or via the context manager::
+
+        with catalog.load("my_store") as store:
+            store.add([{"properties": {...}, "vector": [...]}])
+            hits = store.search(vector=[...], top_k=5)
+    """
+
+    def save(self, data: Any) -> None:
```
ElenaKhaustova, line 31: This is `save()` and not `_save()` intentionally (as explained in the PR description), but the code itself doesn't say why. This is a subtle coupling to Kedro's `__init_subclass__` qualname check — if a future maintainer "fixes" it to `_save()`, the `__init_subclass__` hook will skip it because the qualname starts with `"Abstract"`, and save will silently stop raising.

Could you add a brief comment above this method? Something like:

```python
# Intentionally overrides `save` (not `_save`): AbstractDataset.__init_subclass__
# skips methods whose qualname starts with "Abstract", so defining `_save` here
# would be ignored by the wrapping machinery.
```

kedro-datasets/kedro_datasets_experimental/vectorstore_base/abstract_vector_store_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+            "handle.add() for inserts and handle.delete() for removals."
+        )
+
+    @abstractmethod
+    def _load(self) -> VectorStoreHandle:
+        """Return a connected handle for the target collection."""
+
+    @abstractmethod
+    def _describe(self) -> dict[str, Any]:
+        """Return a dict of constructor arguments for display / logging."""
```
ElenaKhaustova, line 44: Right now `_exists()` is inherited from `AbstractDataset`, which logs a warning:

```
"'exists()' not implemented for 'WeaviateVectorStoreDataset'. Assuming output does not exist."
```

That message implies something is *missing* that should have been implemented. For vector store datasets, not having a default `exists()` is intentional — what "exists" means is backend-specific (collection exists? has data? backend is reachable?).

I'd suggest overriding `_exists()` here to return `False` silently, leaving it to concrete subclasses to provide meaningful checks:

```python
def _exists(self) -> bool:
    return False
```

This also avoids noisy warnings in test output.

kedro-datasets/kedro_datasets_experimental/vectorstore_base/vector_store_handle.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def raw_client(self) -> Any:
+        """The underlying backend client, for operations outside this interface."""
+
+    @abc.abstractmethod
+    def close(self) -> None:
+        """Release the connection held by this handle."""
```
ElenaKhaustova: Nit: Worth documenting that `close()` should be idempotent. Calling it twice should be harmless.

kedro-datasets/kedro_datasets_experimental/vectorstore_base/vector_store_handle.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        *,
+        vector: list[float] | None = None,
+        text: str | None = None,
+        top_k: int = 10,
+        filters: Any = None,
+    ) -> list[dict[str, Any]]:
+        """Return the ``top_k`` nearest neighbours.
+
+        Exactly one of ``vector`` or ``text`` must be provided.  This
```
ElenaKhaustova, line 51: The docstring says each concrete backend is responsible for raising if neither or both `vector`/`text` are given. Worth adding a note that backends may also raise `NotImplementedError` if they only support one of the two modes (e.g. FAISS doesn't support text search without an external vectorizer). This sets expectations for implementers.

kedro-datasets/kedro_datasets_experimental/tests/vectorstore_base/test_vectorstore_base.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def search(
+        self,
+        *,
+        vector: list[float] | None = None,
+        text: str | None = None,
+        top_k: int = 10,
+        filters: Any = None,
+    ) -> list[dict[str, Any]]:
+        if vector is None and text is None:
```
ElenaKhaustova: The contract says "exactly one of vector or text must be provided", but the `DummyHandle` only validates the "neither provided" case. It silently accepts both being provided. Since this dummy serves as a reference implementation for backend authors, it would be more instructive to validate the full constraint:

```python
if (vector is None) == (text is None):
    raise ValueError("Exactly one of vector or text must be provided.")
```

You could also add a corresponding test in `TestVectorStoreHandle`:

```python
def test_search_rejects_both_vector_and_text(self, handle):
    with pytest.raises((ValueError, DatasetError)):
        handle.search(vector=[0.1], text="hello")
```

kedro-datasets/kedro_datasets_experimental/tests/vectorstore_base/test_vectorstore_base.py
```diff
@@ -0,0 +1,186 @@
...
+# AbstractVectorStoreDataset contract
+# ---------------------------------------------------------------------------
+
+
+class TestAbstractVectorStoreDataset:
+    def test_load_returns_handle(self, dataset):
+        handle = dataset.load()
+        assert isinstance(handle, VectorStoreHandle)
+
+    def test_save_raises_dataset_error(self, dataset):
```
ElenaKhaustova, line 101: Consider adding a test for `save(None)` to document that the base class's `None` guard is intentionally bypassed:

```python
def test_save_none_raises_dataset_error(self, dataset):
    with pytest.raises(DatasetError, match="Saving is not supported"):
        dataset.save(None)
```

Since `save()` is overridden directly (not `_save()`), the `AbstractDataset.save()` `None` check never runs. This test makes that behavior explicit.

kedro-datasets/kedro_datasets_experimental/tests/vectorstore_base/test_vectorstore_base.py
```diff
@@ -0,0 +1,186 @@
+"""Contract tests for AbstractVectorStoreDataset and VectorStoreHandle.
```
ElenaKhaustova, line 1: The file is named `test_abstract_vector_store_dataset.py` but it covers both `AbstractVectorStoreDataset` and `VectorStoreHandle` equally. Consider renaming to `test_vectorstore_base.py` (matching the package name) so it's clear the file tests the full module, not just the dataset class.

ElenaKhaustova (changes requested):
Thank you, @lrcouto! The abstraction looks good 👍

A  few things I'd like to see addressed before merge, and made some smaller suggestions, see inline comments.

lrcouto:
Thank you @ElenaKhaustova. Added the suggested changes.

- Renamed the test file to `test_vectorstore_base.py`
- Added `_exists() -> False` override to prevent the warning from AbstractDataset from coming through
- Added NotImplementedError note to search() docstring for backends supporting only one mode  
- Added more detail to docstrings, idempotency requirement to close() and an explanation of why save() is used over _save()
- Fixed `DummyHandle.search()` to reject both-provided case: `if (vector is None) == (text is None)`. Did the same to `DummyHandle.delete()` since it was similar.
- Added `super().__init__()` to `DummyVectorStoreDataset.__init__()` so it's more consistent with other dataset tests.
- Added new tests for search() and delete().

kedro-datasets/RELEASE.md
```diff
@@ -6,6 +6,7 @@
 # Release 9.5.0
 
 ## Major features and improvements
+* Added `vectorstore_base.AbstractVectorStoreDataset` and `vectorstore_base.VectorStoreHandle`, backend-agnostic abstract base classes for vector store datasets.
```
ElenaKhaustova: Since we already released 9.5.0, it should go to the upcoming release section.

ElenaKhaustova (approved):
Thank you, @lrcouto, it looks good now!

Left one minor comment about the release note.

Approving it, but let's merge after we're happy with the https://github.com/kedro-org/kedro-plugins/issues/1351 implementation on top of it.

kedro-datasets/kedro_datasets_experimental/tests/vectorstore_base/test_vectorstore_base.py
```diff
@@ -0,0 +1,205 @@
...
+from kedro.io.core import DatasetError
+
+from kedro_datasets_experimental.vectorstore_base import (
+    AbstractVectorStoreDataset,
+    VectorStoreHandle,
+)
+
+# ---------------------------------------------------------------------------
+# Dummy in-memory implementations
+# ---------------------------------------------------------------------------
```
ankatiyar: These 3 line comments are unnecessary I think. The ones below on this page as well

ankatiyar (commented):
Thanks @lrcouto the code looks good! I was just wondering since this folder "vectorstore_base" only contains abstract dataset classes, it might be confused for an actual usable dataset module. Could we decide if we should have a folder for all abstract classes, or a `_utils` folder or should we name it `_vectorstore_base` to make it clear? cc @ElenaKhaustova

lrcouto:
> Thanks @lrcouto the code looks good! I was just wondering since this folder "vectorstore_base" only contains abstract dataset classes, it might be confused for an actual usable dataset module. Could we decide if we should have a folder for all abstract classes, or a `_utils` folder or should we name it `_vectorstore_base` to make it clear? cc @ElenaKhaustova

I like _vectorstore_base. Makes it clear that it's not an usable dataset but also you can know what's in there at a glance.

ElenaKhaustova:
> Thanks @lrcouto the code looks good! I was just wondering since this folder "vectorstore_base" only contains abstract dataset classes, it might be confused for an actual usable dataset module. Could we decide if we should have a folder for all abstract classes, or a `_utils` folder or should we name it `_vectorstore_base` to make it clear? cc @ElenaKhaustova

Good question!

I'd suggest keeping `vectorstore_base` as-is. The `_base` suffix already signals `foundational, not directly usable.`

The key reason against a leading underscore `_vectorstore_base` is that both classes are public API:

- Backend authors import `AbstractVectorStoreDataset` and `VectorStoreHandle` to subclass them (e.g. `WeaviateVectorStoreHandle(VectorStoreHandle)`)
- Users will import `VectorStoreHandle` for type hints in their nodes

I think leading underscore would signal "don't import this" for something explicitly intended for both audiences.

A shared `_base/` or`_utils/`folder would also be premature. There's only one abstraction pattern right now, and if more emerge, they'd likely be domain-specific too.

ankatiyar:
@ElenaKhaustova Makes sense!

ankatiyar (approved):
LGTM 💯

SajidAlamQB (approved):
LGTM really clean foundation, thanks @lrcouto.

PR #1433, opened 2026-06-25
fix(datasets): harden experimental PyTorchDataset against unsafe torch.load

### LINKED ISSUE #1432: Harden experimental `PyTorchDataset`: fix misleading docstring, broken load_args, and unsafe torch.load default
## Description

A security report ([#1431](https://github.com/kedro-org/kedro-plugins/issues/1431)) flagged that `kedro_datasets_experimental.pytorch.PyTorchDataset` deserialises untrusted pickle data via `torch.load`, while its docstring advertises it as pickle-safe.

**Triage / severity:** The dataset is experimental (not TSC-maintained, not run in CI), and the RCE-on-load behaviour is inherent to `torch.load(weights_only=False)`. Standard PyTorch behaviour that applies to any pickle-based loader and requires an attacker to already control the loaded file. We are assessing the severity for this and whether it requires a security advisory.

## Tasks

**1. Fix the misleading docstring** (`pytorch_dataset.py`)

The current text claims the dataset uses "PyTorch's recommended zipfile serialisation protocol to avoid security issues with Pickle." This is false — the zipfile is only a container; the payload is still unpickled by `torch.load`. *Why:* this is the highest-value fix. By advertising the dataset as the pickle-*safe* option, we may lead users to point it at untrusted checkpoints they would never feed to `PickleDataset`. Replace with an accurate note that loading deserialises a pickle and is unsafe for untrusted files unless `weights_only=True`.

**2. Fix `load_args` being dropped on load** (`pytorch_dataset.py`)

`__init__` builds `self._load_args` from `DEFAULT_LOAD_ARGS` + user `load_args`, but `load()` passes `**self._fs_open_args_load` instead of `**self._load_args`. *Why:* this is a real functional bug independent of security — a user who sets `load_args: {weights_only: true}` has it silently ignored, so there is currently no working way to harden the dataset. Pass `self._load_args` to `torch.load` and route filesystem open args to the fsspec layer where they belong.

**3. Default to `weights_only=True` and clarify the `torch` dependency**

Set `weights_only=True` by default in `load()` (via `DEFAULT_LOAD_ARGS`) with a documented opt-out for trusted files. *Why:* a standard install can currently resolve a vulnerable `torch < 2.6` where the unsafe default applies; enforcing `weights_only=True` in our own default protects users regardless of the installed PyTorch version (the `weights_only` argument has existed since torch 1.13). Document that `weights_only=False` should only be used for trusted files, and recommend `safetensors` / `torch>=2.6` for untrusted inputs.

merelcht:
## Description

Closes #1432. Addresses a security report against the experimental `kedro_datasets_experimental.pytorch.PyTorchDataset`.

The dataset was documented as a pickle-*safe* alternative, but `load()` called `torch.load` without `weights_only=True`, which on PyTorch < 2.6 defaults to `weights_only=False` and deserialises arbitrary pickle opcodes, allowing remote code execution from a crafted `.pt` file. It also silently dropped user-supplied `load_args`, so there was no working way to opt into safe loading.

To set expectations: this is an **experimental** dataset (not TSC-maintained, contributed by an OS user) and the underlying behaviour is inherent to PyTorch's `torch.load`. 

## What changed

- **Enforce `weights_only=True` by default** via `DEFAULT_LOAD_ARGS`, with a documented opt-out (`load_args: {weights_only: false}`) for trusted files.
- **Fix `load_args` being dropped on load** — `load()` now passes `self._load_args` to `torch.load` (was `self._fs_open_args_load`), and filesystem open args are routed to the fsspec layer via `self._fs.open(...)`. This also makes `fs_args.open_args_load/save` actually take effect.
- **Correct the misleading docstring** — it no longer claims to "avoid security issues with Pickle"; instead it warns that loading deserialises a pickle and is unsafe for untrusted files unless `weights_only=True`, and points to `torch>=2.6` / `safetensors` for untrusted inputs.
- Added tests for the safe default and `load_args` propagation

## Why not pin `torch>=2.6`

The `weights_only` argument has existed since torch 1.13, so enforcing `weights_only=True` in our own default protects users regardless of the installed PyTorch version. A hard version pin would needlessly break existing installs, so the dependency is left unpinned and the docstring recommends `torch>=2.6`/`safetensors` for untrusted inputs.

## On the Bandit B614 suppression

The original code suppressed Bandit's B614 rule (`torch.load` without `weights_only`) with `# nosec: B614`, at the time hiding a real vulnerability. I removed it initially, but B614 is a static analysis rule that requires `weights_only=True` as a **literal** to pass. Since `weights_only` is passed via `**self._load_args` to preserve the user's ability to opt out for trusted files, Bandit cannot statically resolve it regardless of the default.

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/pytorch/pytorch_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/pytorch/pytorch_dataset.py
@@ -16,7 +16,19 @@ from kedro.io.core import (
 
 
 class PyTorchDataset(AbstractVersionedDataset[Any, Any]):
-    """`PyTorchDataset` loads and saves PyTorch models' `state_dict` using PyTorch's recommended zipfile serialization protocol to avoid security issues with Pickle.
+    """`PyTorchDataset` loads and saves PyTorch models' `state_dict` using ``torch.save``
+    and ``torch.load``.
+
+    .. warning::
+        Loading is **not** safe for untrusted files. ``torch.load`` deserializes a
+        pickle stream (the zipfile produced by ``torch.save`` is only a container
+        around that pickle), so a maliciously crafted ``.pt`` file can execute
+        arbitrary code on load. To mitigate this, ``PyTorchDataset`` enforces
+        ``weights_only=True`` by default, which restricts loading to tensors and a
+        small allow-list of safe types. Only set ``load_args: {weights_only: false}``
+        for files you fully trust, and prefer ``torch>=2.6`` (where ``weights_only=True``
+        is also the upstream default) or a non-pickle format such as ``safetensors``
+        when handling untrusted inputs.
 
     ### Example usage for the [YAML API](https://kedro.readthedocs.io/en/stable/data/data_catalog_yaml_examples.html)
 
@@ -47,7 +59,11 @@ class PyTorchDataset(AbstractVersionedDataset[Any, Any]):
 
     """
 
-    DEFAULT_LOAD_ARGS: dict[str, Any] = {}
+    # Enforce safe deserialization by default. ``weights_only=True`` restricts
+    # ``torch.load`` to tensors and a small allow-list of safe types, blocking the
+    # arbitrary-code-execution path through pickle's ``__reduce__``. Users can opt
+    # out for trusted files via ``load_args``.
+    DEFAULT_LOAD_ARGS: dict[str, Any] = {"weights_only": True}
     DEFAULT_SAVE_ARGS: dict[str, Any] = {}
 
     def __init__(  # noqa: PLR0913
@@ -64,6 +80,9 @@ class PyTorchDataset(AbstractVersionedDataset[Any, Any]):
         _fs_args = deepcopy(fs_args) or {}
         _fs_open_args_load = _fs_args.pop("open_args_load", {})
         _fs_open_args_save = _fs_args.pop("open_args_save", {})
+        # PyTorch serialization is binary; open in binary mode by default.
+        _fs_open_args_load.setdefault("mode", "rb")
+        _fs_open_args_save.setdefault("mode", "wb")
         _credentials = deepcopy(credentials) or {}
 
         protocol, path = get_protocol_and_path(filepath, version)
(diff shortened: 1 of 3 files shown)


aalbalawi1:
Hi @merelcht,

I appreciate the quick fix, but I'm confused by the triage rationale.

You're fixing:
1. A **misleading docstring** that falsely promises pickle-safety
2. A **broken `load_args`** that silently ignores `weights_only=True`
3. **Defaulting to `weights_only=True`** to protect users

Yet you're saying this is "not a security advisory" while explicitly acknowledging that the dataset:
- **Falsely advertises itself as pickle-safe**
- **Has no working way to enforce safe loading**
- **Defaults to unsafe behavior on PyTorch < 2.6**

Similar issues in **PyTorch (CVE-2025-32434, 9.8 Critical)** and **lmdeploy (8.8 High)** received CVEs.

The fix is good. The classification is inconsistent. A CVE would simply reflect the reality you've already confirmed.

Can we reconsider, or at least provide a formal response so I can pursue this through MITRE?

Thanks again for the work on this.

kedro-datasets/kedro_datasets_experimental/pytorch/pytorch_dataset.py
```diff
@@ -104,11 +123,16 @@ def _describe(self) -> dict[str, Any]:
 
     def load(self) -> Any:
         load_path = get_filepath_str(self._get_load_path(), self._protocol)
-        return torch.load(load_path, **self._fs_open_args_load)  #nosec: B614
+        with self._fs.open(load_path, **self._fs_open_args_load) as fs_file:
+            # nosec: B614 - weights_only defaults to True via DEFAULT_LOAD_ARGS,
```
ElenaKhaustova: Nit: This comment line starts with `# nosec: B614`, which looks like a Bandit directive, and the actual suppression `# nosec: B614` comes later. Shall we change this to a plain comment to avoid confusion?

`# B614 justification: weights_only defaults to True via DEFAULT_LOAD_ARGS,...`

kedro-datasets/kedro_datasets_experimental/tests/pytorch/test_pytorch_dataset.py
```diff
@@ -63,3 +63,23 @@ def test_save_and_load_dataset(self, pytorch_dataset, dummy_model, filepath_mode
...
+        """User-supplied load_args must reach torch.load, not be silently dropped."""
+        dataset = PyTorchDataset(
+            filepath=filepath_model, load_args={"weights_only": False}
+        )
+        dataset.save(TheModelClass())
+        mocked_load = mocker.patch(
+            "kedro_datasets_experimental.pytorch.pytorch_dataset.torch.load"
+        )
+        dataset.load()
+        assert mocked_load.call_args.kwargs["weights_only"] is False
```
ElenaKhaustova, line 85: The save side also changed and `torch.save` now receives `self._save_args` instead of `self._fs_open_args_save`. Could we add a symmetric test?

ElenaKhaustova (approved):
Thank you, @merelcht!

Minor: the `save()` method also changed from passing `self._fs_open_args_save` to `torch.save`, but the PR description only calls out the load-side fix.

kedro-datasets/kedro_datasets_experimental/pytorch/pytorch_dataset.py
merelcht: Good point! I'll change that.

kedro-datasets/kedro_datasets_experimental/tests/pytorch/test_pytorch_dataset.py
merelcht: Done!

ankatiyar (approved):
LGTM!

Issue #918, opened 2024-10-29
ibis.FileDataset read files from web


mark-druffel:
## Description
ibis.FileDataset fails when trying to read files from certain domains (or all domains) on the web. With ibis, I can read data from hugging face:
```
import ibis
con = ibis.duckdb.connect()
tracks = con.read_csv("hf:/datasets/maharshipandya/spotify-tracks-dataset/dataset.csv")
```
Based on a conversation with @deepyaman, the above code might only work because [duckdb has a hugging face extension](https://duckdb.org/2024/05/29/access-150k-plus-datasets-from-hugging-face-with-duckdb.html). That said, for what it's worth I found a a github repo with the same file and tried to read it with ibis and it worked:

```
import ibis
con = ibis.duckdb.connect()
tracks = con.read_csv("https://raw.githubusercontent.com/seanwryan/DS210-Final-Project/refs/heads/main/spotify.csv")
```
However, when adding either file (`hf:/` or `raw.githubusercontent`) as a FileDataset, my pipeline fails:
```
tracks:
  type: ibis.FileDataset
  filepath: hf://datasets/maharshipandya/spotify-tracks-dataset/dataset.csv
  file_format: csv
  connection: ${connection:spotify}
  load_args:
    sep: ","
  save_args:
    materialized: view
    overwrite: True
```
`kedro.io.core.DatasetError: Failed while loading data from dataset FileDataset(backend=duckdb, file_format=csv, filepath=hf:/datasets/maharshipandya/spotify-tracks-dataset/dataset.csv, load_args={'sep': ,}, save_args={'materialized': view, 'overwrite': True}).
IO Error: No files found that match the pattern "/spotify/hf:/datasets/maharshipandya/spotify-tracks-dataset/dataset.csv"`:

The issue appears to be with `_get_load_path()`.

## Context
I don't fully understand the extent of the issue so it's hard to advocate for fixing it. Obviously, being able to read files from data lakes (s3, azure, etc.) is essential, but I think those already work with `_get_load_path()`.  Having the ability to plug a url pointing to a file anywhere on the web could be really convenient, but I don't have a really solid use case for it at the moment. 

## Possible Implementation
I'm not entirely sure this is valid, but since ibis seems to work on it's own in these examples perhaps ibis.FileDataset could use pass a raw path to ibis before failing... 

## Possible Alternatives
The easiest workaround is to just download the files.

deepyaman:
Going to copy in my response from our Slack conversation:

[...]

PR #1436, opened 2026-06-27
fix(datasets): preserve remote URIs in ibis.FileDataset

### LINKED ISSUE #1298: bug(datasets): ibis.FileDataset does not work with abfss (or other cloud protocols?)

deepyaman:
## Description

Fixes #1298 and #918.

`ibis.FileDataset` wrapped the filepath in `PurePosixPath`, which collapses the
`//` in remote URIs (`s3://` → `s3:/`, `abfss://` → `abfss:/`, `hf://`,
`https://`, ...). Every subsequent operation then treated the mangled string as
a local path, so cloud and web reads/writes failed (the `s3://` example in the
docstring never actually worked).

This mirrors how `spark.SparkDataset` already solves the same problem: peel the
protocol prefix off with `split_filepath`, store the remainder as a
`PurePosixPath`, and reattach the prefix (`self._fs_prefix + str(...)`) before
handing the path to Ibis. The Ibis backend still does the actual I/O and its own
authentication via `connection`/`credentials`.

For Kedro version discovery and existence checks on remote paths, an `fsspec`
filesystem is built **lazily** (so backend-only users aren't forced to install
`adlfs`, `s3fs`, etc.) and a new `fs_args` argument carries its credentials,
kept separate from the Ibis connection `credentials`.

Note: this supersedes the path handling from #1380 — that PR correctly fixed a
*different* bug (#1378, Polars calling `len()` on a `PurePosixPath`), and its
regression test is preserved here.

### Minor behavior changes (cosmetic; not release-noted)

- `ibis.FileDataset`'s string representation now shows the full URI (prefix +
  path) rather than the previously mangled, local-looking path.
- Trailing slashes in paths are normalized away (e.g. `s3://bucket/dir/` becomes
  `s3://bucket/dir`).

## Development notes

- Relocated `split_filepath` to a dependency-free `kedro_datasets/_utils/file_utils.py`
  and re-exported it from `databricks_utils` (which imports `pyspark`), so it can
  be reused by Ibis without dragging in Spark.
- Added tests: remote-URI round-trip for `s3`/`abfss`/`hf`, the lazy-`fsspec`
  guarantee (no filesystem built for unversioned remote I/O), protocol-less
  globbing for remote versioning, and `fs_args` forwarding.
- Verified end-to-end by reading a public `https://` CSV through DuckDB.
- A small consistency follow-up applying the same `removesuffix("://")` tidy-up
  to `spark.SparkDataset` is in #1441.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -8,6 +8,7 @@
 - Repaired `polars.PolarsDatabaseDataset` end-to-end and added a full test suite for it.
 - Fixed `opik.TraceDataset` so `credentials.project_name` is now passed to `configure()` and persisted to Opik's session configuration.
 - Added `os.PathLike` support for `Spark` datasets.
+- Fixed `ibis.FileDataset` to support remote filepaths (e.g. `s3://`, `abfss://`, `hf://`) and added an `fs_args` argument to authenticate the filesystem used for version discovery.
 
 ## Community contributions
 - [@PragnyaKhandelwal](https://github.com/PragnyaKhandelwal)
--- a/kedro-datasets/kedro_datasets/_utils/file_utils.py
+++ b/kedro-datasets/kedro_datasets/_utils/file_utils.py
@@ -1,0 +1,24 @@
+"""Filesystem path utilities shared across datasets."""
+from __future__ import annotations
+
+import os
+
+
+def split_filepath(filepath: str | os.PathLike) -> tuple[str, str]:
+    """Split a filepath into its protocol prefix (e.g. ``s3://``) and the rest.
+
+    Unlike ``pathlib``-based parsing, this preserves the ``//`` after the
+    protocol, so cloud/remote URIs (``s3://``, ``abfss://``, ``hf://``, ...)
+    survive a round trip when the prefix is concatenated back onto the path.
+
+    Args:
+        filepath: Raw filepath, e.g. ``s3://bucket/data.parquet`` or a local path.
+
+    Returns:
+        A ``(prefix, path)`` tuple. ``prefix`` is ``""`` for local paths;
+        otherwise it ends with ``://`` (e.g. ``s3://``).
+    """
+    split_ = str(filepath).split("://", 1)
+    if len(split_) == 2:  # noqa: PLR2004
+        return split_[0] + "://", split_[1]
+    return "", split_[0]
--- a/kedro-datasets/kedro_datasets/ibis/file_dataset.py
+++ b/kedro-datasets/kedro_datasets/ibis/file_dataset.py
@@ -5,10 +5,12 @@ from copy import deepcopy
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING, Any, ClassVar
 
+import fsspec
 import ibis.expr.types as ir
 from kedro.io import AbstractVersionedDataset, DatasetError, Version
 
 from kedro_datasets._utils import ConnectionMixin
+from kedro_datasets._utils.file_utils import split_filepath
 
 if TYPE_CHECKING:
     from ibis import BaseBackend
@@ -83,6 +85,7 @@ class FileDataset(ConnectionMixin, AbstractVersionedDataset[ir.Table, ir.Table])
         credentials: dict[str, Any] | None = None,
         load_args: dict[str, Any] | None = None,
         save_args: dict[str, Any] | None = None,
+        fs_args: dict[str, Any] | None = None,
         version: Version | None = None,
         metadata: dict[str, Any] | None = None,
     ) -> None:
@@ -118,6 +121,12 @@ class FileDataset(ConnectionMixin, AbstractVersionedDataset[ir.Table, ir.Table])
                 `read_{file_format}` method.
             save_args: Additional arguments passed to the Ibis backend's
                 `to_{file_format}` method.
+            fs_args: Extra arguments to pass into underlying filesystem class
+                constructor (e.g. ``{"project": "my-project"}`` for
+                ``GCSFileSystem``). Used only to discover versions and check
+                existence of a remote ``filepath``; reading and writing is left
+                to the Ibis backend. Note that fsspec credentials go here, not
+                in ``credentials`` (which configures the Ibis connection).
             version: If specified, should be an instance of
                 ``kedro.io.core.Version``. If its ``load`` attribute is
                 None, the latest version will be loaded. If its ``save``
@@ -130,12 +139,17 @@ class FileDataset(ConnectionMixin, AbstractVersionedDataset[ir.Table, ir.Table])
         _connection_config = connection or self.DEFAULT_CONNECTION_CONFIG
         _credentials = deepcopy(credentials) or {}
         self._connection_config = {**_connection_config, **_credentials}
+        self._fs_args = deepcopy(fs_args or {})
         self.metadata = metadata
 
+        self._fs_prefix, path = split_filepath(filepath)
+        self._fs: fsspec.AbstractFileSystem | None = None
+
         super().__init__(
-            filepath=PurePosixPath(filepath),
+            filepath=PurePosixPath(path),
             version=version,
-            exists_function=lambda filepath: Path(filepath).exists(),
+            exists_function=self._fs_exists if self._fs_prefix else None,
+            glob_function=self._fs_glob if self._fs_prefix else None,
         )
 
         # Set load and save arguments, overwriting defaults if provided.
@@ -159,20 +173,34 @@ class FileDataset(ConnectionMixin, AbstractVersionedDataset[ir.Table, ir.Table])
         """The ``Backend`` instance for the connection configuration."""
         return self._connection
 
+    def _filesystem(self) -> fsspec.AbstractFileSystem:
+        # Build lazily so backend-only users don't need adlfs, s3fs, etc.
+        if self._fs is None:
+            protocol = self._fs_prefix.removesuffix("://")
+            self._fs = fsspec.filesystem(protocol, **self._fs_args)
+        return self._fs
+
+    def _fs_exists(self, path: str) -> bool:
+        return self._filesystem().exists(path)
+
+    def _fs_glob(self, pattern: str) -> list[str]:
+        return self._filesystem().glob(pattern)
+
     def load(self) -> ir.Table:
-        load_path = str(self._get_load_path())
+        load_path = self._fs_prefix + str(self._get_load_path())
         reader = getattr(self.connection, f"read_{self._file_format}")
         return reader(load_path, table_name=self._table_name, **self._load_args)
 
     def save(self, data: ir.Table) -> None:
-        save_path = self._get_save_path()
-        Path(save_path).parent.mkdir(parents=True, exist_ok=True)
+        save_path = self._fs_prefix + str(self._get_save_path())
+        if not self._fs_prefix:  # only local paths need their parent created
+            Path(save_path).parent.mkdir(parents=True, exist_ok=True)
         writer = getattr(self.connection, f"to_{self._file_format}")
-        writer(data, str(save_path), **self._save_args)
+        writer(data, save_path, **self._save_args)
 
     def _describe(self) -> dict[str, Any]:
         return {
-            "filepath": self._filepath,
+            "filepath": self._fs_prefix + str(self._filepath),
             "file_format": self._file_format,
             "table_name": self._table_name,
             "backend": self._connection_config["backend"],
@@ -187,4 +215,4 @@ class FileDataset(ConnectionMixin, AbstractVersionedDataset[ir.Table, ir.Table])
         except DatasetError:
             return False
 
-        return Path(load_path).exists()
+        return self._exists_function(str(load_path))
--- a/kedro-datasets/tests/ibis/test_file_dataset.py
+++ b/kedro-datasets/tests/ibis/test_file_dataset.py
@@ -95,6 +95,64 @@ class TestFileDataset:
         reloaded = ds.load()
         assert_frame_equal(dummy_table.execute(), reloaded.execute())
 
+    @pytest.mark.parametrize(
+        "filepath",
+        [
+            "s3://bucket/data/cars.parquet",
+            "abfss://container@account.dfs.core.windows.net/data/cars.parquet",
+            "hf://datasets/user/repo/cars.parquet",
+        ],
+    )
+    def test_load_save_preserve_remote_uri(self, mocker, filepath, dummy_table):
+        """Cloud/remote URIs must reach Ibis intact; the protocol `//` should
+        not be collapsed by ``PurePosixPath`` (see #1298, #918)."""
+        mocker.patch.dict(FileDataset._connections, clear=True)
+        backend = mocker.patch("ibis.duckdb")
+        ds = FileDataset(
+            filepath=filepath,
+            file_format="parquet",
+            connection={"backend": "duckdb"},
+        )
+        connection = backend.connect.return_value
+
+        ds.load()
+        connection.read_parquet.assert_called_once_with(filepath, table_name=None)
+
+        ds.save(dummy_table)
+        connection.to_parquet.assert_called_once_with(dummy_table, filepath)
+
+    def test_remote_exists_uses_filesystem(self, mocker):
+        """`exists` on a remote path should go through ``fsspec`` with a
+        protocol-less path."""
+        mocker.patch.dict(FileDataset._connections, clear=True)
+        mocker.patch("ibis.duckdb")
+        filesystem = mocker.patch("fsspec.filesystem")
+        filesystem.return_value.exists.return_value = True
+        ds = FileDataset(
+            filepath="s3://bucket/data/cars.parquet",
+            file_format="parquet",
+            connection={"backend": "duckdb"},
+        )
+        assert ds.exists()
+        filesystem.return_value.exists.assert_called_once_with(
+            "bucket/data/cars.parquet"
+        )
+
+    def test_remote_unversioned_does_not_build_filesystem(self, mocker, dummy_table):
+        """Unversioned remote IO is delegated to Ibis, so no ``fsspec``
+        filesystem (and therefore no extra dependency like ``s3fs``) is needed."""
+        mocker.patch.dict(FileDataset._connections, clear=True)
+        mocker.patch("ibis.duckdb")
+        filesystem = mocker.patch("fsspec.filesystem")
+        ds = FileDataset(
+            filepath="s3://bucket/data/cars.parquet",
+            file_format="parquet",
+            connection={"backend": "duckdb"},
+        )
+        ds.save(dummy_table)
+        ds.load()
+        filesystem.assert_not_called()
+
     @pytest.mark.parametrize("load_args", [{"filename": True}], indirect=True)
     def test_load_extra_params(self, file_dataset, load_args, dummy_table):
         """Test overriding the default load arguments."""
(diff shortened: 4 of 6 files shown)


kedro-datasets/kedro_datasets/_utils/file_utils.py
```diff
@@ -0,0 +1,24 @@
+"""Filesystem path utilities shared across datasets."""
+from __future__ import annotations
+
+import os
+
+
+def split_filepath(filepath: str | os.PathLike) -> tuple[str, str]:
```
deepyaman, line 7: Just want to be sure I understand--is this absolutely necessary? Is this not covered by `get_protocol_and_path` or `_parse_filepath` in Kedro?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -118,6 +121,10 @@ def __init__(  # noqa: PLR0913
                 `read_{file_format}` method.
             save_args: Additional arguments passed to the Ibis backend's
                 `to_{file_format}` method.
+            fs_args: Extra arguments (e.g. credentials) for the ``fsspec``
```
deepyaman: `fs_args` usually has a very consistent docstring (and usage) across backends; is this docstring very aligned? Didn't check, but making sure.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -130,12 +137,19 @@ def __init__(  # noqa: PLR0913
         _connection_config = connection or self.DEFAULT_CONNECTION_CONFIG
         _credentials = deepcopy(credentials) or {}
         self._connection_config = {**_connection_config, **_credentials}
+        self._fs_args = deepcopy(fs_args or {})
         self.metadata = metadata
 
+        # Keep the protocol prefix (e.g. "s3://") separate; storing it in a
+        # PurePosixPath would collapse the "//". It's reattached for Ibis.
+        self._fs_prefix, path = split_filepath(filepath)
+        self._fs: fsspec.AbstractFileSystem | None = None  # built lazily
```
deepyaman: Is the comment necessary? Is it normal to comment in these circumstances across the codebase?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -130,12 +137,19 @@ def __init__(  # noqa: PLR0913
         _connection_config = connection or self.DEFAULT_CONNECTION_CONFIG
         _credentials = deepcopy(credentials) or {}
         self._connection_config = {**_connection_config, **_credentials}
+        self._fs_args = deepcopy(fs_args or {})
         self.metadata = metadata
 
+        # Keep the protocol prefix (e.g. "s3://") separate; storing it in a
```
deepyaman: Is this and the following explanatory comment necessary? Is it explained in these situations across the codebase usually?

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -130,12 +137,19 @@ def __init__(  # noqa: PLR0913
...
+        # PurePosixPath would collapse the "//". It's reattached for Ibis.
+        self._fs_prefix, path = split_filepath(filepath)
+        self._fs: fsspec.AbstractFileSystem | None = None  # built lazily
+
         super().__init__(
-            filepath=PurePosixPath(filepath),
+            filepath=PurePosixPath(path),
             version=version,
-            exists_function=lambda filepath: Path(filepath).exists(),
+            exists_function=self._fs_exists if self._fs_prefix else None,
```
deepyaman, line 151: Is the `None` case expected? Should we ever hit it? Would it behave correctly, if so?

Doesn't seem there was that None case before, so, again, looking for consistency with how other backends handle fsspec paths.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -159,20 +173,33 @@ def connection(self) -> BaseBackend:
...
     def load(self) -> ir.Table:
-        load_path = str(self._get_load_path())
+        load_path = self._fs_prefix + str(self._get_load_path())
         reader = getattr(self.connection, f"read_{self._file_format}")
         return reader(load_path, table_name=self._table_name, **self._load_args)
 
     def save(self, data: ir.Table) -> None:
-        save_path = self._get_save_path()
-        Path(save_path).parent.mkdir(parents=True, exist_ok=True)
+        save_path = self._fs_prefix + str(self._get_save_path())
```
deepyaman, line 195: Doesn't one of those aforementioned Kedronutils (`get_protocol_and_path`?) already give you stuff in this format? Are we reinventing the wheel, or is this truly different?

kedro-datasets/RELEASE.md
```diff
@@ -7,6 +7,7 @@
 - Repaired `polars.PolarsDatabaseDataset` end-to-end and added a full test suite for it.
 - Fixed `opik.TraceDataset` so `credentials.project_name` is now passed to `configure()` and persisted to Opik's session configuration.
 - Added `os.PathLike` support for `Spark` datasets.
+- Fixed `ibis.FileDataset` to preserve remote URIs (e.g. `s3://`, `gcs://`, `abfss://`, `hf://`, `https://`) instead of collapsing the protocol's `//`, so cloud and web filepaths now work end-to-end. Added an `fs_args` argument to authenticate the `fsspec` filesystem used for version discovery and existence checks on remote paths. Note: `ibis.FileDataset`'s string representation now shows the full URI again, and trailing slashes in paths are normalized away (e.g. `s3://bucket/dir/` becomes `s3://bucket/dir`).
```
deepyaman: Is this not very verbose? Are entries usually this detailed? Is it justifiable?

deepyaman (commented):
Please see comments.

Also, just want to confirm this _is_ the right way to resolve the issues (i.e. now using PyArrow filesystems or something? Had that idea before...).

kedro-datasets/kedro_datasets/_utils/file_utils.py
deepyaman: Yes, unfortunately. `get_protocol_and_path`/`get_filepath_str` are built for fsspec, which takes protocol-*less* paths plus a filesystem object — so `get_filepath_str` strips the protocol for everything except http(s). Worse, the parsing is lossy for protocols not in `CLOUD_PROTOCOLS`: e.g. `hf://datasets/maharshipandya/.../dataset.csv` (the #918 case) round-trips to `hf:///maharshipandya/.../dataset.csv` — the `datasets` host segment is dropped and the slashes are wrong. Ibis needs the full, untouched URI, so I reuse `split_filepath` — the same dependency-free helper `spark.SparkDataset` already uses for this exact problem. I only moved it here so importing it does not drag in pyspark (`databricks_utils` imports `pyspark` at module top); it is re-exported there for backwards compatibility.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Good catch — it was not. Aligned the opening with the standard wording (e.g. `pandas.CSVDataset`): "Extra arguments to pass into underlying filesystem class constructor (e.g. ...)." I kept a short Ibis-specific clarification because we deliberately drop the `open_args_load`/`open_args_save` half of the usual `fs_args` contract — Ibis does the actual I/O, so the filesystem is only used for version discovery / existence checks.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: No — removed. The `_filesystem()` method makes the laziness self-evident.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Removed for consistency — `SparkDataset` does the identical `fs_prefix, filepath = split_filepath(filepath)` with no comment, so this now matches.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Yes, expected — and hit for every local path. Passing `None` tells `AbstractVersionedDataset` to fall back to its defaults (`_local_exists` / `iglob`), which is identical to the previous local behavior (the old explicit lambda was just `Path(filepath).exists()`, i.e. `_local_exists`). This mirrors `SparkDataset`, which initializes `exists_function = glob_function = None` and only overrides them for remote prefixes, leaving local paths on the kedro defaults.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Same as the `split_filepath` thread: `get_filepath_str` returns a protocol-*less* path (for fsspec to consume with a filesystem object), not the native URI Ibis needs. So I reattach the prefix — `self._fs_prefix + str(self._get_save_path())` — which is exactly what `SparkDataset.load`/`save`/`_exists` do.

kedro-datasets/RELEASE.md
deepyaman: Agreed — trimmed to a single line matching the neighboring entries. (Dropped the repr / trailing-slash notes; they are cosmetic edge cases, not worth a release bullet.)

deepyaman:
Good question — I looked specifically into PyArrow filesystems (your Option 2 from #1298), and I'm confident fsspec is the right call here.

The key framing: the filesystem library is **only ever used for Kedro's version discovery (`glob`) and `exists()` — never for data I/O**, which Ibis always does natively (and authenticates itself via `connection`/`credentials`). So the only question that matters is which library does glob + exists across the protocols we care about. On that narrow question, `pyarrow.fs` falls short on two dealbreakers for *these* issues:

1. **No glob.** `pyarrow.fs.FileSystem` exposes only `get_file_info()` + `FileSelector` (listing under a base dir) — no wildcard glob. Kedro's `_fetch_latest_load_version` calls `glob_function(".../*/...")`, so we'd have to hand-roll glob-pattern matching on top of `FileSelector`. (This is exactly the uncertainty you flagged in #1298 — confirmed against pyarrow 20: it doesn't.)

2. **No `hf://`.** `pyarrow.fs` only knows Local/S3/GCS/HDFS/Azure; `FileSystem.from_uri("hf://...")` raises *"Unrecognized filesystem type."* That's literally #918. fsspec covers `hf` (via `huggingface_hub`), plus `s3`/`gcs`/`abfss`/`https`/`file`, and has a native `glob`.

Everything else favors fsspec or is neutral:
- **Consistency** — every other Kedro dataset uses fsspec, and `spark.SparkDataset` (the closest analog: engine does native I/O, fsspec only for versioning) uses this exact `split_filepath` + lazy-`fsspec` fallback. This PR is the same pattern.
- **Dependency** — neutral: both are already present (fsspec via kedro core, pyarrow via Ibis), so "pyarrow is already a dep" doesn't tip it.
- **Credentials** — fsspec's `fs_args`/storage-options is the idiom Kedro users already know.

On Option 3 (start with fsspec, migrate to pyarrow.fs later): I don't think there's a migration to make — glob + broad protocol coverage (incl. `hf`) are fsspec's strengths and pyarrow.fs's gaps, so if anything the bridge runs the other way (`pyarrow.fs.PyFileSystem(FSSpecHandler(...))` wraps fsspec, not vice-versa). And since data I/O is Ibis's job regardless, there's no arrow-native read path we'd be forgoing.

tl;dr: fsspec, scoped to version discovery only, matching `SparkDataset`. PyArrow filesystems would mean reimplementing glob and still not solving #918.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -159,20 +172,33 @@ def connection(self) -> BaseBackend:
         """The ``Backend`` instance for the connection configuration."""
         return self._connection
 
+    def _filesystem(self) -> fsspec.AbstractFileSystem:
+        # Build lazily so backend-only users don't need adlfs, s3fs, etc.
+        if self._fs is None:
+            self._fs = fsspec.filesystem(self._fs_prefix.strip("://"), **self._fs_args)
```
SajidAlamQB: nit: `.strip("://")` strips the char `set {:,/}` from both ends rather than the :// suffix. Works for every real protocol so its probably fine, but it would mangle chained ones.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
```diff
@@ -118,6 +121,11 @@ def __init__(  # noqa: PLR0913
                 `read_{file_format}` method.
             save_args: Additional arguments passed to the Ibis backend's
                 `to_{file_format}` method.
+            fs_args: Extra arguments to pass into underlying filesystem class
```
SajidAlamQB, line 124: Maybe a line here noting that fsspec creds for version discovery go in fs_args, not credentials, since credentials is merged into the Ibis backend connection, not passed to fsspec. 

Other fsspec datasets do `{**credentials, **fs_args}`, so someone following that pattern is easy to trip on.

SajidAlamQB (commented):
Nice @deepyaman , this looks good to me I traced it through and the core fix is solid.

Just left some minor nits otherwise.

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Good catch — switched to `self._fs_prefix.removesuffix("://")`. `split_filepath` guarantees the prefix is exactly `<protocol>://` (it returns `split_[0] + "://"`), so removing that suffix is exact and sidesteps the chained-`://` concern entirely — no `maxsplit` reasoning needed. (The `.strip("://")` was carried over from `SparkDataset`, which has the same latent imprecision; happy to send a follow-up there too.)

kedro-datasets/kedro_datasets/ibis/file_dataset.py
deepyaman: Added a brief note to the `fs_args` docstring: "Note that fsspec credentials go here, not in `credentials` (which configures the Ibis connection)." That should steer anyone expecting the usual `{**credentials, **fs_args}` pattern from the other fsspec datasets.

kedro-datasets/tests/ibis/test_file_dataset.py
```diff
@@ -95,6 +95,64 @@ def test_save_and_load_parquet(
         reloaded = ds.load()
         assert_frame_equal(dummy_table.execute(), reloaded.execute())
 
+    @pytest.mark.parametrize(
+        "filepath",
+        [
+            "s3://bucket/data/cars.parquet",
```
ElenaKhaustova, line 101: The reporter in #918 also hit this with `https://` URLs (reading from `raw.githubusercontent.com`). Adding one to the parametrize list would give explicit coverage of that scenario.

ElenaKhaustova (commented):
The fix looks good to me!

Left one optional nit below.

ElenaKhaustova (approved):

SajidAlamQB (approved):

kedro-datasets/tests/ibis/test_file_dataset.py
deepyaman: Sorry, I'd set this to auto-merge; in any case, `https://` doesn't exercise a different path; I did double-check the end-to-end and post it as a comment to #918 for good measure!

PR #1441, opened 2026-06-29
fix(datasets): use removesuffix for Spark protocol

deepyaman:
## Description

`SparkDataset` derives the fsspec protocol from the filepath prefix with
`fs_prefix.strip("://")`. `str.strip` removes the *character set* `{":", "/"}`
from both ends rather than the `://` suffix — it happens to work for every real
protocol, but is imprecise and would mangle an unusual one. Since `split_filepath`
guarantees the prefix ends in exactly `://`, `removesuffix("://")` drops precisely
that suffix.

Behavior is identical for all real protocols, so this is effectively a
robustness/clarity change (not worth a release note).

Follow-up to #1436, which applied the same change to `ibis.FileDataset`
([review thread](https://github.com/kedro-org/kedro-plugins/pull/1436#discussion_r3492364694)).

## Development notes

One-line change; no behavior change for any real protocol. Existing `SparkDataset`
tests cover the filesystem-construction path.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset.py
@@ -240,7 +240,7 @@ class SparkDataset(AbstractVersionedDataset[DataFrame, DataFrame]):
                 glob_function = partial(dbfs_glob, dbutils=dbutils)
                 exists_function = partial(dbfs_exists, dbutils=dbutils)
         else:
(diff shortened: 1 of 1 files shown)


SajidAlamQB (approved):
👍

ankatiyar (approved):

PR #1444, opened 2026-07-01
feat(datasets): Add WeaviateVectorStoreDataset

lrcouto:
## Description

Implements `WeaviateVectorStoreDataset` and `WeaviateVectorStoreHandle`, as the first concrete backend for the `AbstractVectorStoreDataset` abstraction introduced in https://github.com/kedro-org/kedro-plugins/pull/1430

Part of https://github.com/kedro-org/kedro-plugins/issues/1351

---

### `WeaviateVectorStoreDataset`

A Kedro dataset that connects to a [Weaviate](https://weaviate.io/) collection and returns a `WeaviateVectorStoreHandle` on `load()`. Three connection modes:

- **`local`** (default) — connects to a locally running Weaviate instance
- **`cloud`** — connects to Weaviate Cloud; requires `url` and optionally `credentials: {api_key: ...}`
- **`custom`** — passes `connection_params` directly to `weaviate.connect_to_custom()` for self-hosted deployments

`save()` is intentionally disabled and raises `DatasetError` — vector stores are not write-once filesystems, so all mutations go through the handle.

### `WeaviateVectorStoreHandle`

Returned by `load()`. Owns the gRPC connection; callers must close it either explicitly or via the context manager:

```python
  with catalog.load("my_store") as store:
      ids = store.add([{"properties": {"text": "..."}, "vector": [...]}])
      results = store.search(vector=[...], top_k=5)
      store.delete(ids=ids)
```

**`add(records)`** — inserts a batch of records via `collection.data.insert_many`. Each record is a plain dict; "vector" and "id" are reserved keys (popped before storage). All other keys become Weaviate object properties. Returns a list of UUID strings in insertion order. Raises `DatasetError` on any failure.

**`delete(*, ids=None, filters=None)`** — exactly one of vector or text required. vector triggers near_vector; text triggers near_text (requires a server-side vectorizer on the collection). Returns a list of dicts, each containing all stored properties plus "id" (UUID string) and "distance" (float). "id" and "distance" always take precedence over same-named properties.

**`search(*, vector=None, text=None, top_k=10, filters=None)`** — exactly one of `vector` or `text` required. `vector` triggers `near_vector`; `text` triggers `near_text`. Returns a list of dicts, each containing all stored properties plus `"id"` (UUID string) and `"distance"` (float). `"id"` and `"distance"` always take precedence over same-named properties.

**`describe()`** — returns collection name and current object count.

**`raw_client`** — escape hatch exposing the underlying `weaviate.WeaviateClient` for operations outside this interface.

---

## Design decisions

**`load()` returns a handle, not data.** Vector stores are stateful connections, not serialisable blobs. The handle owns the connection lifecycle, which maps to Kedro's `load()` contract without requiring a custom runner or session hook.

**`save()` raises unconditionally.** Overriding `save()` directly (not `_save()`) sidesteps the metaclass alias and makes the intent explicit: vector store mutations go through the handle, not the catalog write path. Explained in more detail in https://github.com/kedro-org/kedro-plugins/pull/1430

**Backend-native filters.** `search()` and `delete()` accept `weaviate.classes.query.Filter` objects directly.

**`get_or_create` not available in weaviate-client ≥4.6.** The `_load()` implementation uses `collections.exists()` + `create()`/`get()` rather than the non-existent `get_or_create()`.

## Testing

Besides the unit tests included in the PR, this vector store dataset can be tested using the [GraphRAG ](https://github.com/lrcouto/graphrag/tree/weaviate-backend-test) demo. A separate branch has this dataset used as the vector store for the project instead of ChromaDB. Both of them should work identically.

Needs a Python env with `weaviate-client>=4.0`, `openai`, `kedro`, and a [Weaviate Cloud cluster](https://console.weaviate.cloud) (free tier works fine).

```bash
# Add OpenAI and Weaviate credentials to conf/local/credentials.yml:
# openai:
#   api_key: "<your-key>"
# weaviate_cloud:
#   api_key: "<your-key>"
# And set the cluster URL in conf/base/catalog.yml

# Build vector index (embeds 18 documents into Weaviate via add())
kedro run --pipelines vector_indexing

# Run agent Q&A (queries Weaviate via search())
kedro run --pipelines query_answering

# Run everything end-to-end
kedro run

# Run everything through the Streamlit app
streamlit run app/streamlit_app.py         
```

The catalog entry used in the demo:

```yaml
weaviate_collection:
  type: graphrag.datasets.weaviate_vector_store_dataset.WeaviateVectorStoreDataset
  collection_name: HealthcareKnowledge
  connection_type: cloud
  url: "https://<cluster>.weaviate.cloud"
  credentials: weaviate_cloud
```

## Checklist

### DIFF
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -5,7 +5,7 @@
 
 | Type                     | Description                                      | Location                           |
 | ------------------------ | ------------------------------------------------ | ---------------------------------- |
-| `weaviate.WeaviateVectorStoreDataset` | A dataset for loading and saving data to Weaviate vector database collections. | `kedro_datasets_experimental.weaviate` |
+| `weaviate.WeaviateVectorStoreDataset` | A dataset that loads a handle for adding, searching, and deleting entries in Weaviate vector database collections. | `kedro_datasets_experimental.weaviate` |
 
 ## Bug fixes and other changes
 ## Community contributions
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,12 +1,18 @@
 # Upcoming release
 ## Major features and improvements
+* Added `vectorstore_base.AbstractVectorStoreDataset` and `vectorstore_base.VectorStoreHandle`, backend-agnostic abstract base classes for vector store datasets.
+- Added the following new **experimental** datasets:
+
+| Type                     | Description                                      | Location                           |
+| ------------------------ | ------------------------------------------------ | ---------------------------------- |
+| `weaviate.WeaviateVectorStoreDataset` | A dataset for loading and saving data to Weaviate vector database collections. | `kedro_datasets_experimental.weaviate` |
+
 ## Bug fixes and other changes
 ## Community contributions
 
 # Release 9.5.0
 
 ## Major features and improvements
-* Added `vectorstore_base.AbstractVectorStoreDataset` and `vectorstore_base.VectorStoreHandle`, backend-agnostic abstract base classes for vector store datasets.
 * Added `send_individually` option to `APIDataset` to send list items as individual requests instead of batched arrays.
 
 ## Bug fixes and other changes
--- a/kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
@@ -26,14 +26,15 @@ MODULE = "kedro_datasets_experimental.weaviate.weaviate_vector_store_dataset"
 @pytest.fixture
 def mock_client():
     client = MagicMock()
-    client.collections.get_or_create.return_value = MagicMock(name="MyCollection")
+    client.collections.exists.return_value = False
+    client.collections.create.return_value = MagicMock(name="MyCollection")
     client.collections.get.return_value = MagicMock(name="MyCollection")
     return client
 
 
 @pytest.fixture
 def mock_collection(mock_client):
-    collection = mock_client.collections.get_or_create.return_value
+    collection = mock_client.collections.create.return_value
     collection.name = "MyCollection"
     agg_result = MagicMock()
     agg_result.total_count = 42
@@ -231,12 +232,23 @@ class TestConnect:
 
 
 class TestLoad:
-    def test_load_returns_handle(self, mock_client, mock_collection):
+    def test_load_creates_collection_when_missing(self, mock_client, mock_collection):
+        mock_client.collections.exists.return_value = False
         with patch(f"{MODULE}.weaviate.connect_to_local", return_value=mock_client):
             ds = WeaviateVectorStoreDataset(collection_name="MyCollection")
             handle = ds._load()
             assert isinstance(handle, WeaviateVectorStoreHandle)
-            mock_client.collections.get_or_create.assert_called_once_with("MyCollection")
+            mock_client.collections.exists.assert_called_once_with("MyCollection")
+            mock_client.collections.create.assert_called_once_with("MyCollection")
+
+    def test_load_gets_existing_collection(self, mock_client, mock_collection):
+        mock_client.collections.exists.return_value = True
+        with patch(f"{MODULE}.weaviate.connect_to_local", return_value=mock_client):
+            ds = WeaviateVectorStoreDataset(collection_name="MyCollection")
+            handle = ds._load()
+            assert isinstance(handle, WeaviateVectorStoreHandle)
+            mock_client.collections.get.assert_called_once_with("MyCollection")
+            mock_client.collections.create.assert_not_called()
 
     def test_load_get_only_when_no_create(self, mock_client):
         with patch(f"{MODULE}.weaviate.connect_to_local", return_value=mock_client):
@@ -246,10 +258,10 @@ class TestLoad:
             )
             ds._load()
             mock_client.collections.get.assert_called_once_with("MyCollection")
-            mock_client.collections.get_or_create.assert_not_called()
+            mock_client.collections.exists.assert_not_called()
 
     def test_collection_error_closes_client_and_raises(self, mock_client):
-        mock_client.collections.get_or_create.side_effect = RuntimeError("not found")
+        mock_client.collections.exists.side_effect = RuntimeError("not found")
         with patch(f"{MODULE}.weaviate.connect_to_local", return_value=mock_client):
             ds = WeaviateVectorStoreDataset(collection_name="Missing")
             with pytest.raises(DatasetError, match="Failed to access Weaviate collection"):
--- a/kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
+++ b/kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
@@ -311,14 +311,93 @@ class TestHandle:
         assert result == {"collection": "MyCollection", "count": 42}
         mock_collection.aggregate.over_all.assert_called_once_with(total_count=True)
 
-    def test_add_raises_not_implemented(self, handle):
-        with pytest.raises(NotImplementedError, match="ST3"):
-            handle.add([])
-
-    def test_delete_raises_not_implemented(self, handle):
-        with pytest.raises(NotImplementedError, match="ST3"):
-            handle.delete(ids=["a"])
-
     def test_search_raises_not_implemented(self, handle):
         with pytest.raises(NotImplementedError, match="ST4"):
             handle.search(vector=[0.1, 0.2])
+
+
+# ---------------------------------------------------------------------------
+# WeaviateVectorStoreHandle — add()
+# ---------------------------------------------------------------------------
+
+
+class TestHandleAdd:
+    @pytest.fixture
+    def handle(self, mock_client, mock_collection):
+        return WeaviateVectorStoreHandle(mock_client, mock_collection)
+
+    @pytest.fixture
+    def insert_result(self):
+        import uuid
+        result = MagicMock()
+        result.uuids = {0: uuid.UUID("aaaaaaaa-0000-0000-0000-000000000001"),
+                        1: uuid.UUID("bbbbbbbb-0000-0000-0000-000000000002")}
+        result.errors = {}
+        return result
+
+    def test_add_calls_insert_many(self, handle, mock_collection, insert_result):
+        mock_collection.data.insert_many.return_value = insert_result
+        records = [
+            {"vector": [0.1, 0.2], "text": "hello"},
+            {"vector": [0.3, 0.4], "text": "world"},
+        ]
+        handle.add(records)
+        mock_collection.data.insert_many.assert_called_once()
+        objects = mock_collection.data.insert_many.call_args[0][0]
+        assert len(objects) == 2
+        assert objects[0].properties == {"text": "hello"}
+        assert objects[0].vector == [0.1, 0.2]
+        assert objects[1].properties == {"text": "world"}
+
+    def test_add_returns_uuid_strings(self, handle, mock_collection, insert_result):
+        mock_collection.data.insert_many.return_value = insert_result
+        uuids = handle.add([{"vector": [0.1], "text": "a"}, {"vector": [0.2], "text": "b"}])
+        assert uuids == [
+            "aaaaaaaa-0000-0000-0000-000000000001",
+            "bbbbbbbb-0000-0000-0000-000000000002",
+        ]
+
+    def test_add_passes_optional_id(self, handle, mock_collection, insert_result):
+        mock_collection.data.insert_many.return_value = insert_result
+        handle.add([{"id": "my-uuid", "vector": [0.1], "text": "x"}])
+        obj = mock_collection.data.insert_many.call_args[0][0][0]
+        assert obj.uuid == "my-uuid"
+        assert "id" not in obj.properties
+
+    def test_add_raises_on_partial_errors(self, handle, mock_collection):
+        result = MagicMock()
+        result.uuids = {}
+        result.errors = {0: "connection reset"}
+        mock_collection.data.insert_many.return_value = result
+        with pytest.raises(DatasetError, match="add\\(\\) failed for 1 record"):
+            handle.add([{"vector": [0.1], "text": "bad"}])
+
+
+# ---------------------------------------------------------------------------
+# WeaviateVectorStoreHandle — delete()
+# ---------------------------------------------------------------------------
+
+
+class TestHandleDelete:
+    @pytest.fixture
+    def handle(self, mock_client, mock_collection):
+        return WeaviateVectorStoreHandle(mock_client, mock_collection)
+
+    def test_delete_by_ids_calls_delete_by_id(self, handle, mock_collection):
+        handle.delete(ids=["uuid-1", "uuid-2"])
+        assert mock_collection.data.delete_by_id.call_count == 2
+        mock_collection.data.delete_by_id.assert_any_call("uuid-1")
+        mock_collection.data.delete_by_id.assert_any_call("uuid-2")
+
+    def test_delete_by_filter_calls_delete_many(self, handle, mock_collection):
+        f = MagicMock()
+        handle.delete(filters=f)
+        mock_collection.data.delete_many.assert_called_once_with(where=f)
+
+    def test_delete_requires_ids_or_filters(self, handle):
+        with pytest.raises(DatasetError, match="requires exactly one"):
+            handle.delete()
+
+    def test_delete_rejects_both(self, handle):
+        with pytest.raises(DatasetError, match="not both"):
+            handle.delete(ids=["x"], filters=MagicMock())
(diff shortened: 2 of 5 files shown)


ElenaKhaustova:
@lrcouto, could you please change the base to https://github.com/kedro-org/kedro-plugins/pull/1430? So this PR only has changes related to the newly added dataset.

kedro-datasets/RELEASE.md
```diff
@@ -1,6 +1,12 @@
 # Upcoming release
 ## Major features and improvements
 * Added `vectorstore_base.AbstractVectorStoreDataset` and `vectorstore_base.VectorStoreHandle`, backend-agnostic abstract base classes for vector store datasets.
+- Added the following new **experimental** datasets:
+
+| Type                     | Description                                      | Location                           |
+| ------------------------ | ------------------------------------------------ | ---------------------------------- |
+| `weaviate.WeaviateVectorStoreDataset` | A dataset for loading and saving data to Weaviate vector database collections. | `kedro_datasets_experimental.weaviate` |
```
ankatiyar: just loading?

lrcouto: I copied the description from the chromadb dataset, but you're right, it doesn't save. Maybe we can write it as something like " A dataset that loads a handle for adding, searching, and deleting entries in Weaviate vector database collections."

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,361 @@
...
+            List of UUID strings for the inserted objects, in the same order
+            as the input records.
+
+        Raises:
+            DatasetError: If the batch insert call fails, or if Weaviate
+                returns errors for one or more objects.
+        """
+        objects = []
+        for record in records:
+            props = record.copy()
```
ElenaKhaustova: **Record format doesn't match the ABC contract.**

The ABC's `add()` docstring (`vector_store_handle.py`, line 27-33) documents this input format:

```python
handle.add([{
    "properties": {"text": "hello"},  # required, nested dict
    "vector": [0.1, 0.2],            # optional
    "id": "some-uuid",               # optional
}])
```

But this implementation treats the record as a flat dict — pops `"vector"` and `"id"`, everything else becomes properties:

```python
props = record.copy()
vector = props.pop("vector", None)
uid = props.pop("id", None)
objects.append(DataObject(properties=props, uuid=uid, vector=vector))
```

This means a user following the ABC docs and passing `{"properties": {"text": "hello"}, "vector": [...]}` would get a Weaviate property literally named `"properties"` — not what they intended.

The nested format is the right choice for the ABC because it guarantees portability: user properties are namespaced under `"properties"` and can never collide with reserved keys (which may differ across backends). The implementation here should unwrap accordingly:

```python
props = record.get("properties", {})
vector = record.get("vector")
uid = record.get("id")
objects.append(DataObject(properties=props, uuid=uid, vector=vector))
```

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,361 @@
...
+
+        try:
+            if vector is not None:
+                results = self._collection.query.near_vector(near_vector=vector, **common_kwargs)
+            else:
+                results = self._collection.query.near_text(query=text, **common_kwargs)
+        except Exception as e:
+            raise DatasetError(f"search() failed: {e}") from e
+
+        return [
```
ElenaKhaustova, line 286: **Same mismatch on the output side.**

The ABC's `search()` docs (`vector_store_handle.py`, line 64-66) say results contain `"id"` (str) and `"properties"` (dict), with optional `"distance"` and `"vector"`. But this returns flat dicts with properties spread at the top level:

```python
return [
    {**obj.properties, "id": str(obj.uuid), "distance": obj.metadata.distance}
    for obj in results.objects
]
```

Code written against the ABC — `results[0]["properties"]["text"]` — would `KeyError`.

This should wrap properties to match the contract:

```python
return [
    {
        "id": str(obj.uuid),
        "properties": dict(obj.properties),
        "distance": obj.metadata.distance,
    }
    for obj in results.objects
]
```

This also eliminates the key collision issue where `"id"` or `"distance"` could silently override same-named user properties.

kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,513 @@
...
+    def insert_result(self):
+        result = MagicMock()
+        result.uuids = {0: uuid.UUID("aaaaaaaa-0000-0000-0000-000000000001"),
+                        1: uuid.UUID("bbbbbbbb-0000-0000-0000-000000000002")}
+        result.errors = {}
+        return result
+
+    def test_add_calls_insert_many(self, handle, mock_collection, insert_result):
+        mock_collection.data.insert_many.return_value = insert_result
+        records = [
```
ElenaKhaustova, line 309: The tests should be updated to match the nested record format. For example in `test_add_calls_insert_many`:

```python
# Current (flat):
records = [
    {"vector": [0.1, 0.2], "text": "hello"},
    {"vector": [0.3, 0.4], "text": "world"},
]

# Should be (nested):
records = [
    {"properties": {"text": "hello"}, "vector": [0.1, 0.2]},
    {"properties": {"text": "world"}, "vector": [0.3, 0.4]},
]
```

And for `test_search_returns_formatted_results`, the expected output should contain a `"properties"` dict rather than flat keys.

The test `test_search_id_and_distance_override_same_named_properties` can be removed — with the nested format, there's no collision risk since user properties live under `"properties"`.

ElenaKhaustova (changes requested):
Thank you, @lrcouto! Good first concrete implementation. The abstraction holds up well structurally, but there's a record format mismatch with the ABC contract that needs fixing.

lrcouto:
Thank you @ElenaKhaustova ! The format should be fixed now. Updated the PR description and the project example as well.

kedro-datasets/pyproject.toml
```diff
@@ -254,6 +254,9 @@ safetensors = ["kedro-datasets[safetensors-safetensorsdataset]"]
 video-videodataset = ["opencv-python~=4.13.0.92"]
 video = ["kedro-datasets[video-videodataset]"]
 
+vectorstore-weaviatevectorstoredataset = ["weaviate-client>=4.0"]
+vectorstore-weaviate = ["kedro-datasets[vectorstore-weaviatevectorstoredataset]"]
```
merelcht: Nit: this is a super long name, can we just call use `weaviate-dataset` and `weaviate`? Similar to pytorch

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+
+class WeaviateVectorStoreHandle(VectorStoreHandle):
+    """Handle for interacting with a Weaviate collection.
+
+    Returned by ``WeaviateVectorStoreDataset.load()``. Owns the gRPC connection
+    to Weaviate; the connection must be closed after use, either by calling
+    ``close()`` explicitly or by using the handle as a context manager::
+
+        with catalog.load("my_store") as store:
+            store.describe()
```
merelcht, line 42: I'm assuming the connection will be closed automatically when not instantiating a catalog yourself as in this example? Might be good to clarify you only need to manually handle closing if you're instantiating catalog & this dataset instead of just letting Kedro do it for you.

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+        try:
+            if ids is not None:
+                for uid in ids:
+                    self._collection.data.delete_by_id(uid)
+            else:
+                self._collection.data.delete_many(where=filters)
+        except Exception as e:
+            raise DatasetError(f"delete() failed: {e}") from e
+
+    # --- read path (ST4) ---
```
merelcht: What does ST4 mean?

merelcht (commented):
I left some questions, but generally looks good to me!

Can you explain how it works with connecting to the collection? I see you just have to pass a str with the collection name, but how does that actually work? How do you create a collection in the first place? Might be helpful to at least link to the weaviate docs on collections, so users can understand what they need to do to use this dataset.

kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,510 @@
+"""Tests for WeaviateVectorStoreDataset and WeaviateVectorStoreHandle (ST2).
```
SajidAlamQB: What do these ST2 mean? Maybe we can drop them

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
lrcouto: This is a leftover from my own task breakdown, it's "step 4" of the implementation. Should not have been committed.

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,365 @@
...
+        """
+        if ids is None and filters is None:
+            raise DatasetError("delete() requires exactly one of 'ids' or 'filters'.")
+        if ids is not None and filters is not None:
+            raise DatasetError("delete() accepts 'ids' or 'filters', not both.")
+
+        try:
+            if ids is not None:
+                for uid in ids:
+                    self._collection.data.delete_by_id(uid)
```
SajidAlamQB:
```suggestion
                self._collection.data.delete_many(
                    where=wvc.query.Filter.by_id().contains_any(ids)
                )
```
Would it make sense to do something like this? Reading their docs there is a 10k query limitation though, https://docs.weaviate.io/weaviate/manage-objects/delete#delete-multiple-objects-by-id

SajidAlamQB (commented):
Thanks @lrcouto it looks mostly good just left some minor comments this is the template I'll mirror for the ChromaDB refactor too 👍

kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
lrcouto: Same as I told Merel, those are from my own task breakdown system on Obsidian 😅

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
lrcouto: No, you'd need to manually handle closing in all cases. It has to do with how Weaviate python client handles sockets when using the `connect()` function. It has a persistent connection pool so you don't have to re-do all the setup, TCP handshake, etc. every time you want to add something to it, search it, or anything else. IIRC that's regular DB behavior so it makes sense.

Can be seen here: 
https://github.com/weaviate/weaviate-python-client/blob/17a988745cd9f21f84fd9e88e8bc0a4b0aa95e25/weaviate/connect/v4.py#L900

And this connection is taken down using `close()`, which is what we use in the dataset's closing method as well.

https://github.com/weaviate/weaviate-python-client/blob/17a988745cd9f21f84fd9e88e8bc0a4b0aa95e25/weaviate/connect/v4.py#L710-L740

The docstring on both the Weaviate methods that implement this connection and on the WeaviateVectorStoreHandle itself mention that you either need to manually close, or use a context manager to handle it closing when expected. In this case, the Kedro runner will not clean this up because calling `catalog.release()` on this dataset affects the dataset object itself, but not the vector store handle object handed out by `load()`. Because this dataset, [by design](https://github.com/kedro-org/kedro-plugins/issues/1157#issuecomment-4032483936), is kinda stretching out the idea of what a Kedro dataset is. It's not like a CSV or a Json static object, it's a python client that has an active connection to a cloud server/docker container/etc.

TL;DR this one is a bit weird and kinda pushing the limits of being a dataset.

I think we _could_ try to make the dataset itself handle the connection, but here's what I thought:
- This is the the first implementation from the VectorStoreHandle abstraction that we are doing
- It would deviate from the design proposed on the spike
- Might make it too different from other future VectorStore datasets that we are planning.
- This is following how it seems that the Weaviate python client expects you to use it.

So for now maybe we can leave it super well documented and see how it goes with other VectorStore implementations? What do you think?

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
lrcouto: That's a good point! I've added this batch size for the delete method, with a default of 10k (default number for self-hosted usage).

kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,484 @@
...
```
ElenaKhaustova: We still have ST2 here

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,501 @@
...
+        except Exception as e:
+            client.close()
+            raise DatasetError(
+                f"Failed to access Weaviate collection '{self._collection_name}': {e}"
+            ) from e
+        return WeaviateVectorStoreHandle(
+            client, collection, delete_batch_size=self._delete_batch_size
+        )
+
+    def _describe(self) -> dict[str, Any]:
```
ElenaKhaustova, line 495: Nit: `delete_batch_size` is a user-configurable constructor parameter now but isn't included in `_describe()`. Consider adding `"delete_batch_size": self._delete_batch_size` to help with debugging when someone uses a non-default value.

kedro-datasets/kedro_datasets_experimental/tests/weaviate/test_weaviate_vector_store_dataset.py
```diff
@@ -0,0 +1,484 @@
...
+            assert contains_any.call_count == 2
+            assert len(contains_any.call_args_list[0].args[0]) == 10_000
+            assert len(contains_any.call_args_list[1].args[0]) == 1
+            assert mock_collection.data.delete_many.call_count == 2
+
+    def test_delete_by_ids_empty_list_is_noop(self, handle, mock_collection):
+        handle.delete(ids=[])
+        mock_collection.data.delete_many.assert_not_called()
+
+    def test_delete_by_filter_calls_delete_many(self, handle, mock_collection):
```
ElenaKhaustova, line 398: Bug: this test will fail as mock doesn't return a proper result object.

The `delete()` implementation now loops until r`esult.matches == 0`, but an unconfigured `MagicMock `doesn't override `__eq__,` so `mock.matches == 0` is always `False`. The loop exhausts all 1000 iterations and raises `DatasetError`.

To fix it, configure the mock return value:
```python
def test_delete_by_filter_calls_delete_many(self, handle, mock_collection):
    mock_result = MagicMock()
    mock_result.matches = 0
    mock_collection.data.delete_many.return_value = mock_result
    f = MagicMock()
    handle.delete(filters=f)
    mock_collection.data.delete_many.assert_called_once_with(where=f)
```

Also worth adding two tests for the new filter re-run behavior:

1. Re-run loop happy path: first call returns `matches=5`, second returns `matches=0`, verify `delete_many` called twice.
2. Max iterations safety cap: mock always returns `matches=1`, verify `DatasetError` is raised with the "still matched objects" message.

ElenaKhaustova (commented):
@lrcouto, thank you for addressing the comments!

Here are a few other things I found after the second pass.

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
ElenaKhaustova: Agree, this is a usability concern long-term. If all backends need manual closing, we could consider having the dataset track handles it creates and close them via `_release()`. But that's a design decision for the abstraction layer that we should probably defer past this PR as we bring more concrete implementations.

ElenaKhaustova (approved):
LGTM 👍

kedro-datasets/kedro_datasets_experimental/weaviate/weaviate_vector_store_dataset.py
merelcht: Thanks for explaining @lrcouto . That totally makes sense. The note in the dataset docstring looks good 👍

SajidAlamQB (approved):
Thank you @lrcouto looks good!

merelcht (approved):
Looks good, thanks @lrcouto ! I'd still like to understand how it works with the Collection, maybe you can show me on a call 🙂

lrcouto:
> Looks good, thanks @lrcouto ! I'd still like to understand how it works with the Collection, maybe you can show me on a call 🙂

The Weaviate python client handles most of it. The load method on the dataset itself funnels this `_collection_name` string into either of the `get`, `create` or `exists` methods from the client's `_Collection` class.

`get` returns a Collection object that you can interact with, it does not make network requests or anything. `create` will send a request and actually create the collection server side, and then also return a Collection object. And `exists` will send a request and return True or False based on the server's response. 

You can see them here: https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/collections/executor.py

And the Collection iself: https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/collection/sync.py

This fits with the design decision of making these vector store datasets these thin Kedro layers around the client. What the dataset is doing is just selecting which of the Weaviate methods should be used based on whether we need to create this collection on the remote server or not, and then returning a handle to the user so they can interact with it.

PR #1454, opened 2026-07-10
feat(datasets): Add Feast

lvijnck:
## Description
 
Provide Feast integration

## Checklist


kedro-datasets/kedro_datasets_experimental/feast/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""``FeastDataset`` provides a Kedro integration for Feast.
```
lvijnck, line 1: Submitting as code next to the blog post, goal to discuss next steps/framework fit before expanding.

ElenaKhaustova (commented):
Thank you @lvijnck, I've made a quick pass and here are a few things we suggest to fix before this is ready for a full review:

## CI is red across the board

- **DCO fails** — the `update` commit isn't signed off.
- **PR title check fails** — `feast(datasets): Add Feast` uses `feast` as the conventional-commit type, which isn't a valid type (`feat`/`fix`/`docs`/etc.). Should be `feat(datasets): Add Feast`.
- **Lint fails** — ruff reports an error in `feast_dataset.py:55` (the f-string embedded directly in the `raise DatasetError(...)` call).
- **Unit tests didn't complete** (canceled) — there are no tests for the new module, and `feast`/`google-cloud-bigquery` extras aren't wired into `pyproject.toml`, so CI can't even install/exercise it.

## Missing plumbing (expected for any new dataset)

- No `feast` extra added to `kedro-datasets/pyproject.toml` (see how `redis`/`snowflake` register their extras and test-deps there for reference).
- No tests under `kedro-datasets/tests/feast/`.
- No `RELEASE.md` entry, no docs page/index entry, no `jsonschema/kedro-catalog-*.json` update — all called out in the PR's own checklist as unchecked.

## Suggestion

Given this is currently tied to a specific project/provider setup (hardcoded `project`/`provider` in `RepoConfig`) rather than being fully generic, it might be a better fit to contribute it to `kedro_datasets_experimental` first. That gives it a lower bar to land (no TSC approval needed, more relaxed docs/testing expectations) while we discuss the framework fit, and it can graduate to `kedro_datasets` once the API is more generic and the plumbing above is in place.

Happy to help iron out the details once you've had a chance to address the above!

I'll also post a few comments regarding the implementation separately.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+        credentials: dict[str, Any] | str | None = None,
+        load_args: dict[str, Any] | None = None,
+        save_args: dict[str, Any] | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
+
+        repo_config = RepoConfig(
+            registry=RegistryConfig(path=registry_path),
+            project="everycure",
+            provider="gcp",
```
ElenaKhaustova: `project="everycure"` and `provider="gcp"` are hardcoded here, which ties this dataset to a specific org/project rather than being a generic, reusable Kedro dataset. Could these be exposed as constructor parameters (with sensible defaults) so other users can actually use this outside of the `everycure` project?

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+    load and save.  Load returns a ``FeastFeatureView`` for point-in-time queries;
+    save writes directly to the backing BigQuery table.
+
+    For query-backed feature views (``BigQuerySource(query=...)``): load-only.
+    Attempting to save raises a ``DatasetError``.
+
+    Example catalog.yml entry::
+
+        feast_features:
+          type: kedro_utils.datasets.FeastDataset
```
ElenaKhaustova: `type: kedro_utils.datasets.FeastDataset` looks like a leftover from a different (private) repo layout — this module doesn't exist here. Should probably be `type: feast.FeastDataset` to match where this actually lives in `kedro_datasets`.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+        keep = entity_col_names | {TIMESTAMP_FIELD_NAME}
+        entity_df = entity_df[[c for c in entity_df.columns if c in keep]].copy()
+
+        if TIMESTAMP_FIELD_NAME not in entity_df.columns:
+            ts = (
+                timestamp
+                if timestamp is not None
+                else pd.Timestamp.now(tz=timezone.utc)
+            )
+            if ts.tzinfo is None:
```
ElenaKhaustova: `FeastFeatureView.get_historical_features` localizes a tz-naive `timestamp` to UTC here, but `FeastFeatureService.get_historical_features` (below, ~L135-141) does the same fallback without the `tz_localize` step. Worth unifying so a naive `timestamp` behaves the same regardless of whether you're loading a feature view or a feature service.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+                f"Feature view '{feature_view_name}' uses a query-based source "
+                f"and is read-only. Add a 'table' to its BigQuerySource to enable writes."
+            )
+
+        project, bq_dataset, table_name = _parse_bq_table(batch_source.table)
+        bq_table = GBQTableDataset(
+            dataset=bq_dataset,
+            table_name=table_name,
+            project=project,
+            credentials=self._credentials,
```
ElenaKhaustova: `credentials` is only ever used on the `save()` path (forwarded into `GBQTableDataset`). On `load()`, Feast's own `FeatureStore`/`BigQuerySource` resolves BigQuery auth independently (typically via ADC), so a `credentials` value configured by the user is silently ignored for reads. Might be worth documenting this asymmetry, or wiring credentials through for reads too if Feast supports it.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+    @staticmethod
+    def _update_bq_table_schema(
+        bq_table: GBQTableDataset,
+        bq_schema: list[bigquery.SchemaField],
+        timestamp_field: str,
+    ) -> None:
+        """Create the table if absent, or add any columns present in Feast but missing in BQ."""
+        desired = {f.name: f for f in bq_schema}
+
+        if not bq_table._exists():
```
ElenaKhaustova: This reaches into `GBQTableDataset` private attributes (`_exists()`, `_connection`, `_project_id`, `_dataset`, `_table_name`) which aren't part of its public API — any refactor of `GBQTableDataset` could silently break this. Also, more substantively: `save()` will now implicitly create the BigQuery table or `ALTER TABLE ... ADD COLUMN` when the Feast schema drifts from what's in BQ. That's a significant side effect (DDL) to run implicitly on every write with no opt-out — might be worth gating behind an explicit flag (e.g. `save_args: {allow_schema_evolution: true}`) rather than always-on.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,358 @@
...
+    # Complex types like Array use the class name as fallback.
+    key = dtype.name if hasattr(dtype, "name") else type(dtype).__name__
+    return _FEAST_FIELD_TYPE_TO_BQ.get(key, "STRING")
+
+
+def _parse_bq_table(table_ref: str) -> tuple[str, str, str]:
+    parts = table_ref.split(".")
+    if len(parts) != 3:
+        raise DatasetError(
+            f"Expected BigQuery table reference 'project.dataset.table', got: {table_ref!r}"
```
ElenaKhaustova: This is what's currently failing the `lint` CI check (ruff flags the f-string used directly inside `raise DatasetError(...)`). Same pattern shows up again around L248-249, L283-284, and L299-300.

ElenaKhaustova (commented):
A few implementation-level comments below (in addition to the top-level checklist items).

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,285 @@
...
+    Example catalog.yml entry::
+
+    .. code-block:: yaml
+
+        feast_features:
+          type: kedro_datasets_experimental.feast.FeastDataset
+          repo:  # forwarded to feast.RepoConfig
+            registry: gs://bucket/feast
+            project: project_name
+            provider: gcp # default is gcp
```
ElenaKhaustova: This comment says the default is `gcp`, but per the `__init__` docstring (and the actual code below) the default `provider` is `"local"`. Looks like a leftover from before the refactor — worth fixing so the example doesn't mislead readers.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,285 @@
...
+        # Optionally also push to the online store (requires an online_store to
+        # be configured in `repo`; it is disabled by default).
+        if write_mode == "online_and_offline":
+            self._feature_store.write_to_online_store(feature_view_name, data)
+
+    def _create_bigquery_table(self, feature_view_name: str) -> None:
+        """Create the feature view's backing BigQuery table if it doesn't exist.
+
+        The table schema is derived from the feature view (entity keys, features,
+        then the event timestamp) and day-partitioned on the timestamp. Raises a
```
ElenaKhaustova: This docstring claims the created table is day-partitioned on the timestamp field, but I don't see `time_partitioning` being set anywhere in `_create_bigquery_table` anymore. Either the partitioning logic got dropped during the simplification, or the docstring needs updating to match current behavior.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,285 @@
...
+        if not batch_source.table:
+            msg = (
+                f"Feature view '{feature_view_name}' uses a query-based source "
+                f"with no table to create."
+            )
+            raise DatasetError(msg)
+
+        timestamp_field = batch_source.timestamp_field
+        schema = self._build_bq_schema(feature_view, timestamp_field)
+        table = bigquery.Table(batch_source.table, schema=schema)
```
ElenaKhaustova: `bigquery.Table(table_ref)` requires a fully-qualified `project.dataset.table` string when given a string (no `default_project` fallback). The `table_project` fallback computed just below (for the 2-part `dataset.table` case) is only used for the `bigquery.Client(...)` call, not to qualify this `Table` reference — so a 2-part `batch_source.table` will raise `ValueError` here before `create_table()` is ever called. Suggest building the fully-qualified ID first, e.g. `table_id = batch_source.table if len(parts) == 3 else f"{table_project}.{batch_source.table}"`, and passing that in instead.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # Run against the project the source table lives in (a fully-qualified
+        # `project.dataset.table` ref), falling back to the offline store's.
+        offline = self._feature_store.config.offline_store
+        parts = batch_source.table.split(".")
+        table_project = (
+            parts[0]
+            if len(parts) == 3
```
ElenaKhaustova: This is what's failing the `lint` CI check (ruff `PLR2004`, magic value in comparison). A named constant (e.g. `_FULLY_QUALIFIED_PARTS = 3`) or `# noqa: PLR2004` would fix it.

kedro-datasets/pyproject.toml
```diff
@@ -386,6 +386,9 @@ experimental = [
     "netcdf4>=1.6.4",
     "xarray>=2023.1.0",
     "rioxarray",
+    "feast[gcp]==0.64.0",
```
ElenaKhaustova: Also flagging that `feast[gcp]==0.64.0` is what's breaking `unit-tests-experimental` in CI — it transitively pulls `llvmlite==0.36.0` (via `numba`), which has no wheels and fails to build on Python ≥3.11. Might be worth bumping to a newer Feast version or adding a constraint that avoids the incompatible `numba`/`llvmlite` pin.

Separately: these deps only went into the catch-all `experimental`/`experimental_test` bundles — could we also add a dedicated `feast = [...]` extra (mirroring `weaviate`, `mlrun`, etc.) so users can `pip install kedro-datasets[feast]` without pulling in every other experimental dataset's dependencies?

ElenaKhaustova (commented):
A few implementation-level comments for this round, in addition to the top-level summary.

ElenaKhaustova:
Thanks @lvijnck, this is a big improvement — moving to `kedro_datasets_experimental`, making `repo` a generic passthrough to `feast.RepoConfig`, dropping the private `GBQTableDataset` reach-through in favor of Feast's own `write_to_offline_store`/`write_to_online_store`, making table creation opt-in via `create_table`, and adding real tests all address the previous round's feedback well.

A couple of things still need attention:

## CI

- **`lint` fails** — ruff (`PLR2004`, magic value) flags `feast_dataset.py:254` (`if len(parts) == 3`). Needs a named constant (e.g. `_FULLY_QUALIFIED_PARTS = 3`) or a `# noqa: PLR2004`.
- **All `unit-tests-experimental` matrix jobs fail** — `feast[gcp]==0.64.0` pulls in `llvmlite==0.36.0` transitively (via `numba`), which has no wheel and fails to build from source on Python ≥3.11:
  ```
  × Failed to build `llvmlite==0.36.0`
  RuntimeError: Cannot install on Python version 3.12.3; only versions ...
  ```
  This blocks every experimental-test matrix leg (and cascades into `merge-gatekeeper` failing too). Needs a `feast` version bump or an extra/constraint that avoids the incompatible `numba`/`llvmlite` pin for the Python versions this repo supports (3.10–3.14).

## Bug: `_create_bigquery_table` can crash on non-fully-qualified table refs

`table = bigquery.Table(batch_source.table, schema=schema)` (L246) is constructed directly from `batch_source.table`. `google-cloud-bigquery`'s `Table(table_ref)` requires a fully-qualified `project.dataset.table` string when given a string — it has no `default_project` fallback. A few lines below (L251-256), the code computes `table_project` specifically to handle the case where `batch_source.table` is just `"dataset.table"` (2 parts) — but that fallback is only used for the `bigquery.Client(project=table_project, ...)` call, never to qualify the `Table` reference itself. So if `batch_source.table` is ever a 2-part string, `bigquery.Table(...)` raises `ValueError` before `client.create_table()` is even reached.

Suggested fix: build the fully-qualified ID first, e.g.
```python
table_id = batch_source.table if len(parts) == 3 else f"{table_project}.{batch_source.table}"
table = bigquery.Table(table_id, schema=schema)
```

## Doc/code mismatches

- L224: the docstring says the created table is "day-partitioned on the timestamp", but the code no longer sets `table.time_partitioning` anywhere (this was present in an earlier revision but appears to have been dropped during the simplification). Either restore the partitioning or update the docstring.
- L89: the YAML example comment `provider: gcp # default is gcp` contradicts both the actual default (`"provider": "local"` in `__init__`) and the `__init__` docstring itself, which says `provider` defaults to `"local"`. Looks like a leftover from before the refactor.

## Packaging

- The new deps (`feast[gcp]==0.64.0`, `google-cloud-bigquery>=3.0`, `pandas-gbq>=0.12.0`) were only added to the catch-all `experimental`/`experimental_test` bundles (pyproject.toml L389, L418), not as a standalone `feast = [...]` extra the way `weaviate`, `mlrun`, etc. are. Right now there's no way to `pip install kedro-datasets[feast]` without pulling in every other experimental dataset's dependencies (torch, opencv, chromadb, ...). Worth adding a dedicated extra.
- No `RELEASE.md` entry yet — other experimental dataset additions (weaviate, opik, mlrun) each got a row in the "Added the following new **experimental** datasets" table; Feast doesn't have one.

Happy to take another pass once these are addressed!

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: fixed

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: fixed, removed for now to keep it simple

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: done

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: fixing

kedro-datasets/pyproject.toml
lvijnck: done

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,286 @@
...
+            offline_store:
+              type: bigquery # default is bigquery
+              location: EU
+            online_store:  # optional; required for online_and_offline writes
+              type: sqlite
+              path: /tmp/online_store.db
+          save_args:
+            feature_view_name: features
+            write_mode: online_and_offline  # or "offline" (default)
+            create_table: true  # bootstrap the BigQuery table from the FV schema
```
lvijnck: This creation is really here to avoid the chicken and egg problem. Feast requires the table to be there. Alternatively we could write to a regular GBKDataset, but it would enforce us to specify the table location in two places. 

Not super happy with it, but think it solves quite a frustrating issue when trying to set it up the first time.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,321 @@
...
+        """Retrieve historical features from the offline store.
+
+        Feast supports two mutually exclusive retrieval modes; pass the
+        arguments for exactly one of them:
+
+        * **Entity-driven** — pass ``entity_df``: a point-in-time join. The
+          DataFrame holds the entity join key column(s) and an event-timestamp
+          column that defines the point-in-time context for each row (Feast
+          requires the timestamp; it is not fabricated).
+        * **Timestamp range** — pass ``start_date``/``end_date`` and *no*
```
ElenaKhaustova, line 64: This "Timestamp range" mode will actually crash for the dataset's default/primary offline store (BigQuery), not just be "unsupported." I checked against Feast **v0.64.0** (the version pinned in this PR):

- `FeatureStore.get_historical_features` forwards `start_date`/`end_date` as `**kwargs` down through `PassthroughProvider.get_historical_features` into `self.offline_store.get_historical_features(..., **kwargs)`.
- `BigQueryOfflineStore.get_historical_features`'s concrete signature has **no `**kwargs`** and no `start_date`/`end_date` params (`entity_df` is also non-optional there) — see `feast/infra/offline_stores/bigquery.py`.
- So calling this in entity-less mode against BigQuery raises a hard `TypeError: get_historical_features() got an unexpected keyword argument 'start_date'`, not a graceful `NotImplementedError`/`ValueError`.

By contrast, `DaskOfflineStore.get_historical_features` (which backs the `{"type": "file"}` store used in `test_load_historical_features_timestamp_range`) does accept `**kwargs` and explicitly handles `entity_df is None` — which is why the new test passes without catching this.

Worth calling out the BigQuery limitation in the docstring (or scoping/guarding this mode to the offline stores that actually support it — Postgres/Dask/Spark/Ray per Feast's docs).

ElenaKhaustova (commented):
One follow-up finding on the new entity-less retrieval mode added in this round.

kedro-datasets/pyproject.toml
```diff
@@ -386,6 +387,13 @@ experimental = [
...
```
ElenaKhaustova: worth relaxing this to feast[gcp]>=0.64.0 (or >=0.64.0,<0.65 if there's a real compatibility reason to cap it) unless there's a specific reason for the hard pin

ElenaKhaustova:
**Docs are missing** — every other recently-added experimental dataset (e.g. `opik.EvaluationDataset` in #1364) wires up three things that this PR doesn't touch:

1. A row in `kedro-datasets/docs/api/kedro_datasets_experimental/index.md`, e.g.:
   ```
   [feast.FeastDataset](feast.FeastDataset.md) | ``FeastDataset`` handles storing and retrieving features from Feast.
   ```
   (alphabetically between `databricks.ExternalTableDataset` and `langchain.PromptDataset`)

2. A new stub page `kedro-datasets/docs/api/kedro_datasets_experimental/feast.FeastDataset.md`:
   ```
   ::: kedro_datasets_experimental.feast.FeastDataset
       options:
           members: true
           show_source: true
   ```

3. Two entries in `kedro-datasets/mkdocs.yml` — one in the `plugins:` section (used for the generated `llms.txt`), and one under `nav:` in a new "Feast" group (between "Databricks" and "Langchain"), mirroring how `opik.OpikEvaluationDataset.md` was added in both places in #1364.

Without these, the class docstring won't get rendered on the docs site or picked up by the doctest/nav validation the way other experimental datasets are.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: Yeah it actually _should_ be supported

https://github.com/feast-dev/feast/pull/6569

PR was merged recently, the interface would remain unchanged.

kedro-datasets/pyproject.toml
lvijnck: done

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: This only works with the latest pre-release though, i.e., `0.65.0.dev3`

Seems not that great to pin onto that version?

lvijnck: Update: gave it a test in the test repo

https://github.com/lvijnck/kedro-plugins/pull/1/changes

Works correctly with BQ :)

ElenaKhaustova: Checked PyPI just now, `feast 0.64.0` is still the latest published release, `0.65.0` hasn't shipped yet. So as merged, `feast[gcp]>=0.64.0` will resolve to `0.64.0` for a normal install, and this method will still raise a raw `TypeError` from inside Feast (not a `DatasetError`) if someone calls `get_historical_features(start_date=..., end_date=...) `against the default `BigQuery` offline store. The docstring above doesn't mention this limitation.

The relaxed pin is nice here since it'll self-heal automatically once `0.65` ships with #6569 (no code change needed, per your note that the interface is unchanged). But until then, could we add a one-line caveat to the docstring (e.g. `"timestamp-range retrieval currently requires Feast ≥0.65 for the BigQuery offline store; already supported on Postgres/Dask/Spark/Ray"`) so users hitting this don't get a confusing crash with no explanation?

kedro-datasets/mkdocs.yml
```diff
@@ -344,6 +345,8 @@ nav:
           - yaml.YAMLDataset: api/kedro_datasets/yaml.YAMLDataset.md
       - Experimental Datasets:
         - Overview: api/kedro_datasets_experimental/index.md
+        - Feast:
```
ElenaKhaustova: The new `Feast` nav group is inserted right after `Overview`, before `ChromaDB` — breaking the alphabetical order the rest of this list follows (it should sit after `Databricks`, before `Langchain`, matching where it correctly landed in `docs/api/kedro_datasets_experimental/index.md`).

kedro-datasets/kedro_datasets_experimental/feast/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+
+from typing import Any
+
+import lazy_loader as lazy
+
+# https://github.com/pylint-dev/pylint/issues/4300#issuecomment-1043601901
+FeastDataset: Any
```
ElenaKhaustova: This module's docs page doesn't actually render — I checked the [RTD preview build](https://kedro--1454.org.readthedocs.build/projects/kedro-datasets/en/1454/api/kedro_datasets_experimental/feast.FeastDataset/) for this PR and it's empty (no docstring, no `__init__` signature, no methods — just `FeastDataset` `module-attribute`).

`mkdocstrings` statically analyzes this file and never triggers `__getattr__`, so it only sees the bare `FeastDataset: Any` annotation here. Every other experimental dataset (`chromadb`, `langfuse`, `opik`, `pytorch`, `netcdf`, `databricks`, ...) wraps the lazy-loader setup in an eager import so static analysis can resolve it, e.g.:

```python
try:
    from .feast_dataset import FeastDataset
except (ImportError, RuntimeError):
    # For documentation builds that might fail due to dependency issues
    FeastDataset: Any
```

Adding that `try/except` import here (instead of the bare annotation) should fix the rendering.

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,325 @@
...
+    offline store. Save persists a DataFrame to a feature view's batch source
+    via Feast's ``write_to_offline_store`` (which matches the DataFrame to the
+    feature view's schema and appends) and, when ``write_mode`` is
+    ``online_and_offline``, also to the online store. Feast infers the write
+    schema from the *existing* table; set ``create_table: true`` to bootstrap a
+    (BigQuery) table from the feature view schema when it does not yet exist.
+
+    Example catalog.yml entry::
+
+    .. code-block:: yaml
```
ElenaKhaustova: This uses reStructuredText syntax (`Example catalog.yml entry::` + `.. code-block:: yaml`) rather than the Google-style/Markdown convention used everywhere else in this codebase (an `Examples:` section header + a plain triple-backtick fenced block, e.g. `CSVDataset`, `ChromaDBDataset`).

Since `mkdocs.yml` sets `docstring_style: google`, griffe doesn't recognize `Example catalog.yml entry::` as a section header, so it passes the raw text through. Checked the [RTD preview build](https://kedro--1454.org.readthedocs.build/projects/kedro-datasets/en/1454/api/kedro_datasets_experimental/feast.FeastDataset/) for this PR — the literal text `.. code-block:: yaml` leaks onto the rendered page instead of being interpreted as a code-fence marker.

Suggest switching to:
```python
Examples:
    Using the YAML API:

    ```yaml
    feast_features:
      ...
    ```
```

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,325 @@
...
+
+        # Run against the project the source table lives in (a fully-qualified
+        # `project.dataset.table` ref), falling back to the offline store's.
+        offline = self._feature_store.config.offline_store
+        parts = batch_source.table.split(".")
+        is_fully_qualified = len(parts) == _FULLY_QUALIFIED_BQ_TABLE_PARTS
+        table_project = (
+            parts[0]
+            if is_fully_qualified
+            else (offline.billing_project_id or offline.project_id)
```
ElenaKhaustova, line 292: When `batch_source.table` is a 2-part `dataset.table` reference (not fully-qualified) and `offline_store` has neither `project_id` nor `billing_project_id` set (both default to `None`, and the docstring's own catalog example above sets neither), `table_project` resolves to `None` here — and the `table_id` built below silently becomes the string `"None.my_dataset.my_table"`.

Verified with a real `bigquery.Table(...)` call: it constructs without error (`project` ends up as the literal string `"None"`, not Python `None` — no validation happens client-side), so this only fails later at `client.create_table(...)` with a confusing `NotFound`/`Forbidden` from the BigQuery API against a nonexistent `"None"` project, rather than a clear error pointing at the real missing config. Also not covered by either `create_table` test (both use a fully-qualified 3-part table or a non-BigQuery source).

Worth raising a clear `DatasetError` when `table_project` is `None`, e.g.:
```python
if table_project is None:
    msg = (
        f"Cannot resolve a project for table '{batch_source.table}'; set "
        f"'billing_project_id' or 'project_id' on the offline_store, or use "
        f"a fully-qualified 'project.dataset.table' reference."
    )
    raise DatasetError(msg)
```

ElenaKhaustova (approved):
Thank you, @lvijnck! I made another pass, and it looks good to me 👍

Could you please also fix DCO and add your name under the `Community contributions` section of the `RELEASE.md`?

kedro-datasets/mkdocs.yml
```diff
@@ -174,6 +174,7 @@ plugins:
 
         Experimental LLM and AI:
           - api/kedro_datasets_experimental/chromadb.ChromaDBDataset.md: ChromaDB vector database integration
+          - api/kedro_datasets_experimental/feast.FeastDataset.md: Feast feature store integration
```
ankatiyar, line 177: Is "Experimental LLM and AI" the right heading for this to go under? There's "Experimental Deep Learning" and "Experimental Specialised Formats" below as well - maybe can go under one of those? Just checking

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+
+    Using the YAML API:
+
+    ```yaml
+    feast_features:
+        type: kedro_datasets_experimental.feast.FeastDataset
+        repo:  # forwarded to feast.RepoConfig
+        registry: gs://bucket/feast
+        project: project_name
+        provider: gcp # default is local
```
ankatiyar: Do these keys need to be indented to be under `repo`:

```
repo:
  registry: gs://bucket/feast
  ...
```

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
```diff
@@ -0,0 +1,336 @@
...
+    _WRITE_MODES = frozenset({"offline", "online_and_offline"})
+
+    def __init__(
+        self,
+        *,
+        repo: dict[str, Any],
+        credentials: dict[str, Any] | str | None = None,
+        load_args: dict[str, Any] | None = None,
+        save_args: dict[str, Any] | None = None,
+        metadata: dict[str, Any] | None = None,
```
ankatiyar, line 151: I would add a `self.metadata = metadata` in the constructor to maintain the Kedro convention

kedro-datasets/pyproject.toml
```diff
@@ -386,6 +387,13 @@ experimental = [
     "netcdf4>=1.6.4",
     "xarray>=2023.1.0",
     "rioxarray",
+    "feast[gcp]>=0.64.0",
+    "google-cloud-bigquery>=3.0",
+    "pandas-gbq>=0.12.0",
+    # `u8darts[all]` pulls numba via pyod/shap. Without a floor the resolver
+    # backtracks to the ancient numba 0.53 (loose numpy metadata), which pins
+    # llvmlite 0.36 -- a version with no wheels that fails to build on py>=3.11.
+    "numba>=0.61",
```
ankatiyar: This issue was resolved already in https://github.com/kedro-org/kedro-plugins/pull/1461, is this needed?

ankatiyar (commented):
Thanks @lvijnck the code looks good to me, I have some minor comments/questions :D

kedro-datasets/mkdocs.yml
lvijnck: I don't have a strong opinion about this one, happy to move as needed

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: Correct, fixing

kedro-datasets/pyproject.toml
lvijnck: Done

kedro-datasets/kedro_datasets_experimental/feast/feast_dataset.py
lvijnck: Done

kedro-datasets/pyproject.toml
lvijnck: Let me try to remove

lvijnck: Works, good spot! :)

Think I made these changes before sync with main

kedro-datasets/mkdocs.yml
ravi-kumar-pilla: I think "Experimental LLM and AI" is a bit of a catch-all for experimental datasets that don't fit Deep Learning / Hyperparameter / Specialized Formats. ChromaDB and Weaviate are vector stores, not strictly LLM tools either, so Feast sitting alongside them seems fine to me. 

May be we need to create a separate ticket to re-organize the experimental sections as we are adding new experimental features/datasets

```
Experimental GenAI and Observability:
  - langchain.PromptDataset
  - langfuse.* / opik.*          # prompts, traces, evaluation

Experimental Vector Stores:
  - chromadb.ChromaDBDataset
  - weaviate.WeaviateVectorStoreDataset

Experimental Feature Stores and MLOps:
  - feast.FeastDataset
  - mlrun.*                      # currently missing from llms.txt

Experimental Deep Learning:
  - pytorch / safetensors        # unchanged

Experimental Hyperparameter Tuning:
  - optuna                       # unchanged

Experimental Data Connectors and Platforms:
  - databricks.ExternalTableDataset
  - polars.PolarsDatabaseDataset

Experimental File and Scientific Formats:
  - pypdf, video, netcdf, rioxarray

Experimental Time Series:
  - prophet.ProphetModelDataset  # or fold into "File and Scientific" if you want fewer groups
```

ravi-kumar-pilla (approved):
Thanks for the contribution @lvijnck . Looks good to me

SajidAlamQB (approved):
Thank you @lvijnck looks great!

ankatiyar (approved):
Thank you @lvijnck :D 💯

PR #1464, opened 2026-07-17
feat(datasets): Add `FAISSVectorStoreDataset`

lrcouto:
## Description

Adds `faiss.FAISSVectorStoreDataset`, a new experimental vector-store dataset on the `VectorStoreHandle` abstraction. (https://github.com/orgs/kedro-org/projects/11/views/3?sliceBy%5Bvalue%5D=lrcouto&pane=issue&itemId=173203982&issue=kedro-org%7Ckedro-plugins%7C1353)

FAISS is a different beast than Weaviate or ChromaDB. It's a numerical index library with no server, no metadata storage, no string IDs, and no persistence beyond raw index files. So this dataset is designed with the premise that, if the user is choosing to use the FAISS backend instead of one of those other available options, it's because they either do not want to, or cannot use these abstractions. e.g. If they wanted persistence, filtering, or index management handled for them, they'd pick one of the other backends instead.

So this dataset builds the ID-mapping, metadata side-store, and persistence machinery itself inside the handle, rather than delegating to a client library. Some specific details:

- Persistence is explicit, not automatic. `add()`/`delete()`/`close()` never touch disk. `handle.save()` is the only persistence trigger, framed as a thin wrapper around what you'd do by hand with `raw_client` + `faiss.write_index()`. No remembered path (`save(path=...)` doesn't change what a later bare `save()` defaults to), no overwrite guard, local filesystem only (no `fsspec`). The one addition beyond a literal passthrough is a temp-file + `os.replace()` swap so a crash mid-write can't corrupt an index that was working previously. 
- `save()` is as a first-class method rather than "just use `raw_client` yourself", because a raw `write_index()` call can't touch the ID-mapping/properties sidecar at all (that's our own invention, not something FAISS knows about), so every caller would otherwise hand-roll the identical two-file write.
- Filters are a Python callable predicate over `properties`. FAISS has no native query language to delegate to, so a raw callable is more aligned to the actual mechanics of it. Does a scan over the in-memory properties store, then pushes into FAISS's own search via `IDSelectorBatch`/`SearchParameters`.
- `index_factory` is exposed directly. `"Flat"` default, or approximate types like `"IVF100,Flat"`/`"HNSW32"` rather than restricting to exact search. Locking it to flat-only would go against one of the main reasons to pick FAISS over Chroma or Weaviate in the first place, which is searching large collections fast. `IVF*` types need `handle.train()` before `add()` since they cluster the vector space and need representative data to place the cluster boundaries; `HNSW*` types build incrementally and need no training. `add()` refuses to silently auto-train off an arbitrary first batch, and training quality depends on that batch being representative, which isn't a call this handle can make for the caller.

So in short, this dataset is gonna require the user to be a bit more "hands on", by design.

## Checklist


kedro-datasets/kedro_datasets_experimental/faiss/faiss_vector_store_dataset.py
```diff
@@ -0,0 +1,544 @@
...
+    def _load(self) -> FAISSVectorStoreHandle:
+        index_file = Path(self._index_path) if self._index_path else None
+        meta_file = Path(f"{self._index_path}.meta.json") if self._index_path else None
+
+        if (
+            index_file is not None
+            and meta_file is not None
+            and index_file.exists()
+            and meta_file.exists()
+        ):
```
ElenaKhaustova: Bug: if only one of the two persistence files exists (e.g. crash between the index write and sidecar write in `save()`, or manual deletion), this condition is `False` and we silently fall through to the `else` branch — creating a fresh empty index with no error. That's silent data loss.

Suggestion: check for the one-file-missing case and raise explicitly:

```python
if index_file is not None and meta_file is not None:
    index_exists = index_file.exists()
    meta_exists = meta_file.exists()
    if index_exists != meta_exists:
        missing = meta_file if index_exists else index_file
        raise DatasetError(
            f"Found one persistence file but not the other — "
            f"'{missing}' is missing. Both '{index_file}' and "
            f"'{meta_file}' are required to hydrate."
        )
```

ElenaKhaustova (commented):
Nice work — well-designed, well-documented, and thorough test coverage. A few things below, the first one being the most important.

kedro-datasets/kedro_datasets_experimental/faiss/faiss_vector_store_dataset.py
```diff
@@ -0,0 +1,544 @@
...
+        except Exception as e:
+            raise DatasetError(f"save() failed while writing the FAISS index: {e}") from e
+
+        meta = {
+            "next_id": self._next_id,
+            "id_to_internal": self._id_to_internal,
+            "properties": self._properties,
+        }
+        try:
+            Path(f"{target}.meta.json").write_text(json.dumps(meta))
```
ElenaKhaustova: The index file gets the temp-file + `os.replace()` treatment (nice), but the sidecar write here is a bare `write_text()` — not atomic. A crash mid-write corrupts the sidecar, which then triggers the silent-data-loss issue above.

Worth applying the same temp-file pattern here:

```python
meta_tmp = f"{target}.meta.json.tmp"
Path(meta_tmp).write_text(json.dumps(meta))
os.replace(meta_tmp, f"{target}.meta.json")
```

kedro-datasets/kedro_datasets_experimental/faiss/faiss_vector_store_dataset.py
```diff
@@ -0,0 +1,544 @@
...
+                "save() requires a 'path' argument, since no 'index_path' "
+                "was configured on the dataset."
+            )
+
+        tmp_path = f"{target}.tmp"
+        try:
+            faiss.write_index(self._index, tmp_path)
+            os.replace(tmp_path, target)
+        except Exception as e:
+            raise DatasetError(f"save() failed while writing the FAISS index: {e}") from e
```
ElenaKhaustova, line 383: Nit: if `faiss.write_index` partially writes and then fails, the `.tmp` file is left on disk. Consider cleaning it up:

```python
except Exception as e:
    with contextlib.suppress(OSError):
        os.unlink(tmp_path)
    raise DatasetError(...) from e
```

kedro-datasets/kedro_datasets_experimental/tests/faiss/test_faiss_vector_store_dataset.py
```diff
@@ -0,0 +1,463 @@
...
+        store2 = ds2.load()
+        assert store2.describe()["count"] == 3
+        hits = store2.search(vector=[1.0, 0.0, 0.0, 0.0], top_k=1)
+        assert hits[0]["id"] == "doc1"
+        assert hits[0]["properties"] == {"topic": "ml"}
+
+    def test_load_dimension_mismatch_raises(self, tmp_path, sample_records):
+        index_path = str(tmp_path / "my.index")
+        ds1 = FAISSVectorStoreDataset(dimension=4, index_path=index_path)
+        ds1.load().add(sample_records)
```
ElenaKhaustova: Dead code: `ds1.load()` creates handle 1, `.add(sample_records)` inserts into its in-memory index, but this handle is immediately discarded without saving. The next `ds1.load()` on line 135 creates a *new* fresh handle (no files on disk yet) and saves an empty index. The `sample_records` fixture does nothing here.

Can be simplified to:

```python
def test_load_dimension_mismatch_raises(self, tmp_path):
    index_path = str(tmp_path / "my.index")
    ds1 = FAISSVectorStoreDataset(dimension=4, index_path=index_path)
    ds1.load().save()
    ...
```

ElenaKhaustova (commented):
While reviewing, I also noticed that the `VectorStoreHandle.add()` docstring in `vectorstore_base/vector_store_handle.py` marks "properties" as (required):

```python
records: Each record is a dict with keys:
    - ``"properties"`` (required): payload dict stored alongside
      the vector.
```

But none of the three implementations actually enforce this — all of them silently default to {}:

- Weaviate: record.get("properties", {})
- ChromaDB: record.get("properties") or {}
- FAISS (this PR): record.get("properties", {})

The implementations have it right — requiring "properties" would force users to write `"properties": {}` on every vector-only record for no benefit. The docstring just needs to say (optional) instead.

Could you please open a quick follow-up PR to fix the docstring?

ravi-kumar-pilla:
Thanks @lrcouto for clear PR description

+1 to @ElenaKhaustova's persistence comments, the partial-file hydration case (only index or only `.meta.json` present → silent fresh index) feels like the main blocker for me. Same for making the sidecar write atomic.

It would be nice to have :
- Fixing the dead code in `test_load_dimension_mismatch_raises` as Elena suggested
- Adding a test for the one-file-missing case once fixed

Thank you

lrcouto:
Thank you for the reviews!

I opened a PR for the update on the VectorStoreHandle docstring here as well: https://github.com/kedro-org/kedro-plugins/pull/1465

ElenaKhaustova (approved):
LGTM 👍

SajidAlamQB (approved):
Thank you @lrcouto looks great!

PR #1472, opened 2026-07-31
feat(datasets): add generated catalog schema workflow

### LINKED ISSUE #964: Automatically create dataset JSON schemas through introspection
## Description
When a new dataset is added to `kedro-datasets`, the JSON schema that [enables Kedro catalog validation in an IDE](https://docs.kedro.org/en/stable/development/set_up_vscode.html#configuring-the-kedro-catalog-validation-schema) must be manually updated.

These schemas could be created programmatically through introspection (as [suggested by](https://github.com/kedro-org/kedro/issues/4258#issuecomment-2439443462) @datajoely).

## Context
Manually creating a JSON schemas is an additional maintenance step (that may be easy to overlook).

## Possible Implementation

1. The [inspect](https://docs.python.org/3/library/inspect.html) module could be used to get the dataset's attributes.
2. A pre-commit hook could generate the required `kedro-catalog-X.XX.json` (only for the current package version).
  a. When a developer commits a new dataset, the JSON schema is created for them.
3. Remove references to updating `jsonschema` from `CONTRIBUTING.md` and `PULL_REQUEST_TEMPLATE.md`.

## Possible Alternatives
The introspection code could be triggered manually instead of by a pre-commit hook. The hook would fully automate it, though.

JokeGbenro:
## Description

Closes #964.

This PR adds an automated workflow for generating the Kedro catalog JSON schema, replacing the previous manual maintenance process for `kedro-catalog-1.0.0.json`.

## Development notes

- Added `kedro_datasets._generate_catalog_schema`, which generates the latest catalog schema from dataset source files.
- Added `kedro_datasets._schema_overrides` for schema facts that cannot be inferred from constructor signatures or docstrings.
- Regenerated `static/jsonschema/kedro-catalog-1.0.0.json` from the generator.
- Added a manual pre-commit hook to regenerate the catalog schema.
- Removed the PR checklist item asking contributors to manually update catalog JSON schema files.
- Added tests covering the generator, override merge behavior, source-level dataset discovery, annotation mapping, CLI behavior, and schema write/check behavior.

Tested with:

```bash
python -m ruff check kedro_datasets\_generate_catalog_schema.py kedro_datasets\_schema_overrides.py tests\test_generate_catalog_schema.py
python -m pytest tests\test_generate_catalog_schema.py -q --no-cov
python -m pytest tests\test_generate_catalog_schema.py -q --cov-reset --cov=kedro_datasets._generate_catalog_schema --cov=kedro_datasets._schema_overrides --cov-report=term-missing --no-cov-on-fail
python -m kedro_datasets._generate_catalog_schema --check
python -m json.tool static\jsonschema\kedro-catalog-1.0.0.json
```

 All commits were signed off by including a `Signed-off-by` line in the commit message. [[See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/)](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

## Checklist

### DIFF
--- a/.github/workflows/kedro-datasets.yml
+++ b/.github/workflows/kedro-datasets.yml
@@ -79,6 +79,27 @@ jobs:
           args: --max-concurrency 32 --exclude "@.lycheeignore" site/
           workingDirectory: kedro-datasets
 
+  check-schema:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "latest"
+          python-version: "3.10"
+          activate-environment: true
+      - name: Install dependencies
+        run: |
+          uv pip install "kedro @ git+https://github.com/kedro-org/kedro@main"
+          uv pip install -e kedro-datasets
+      - name: Check generated catalog schema
+        working-directory: kedro-datasets
+        run: python -m kedro_datasets._generate_catalog_schema --check
+
   detect-secrets:
     uses: ./.github/workflows/detect-secrets.yml
     with:
--- a/kedro-datasets/RELEASE.md
+++ b/kedro-datasets/RELEASE.md
@@ -1,14 +1,7 @@
 # Upcoming release
 
 ## Major features and improvements
-## Breaking changes
-## Breaking changes to experimental datasets
-## Bug fixes and other changes
-## Community contributions
-
-# Release 9.6.0
-
-## Major features and improvements
+- Added source-based generation and CI validation for the Kedro catalog JSON schema.
 * Added `vectorstore_base.AbstractVectorStoreDataset` and `vectorstore_base.VectorStoreHandle`, backend-agnostic abstract base classes for vector store datasets.
 - Added the following new **experimental** datasets:
 
--- a/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
+++ b/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
@@ -276,6 +276,8 @@ def _submodule_attrs_for(subpackage: str) -> dict[str, str]:
                 continue
             class_to_module: dict[str, str] = {}
             for key, value in zip(keyword.value.keys, keyword.value.values):
+                if key is None:
+                    continue
                 module_name = ast.literal_eval(key)
                 for class_name in ast.literal_eval(value):
                     class_to_module[class_name] = module_name
--- a/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
+++ b/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
@@ -48,6 +48,7 @@ _NAME_TYPE_MAP: dict[str, str] = {
     "tuple": "array",
     "Version": "object",
     "PathLike": "string",
+    "os.PathLike": "string",
     "Path": "string",
     "PurePath": "string",
     "PurePosixPath": "string",
@@ -271,8 +272,15 @@ def _dataset_spec_from_source(
                     parameters=_parameters_from_init(child),
                     docstring=ast.get_docstring(child),
                 )
+        type_id = f"{subpackage}.{class_name}"
+        if type_id not in SCHEMA_OVERRIDES:
+            raise ValueError(
+                f"{type_id} defines no __init__ method (likely inherited). "
+                "You must add a manual schema override for it in _schema_overrides.py "
+                "so that it does not silently get an empty catalog schema."
+            )
         return DatasetSpec(
-            type_id=f"{subpackage}.{class_name}",
+            type_id=type_id,
             parameters=[],
             docstring=None,
         )
--- a/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
+++ b/kedro-datasets/kedro_datasets/_generate_catalog_schema.py
@@ -1,0 +1,519 @@
+"""Generate the versioned Kedro catalog JSON schema by introspection.
+
+Run as a module::
+
+    python -m kedro_datasets._generate_catalog_schema          # write the schema
+    python -m kedro_datasets._generate_catalog_schema --check  # verify it is current
+
+The schema is derived from the source-level ``__init__`` signature and docstring
+of every public dataset class exported by the ``kedro_datasets`` subpackages.
+This module is the source of truth for ``static/jsonschema/<SCHEMA_FILENAME>``;
+hand edits to that file are overwritten. Genuinely non-introspectable facts
+belong in :mod:`kedro_datasets._schema_overrides`.
+
+Generation reads Python source files instead of importing dataset classes, so it
+does not require optional dataset dependencies such as PySpark or Dask.
+"""
+
+from __future__ import annotations
+
+import argparse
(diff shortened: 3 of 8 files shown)


lrcouto:
Thank you @JokeGbenro , that's a welcome addition.

I'm wondering if it would make more sense if this workflow was triggered by CI, perhaps together with a kedro-datasets release.

Other than that, on the script itself, `_is_union_annotation`, `_json_type_for_union`, `_json_type_for_concrete`, `_resolved_type_hints`, and the `_TYPE_MAP` dict all exist to map resolved Python type objects to JSON types. Bu the actual code only ever produces annotations as strings, so `_json_type_for` always takes the `isinstance(annotation, str)` branch into `_json_type_for_annotation_text/_NAME_TYPE_MAP`. The object-based path is never reached by `main()/write_schema()/check_schema()`. This is directly called by tests but is not actually going to be used when you run the script.

I'd take a look at that, and also don't forget to add the changes to the RELEASE.md.

JokeGbenro:
Hi @lrcouto, thanks for pointing this out.

I’ve addressed the feedback by simplifying the generator around its actual source-based workflow and removing the unreachable runtime type-resolution logic. I also added a CI check to verify that the generated schema is up to date; this will run as part of the existing Kedro-Datasets release validation workflow.

I updated the related tests and added the change to `RELEASE.md`. The focused tests and Ruff checks are passing.

kedro-datasets/RELEASE.md
```diff
@@ -1,14 +1,7 @@
...
 ## Major features and improvements
-## Breaking changes
-## Breaking changes to experimental datasets
-## Bug fixes and other changes
-## Community contributions
-
-# Release 9.6.0
-
-## Major features and improvements
+- Added source-based generation and CI validation for the Kedro catalog JSON schema.
```
ravi-kumar-pilla, line 4: Looks like the `# Release 9.6.0` header got dropped during the merge from main. On `main`, the vectorstore and experimental dataset entries live under their own 9.6.0 section. This diff folds all of that into Upcoming release alongside your new schema bullet.

Your new line belongs under Upcoming. The 9.6.0 content should stay where it was.

.github/workflows/kedro-datasets.yml
```diff
@@ -79,6 +79,27 @@ jobs:
...
+          version: "latest"
+          python-version: "3.10"
+          activate-environment: true
+      - name: Install dependencies
+        run: |
+          uv pip install "kedro @ git+https://github.com/kedro-org/kedro@main"
+          uv pip install -e kedro-datasets
+      - name: Check generated catalog schema
+        working-directory: kedro-datasets
+        run: python -m kedro_datasets._generate_catalog_schema --check
```
ravi-kumar-pilla, line 105: The new `check-schema` job is a good idea, but it is currently failing on CI (`Check generated catalog schema` step). So the guard is in place but red on this branch.

Worth running locally from `kedro-datasets/` and committing whatever it outputs:

```bash
python -m kedro_datasets._generate_catalog_schema
python -m kedro_datasets._generate_catalog_schema --check
```

If `--check` still fails after regenerating, there may be an environment difference between local and CI worth tracking down.

kedro-datasets/tests/test_generate_catalog_schema.py
```diff
@@ -0,0 +1,322 @@
...
+    assert generator.check_schema() is False
+    generator.write_schema()
+    assert schema_path.read_text(encoding="utf-8") == '{\n  "type": "object"\n}\n'
+    assert generator.check_schema() is True
+
+
+def test_schema_path_points_to_static_schema_file():
+    assert generator._schema_path().name == generator.SCHEMA_FILENAME
+    assert generator._schema_path().parent.name == "jsonschema"
+
```
ravi-kumar-pilla, line 332: The unit tests cover the pieces well. One thing missing is a smoke test against the real package. Everything here mocks `_iter_subpackages` or `_iter_dataset_specs`, so drift between source and the committed JSON only shows up in CI.

A simple end to end check would help contributors catch that locally:

```python
def test_committed_schema_matches_generator():
    assert generator.check_schema()
```

That is basically what CI runs. Failures would mean someone needs to regenerate and commit the JSON.

kedro-datasets/kedro_datasets/_generate_catalog_schema.py
```diff
@@ -0,0 +1,436 @@
...
+_NAME_TYPE_MAP: dict[str, str] = {
+    "str": "string",
+    "bool": "boolean",
+    "int": "integer",
+    "float": "number",
+    "dict": "object",
+    "list": "array",
+    "tuple": "array",
+    "Version": "object",
+    "PathLike": "string",
```
ravi-kumar-pilla, line 50: Small one. `PathLike` is in the map but `os.PathLike` is not. Annotations like `os.PathLike[str]` strip down to `os.PathLike`, which falls through to `{"pattern": ".*"}`. Unions like `str | os.PathLike` work because the member resolves to `PathLike`.

If you want those params typed as strings in the schema:

```python
_NAME_TYPE_MAP: dict[str, str] = {
    ...
    "PathLike": "string",
    "os.PathLike": "string",
    ...
}
```

Not blocking. Just tighter validation for PathLike annotations written with the full module prefix.

kedro-datasets/kedro_datasets/_generate_catalog_schema.py
```diff
@@ -0,0 +1,436 @@
+"""Generate the versioned Kedro catalog JSON schema by introspection.
+
+Run as a module::
+
+    python -m kedro_datasets._generate_catalog_schema          # write the schema
```
ravi-kumar-pilla, line 5: The module docstring already explains how to run this, which is good. Since the pre commit hook is manual stage only, contributors might not know they need to regenerate after changing a dataset `__init__`. CI will catch it, but a one line note in the kedro datasets contributing docs or README would save a failed CI round trip.

Something like: after changing a dataset constructor, run `python -m kedro_datasets._generate_catalog_schema` from `kedro-datasets/`.

merelcht:
Hi @JokeGbenro, do you need any help addressing the comments? Let us know when it's ready for review again 🙂

SajidAlamQB:
Nice work @JokeGbenro one thing `_dataset_spec_from_source` falls back to an empty spec when the class doesn't define `__init__` itself. That hits the four huggingface filesystem datasets (they inherit from FilesystemDataset in huggingface/_base.py), so their schema entries are now {"properties": {}} and we silently lose required: ["path"]. Could you resolve base classes via AST, or add those four to `SCHEMA_OVERRIDES` and make the empty-spec fallback warn/fail so the next case doesn't slip through?

JokeGbenro:
Hi, I’m on it and will share an update on the PR soon.

JokeGbenro:
Hi @SajidAlamQB, thanks for catching that — you're right, those four HuggingFace datasets were slipping through with empty properties and losing `required: ["path"]`.

I've fixed it by adding them to `SCHEMA_OVERRIDES` and making the fallback fail loudly with a clear `ValueError` instead of going silent, so the next case like this won't slip through. I went with overrides over AST base-class resolution to keep things explicit and simple.

The regenerated schema now shows the full entries, and `--check` plus the 18 tests are all passing. Branch is updated — happy for you to take another look when you get a chance!

PR #1483, opened 2026-08-07
build(datasets): fix redirected links and speed up lychee CI check

merelcht:
**TL;DR:** lychee now checks links by scanning source files (`.py`/`.yml`/`.md`/`.html`) directly instead of crawling the built docs `site/`, which is what makes it faster — not because fewer links are covered.

### Description
Fixed lychee redirects where it makes sense, and reworked the `check-docs` lychee setup so it doesn't take ~50 mins to run (original slow run: https://github.com/kedro-org/kedro-plugins/actions/runs/31106737864/job/92633738389?pr=1477).

### Changes
- Updated links in `kedro-datasets` Markdown files that resolved via one or more HTTP redirects to point at their final destination instead.
- Reworked the `check-docs` lychee step: it now scans source `.py`/`.yml`/`.md` files directly instead of the built `site/`. The built site duplicates every docstring link across an `.html` and `.md` page per dataset, plus repeats it again on every page's shared nav/footer — scanning source checks each link once instead of once per generated page.
- Excludes `tests/` (fixture code uses `file:///tmp/...` paths that aren't real links).
- Added `kedro-datasets/lychee.toml`: accepts `429` as a valid status instead of retrying it, and excludes loopback addresses and known placeholder hostnames used in docstring example credentials (`localhost:*`, `custom.langfuse.com`, `*.weaviate.network`) that source-scanning surfaces but the old HTML-based scan didn't (lychee skips links inside code blocks by default). Earlier iterations also added per-host concurrency throttling for the old, much higher-volume site scan; removed once the source-scan switch made it unnecessary (confirmed via a real run: 0 rate-limit responses at the new, much lower link volume).
- Also added `.html` to the scanned extensions, so the hardcoded social links in `docs/overrides/*.html` (rendered on every page's footer) stay covered — they'd otherwise have been silently dropped by the switch to source-scanning.

### Result
The latest run took 1 min and summary is as follows compared to the one that took 50 mins:
| Status | Original (~50 min) | Latest (1m 41s) |
|---|---|---|
| 🔍 Total | 18168 | 834 |
| 🔗 Unique | 2074 | 422 |
| ✅ Successful | 16526 | 745 |
| 🔀 Redirected | 167 | 41 |
| 🚫 Errors | 1166 | 22 |
| ⏳ Timeouts | 0 | 1 |
| 👻 Excluded | 1 | 20 |

### DIFF
--- a/kedro-datasets/kedro_datasets_experimental/mlrun/README.md
+++ b/kedro-datasets/kedro_datasets_experimental/mlrun/README.md
@@ -286,5 +286,5 @@ FileNotFoundError: Model artifact not found
 
 - [MLRun Documentation](https://docs.mlrun.org/en/latest/?badge=latest) — check your MLRun version for supported `load_args`/`save_args`
 - [MLRun execution API (MLClientCtx)](https://docs.mlrun.org/en/latest/api/mlrun.execution/index.html) — `log_artifact`, `log_model`, `log_result`, `get_artifact`
-- [Kedro Documentation](https://docs.kedro.org/en/stable/)
+- [Kedro Documentation](https://docs.kedro.org/)
 - [MLRun GitHub](https://github.com/mlrun/mlrun)
--- a/kedro-datasets/kedro_datasets_experimental/mlrun/README.md
+++ b/kedro-datasets/kedro_datasets_experimental/mlrun/README.md
@@ -286,5 +286,5 @@ FileNotFoundError: Model artifact not found
 
 - [MLRun Documentation](https://docs.mlrun.org/en/latest/?badge=latest) — check your MLRun version for supported `load_args`/`save_args`
 - [MLRun execution API (MLClientCtx)](https://docs.mlrun.org/en/latest/api/mlrun.execution/index.html) — `log_artifact`, `log_model`, `log_result`, `get_artifact`
-- [Kedro Documentation](https://docs.kedro.org/)
+- [Kedro Documentation](https://docs.kedro.org/en/stable/)
 - [MLRun GitHub](https://github.com/mlrun/mlrun)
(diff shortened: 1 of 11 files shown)


kedro-datasets/kedro_datasets_experimental/mlrun/README.md
```diff
@@ -286,5 +286,5 @@ FileNotFoundError: Model artifact not found
 
 - [MLRun Documentation](https://docs.mlrun.org/en/latest/?badge=latest) — check your MLRun version for supported `load_args`/`save_args`
 - [MLRun execution API (MLClientCtx)](https://docs.mlrun.org/en/latest/api/mlrun.execution/index.html) — `log_artifact`, `log_model`, `log_result`, `get_artifact`
-- [Kedro Documentation](https://docs.kedro.org/)
+- [Kedro Documentation](https://docs.kedro.org/en/stable/)
```
merelcht:
```suggestion
- [Kedro Documentation](https://docs.kedro.org/)
```

kedro-datasets/kedro_datasets_experimental/opik/README.md
```diff
@@ -1064,7 +1064,7 @@ def my_task(dataset_item: dict) -> dict:
 
 ## Related Resources
 - **Opik Documentation**: [Opik Platform](https://www.comet.com/docs/opik/)
-- **Opik Tracing**: [Tracing overview](https://www.comet.com/docs/opik/tracing/log_traces)
+- **Opik Tracing**: [Tracing overview](https://www.comet.com/docs/opik/tracing/advanced/log_traces)
 - **Opik Evaluation**: [Evaluation overview](https://www.comet.com/docs/opik/evaluation/overview)
-- **Kedro Documentation**: [Kedro Datasets](https://docs.kedro.org/)
+- **Kedro Documentation**: [Kedro Datasets](https://docs.kedro.org/en/stable/)
```
merelcht:
```suggestion
- **Kedro Documentation**: [Kedro Datasets](https://docs.kedro.org/)
```

SajidAlamQB (approved):
Thank you @merelcht

ElenaKhaustova (approved):
Thank you, @merelcht!

Issue #1316, opened 2026-02-17
Ensure that `os.PathLike` objects work for all datasets


merelcht:
## Description
Testing that `os.PathLike objects` work and fix if needed.

The unit tests currently cast paths to string, but the doctests getting included as part of https://github.com/kedro-org/kedro/issues/2287 are probably doing a fair job of ensuring this.

celinaczy:
Hi, I'd also like to contribute here. I'll take a look at 
Holoviews
Huggingface
Ibis
JSON
Langchain

tannnmayy:
I'd like to take **Ibis** for this issue.

Opened a PR adding `os.PathLike` support for `ibis.FileDataset` (type hint + docstring + pathlike round-trip test): https://github.com/kedro-org/kedro-plugins/pull/1487

tannnmayy:
I'd like to take **Email** and **Geopandas** for this issue as well.

- Email: https://github.com/kedro-org/kedro-plugins/pull/1488 (or see latest PR from @tannnmayy)
- Geopandas: pathlike regression tests, same pattern as #1487

tannnmayy:
Following up with concrete PRs for **Email** and **Geopandas** (pathlike regression tests only; type hints already present):

- Email: https://github.com/kedro-org/kedro-plugins/pull/1488
- Geopandas: https://github.com/kedro-org/kedro-plugins/pull/1489

tomatotomata:
Hi @merelcht, I saw #1316 still lists Dask as to be done and I could not find an open Dask PR. I was thinking of taking only that slice: add `pathlib.Path` save/load regression coverage for `CSVDataset` and `ParquetDataset`, keeping the implementation unchanged if the existing fsspec calls already handle PathLike values. I would mirror the existing dataset test pattern and run the focused Dask tests, while noting the repository's current pandas compatibility skip.

@juanchodpg2 mentioned Dask in an older comment, so I do not want to overlap with work they are still doing. Is the Dask slice still available, or should I leave it to them?

merelcht:
Hi @tomatotomata, as far as I'm aware Dask hasn't been done yet 🙂

nightcityblade:
I am taking the still-unchecked Partitions subtask. I will add `os.PathLike` support to the shared `PartitionedDataset` path handling (which also covers `IncrementalDataset`), update both public annotations/docs, and add focused regression coverage.

PR #1491, opened 2026-08-10
test: cover pathlib paths in Dask datasets

tomatotomata:
## What does this change do?

Adds focused regression coverage showing that the Dask `CSVDataset` and `ParquetDataset` accept `pathlib.Path` filepaths for local save/load operations. The tests exercise the existing implementation without changing dataset behavior.

## How was it tested?

- `git diff --check` passes.
- The two focused pytest cases are blocked during test collection because this checkout is missing the `aiobotocore` dependency required by `tests/conftest.py`.

### DIFF
--- a/kedro-datasets/kedro_datasets/dask/csv_dataset.py
+++ b/kedro-datasets/kedro_datasets/dask/csv_dataset.py
@@ -107,18 +107,21 @@ class CSVDataset(AbstractDataset[dd.DataFrame, dd.DataFrame]):
         }
 
     def load(self) -> dd.DataFrame:
+        filepath = os.fspath(self._filepath)
         return dd.read_csv(
-            self._filepath, storage_options=self.fs_args, **self._load_args
+            filepath, storage_options=self.fs_args, **self._load_args
         )
 
     def save(self, data: dd.DataFrame) -> None:
-        data.to_csv(self._filepath, storage_options=self.fs_args, **self._save_args)
+        filepath = os.fspath(self._filepath)
+        data.to_csv(filepath, storage_options=self.fs_args, **self._save_args)
 
     def _exists(self) -> bool:
-        protocol = get_protocol_and_path(self._filepath)[0]
+        filepath = os.fspath(self._filepath)
+        protocol = get_protocol_and_path(filepath)[0]
         file_system = fsspec.filesystem(protocol=protocol, **self.fs_args)
         try:
-            files = file_system.glob(self._filepath)
+            files = file_system.glob(filepath)
         except FileNotFoundError:
             return False
         return bool(files)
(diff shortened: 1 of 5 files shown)


tomatotomata:
I followed the first CI failure through the code rather than just adjusting the assertion. The CSV dataset now normalizes `PathLike` values with `os.fspath()` at the fsspec/Dask load, save, and exists boundaries, and Parquet uses the same boundary for consistency.

I also made the CSV round-trip assertion tolerant of the expected pandas string-dtype difference, applied the repository Black formatting, and refreshed the two shifted detect-secrets baseline line numbers. The follow-ups are on `69932fd`, `a379972`, and `39d78c0`; DCO is passing and lint has passed on the current run. The remaining unit-test matrix is still running, while detect-secrets is still reporting a baseline issue that I’m tracing from its completed log. Let me know what you think.

kedro-datasets/kedro_datasets/dask/csv_dataset.py
```diff
@@ -107,18 +107,19 @@ def _describe(self) -> dict[str, Any]:
         }
 
     def load(self) -> dd.DataFrame:
-        return dd.read_csv(
-            self._filepath, storage_options=self.fs_args, **self._load_args
-        )
+        filepath = os.fspath(self._filepath)
```
merelcht: Is it necessary to do the wrapping of filepath in `os.fspath()`? Most solutions for https://github.com/kedro-org/kedro-plugins/issues/1316 didn't require any changes to the datasets, only to the tests.

merelcht:
Hi @tomatotomata, thanks for starting this contribution. Can you address the review comments?

tomatotomata:
That makes sense. I rechecked the Dask boundaries and removed the production `os.fspath()` wrapping, keeping this contribution focused on the PathLike regression coverage in the two dataset test modules.

The cleanup is on `2a7ad57` with the existing DCO sign-off. `git diff --check` and Python bytecode compilation pass. The two focused pytest cases are blocked in this checkout because the test environment is missing `aiobotocore`; the refreshed CI run is in progress.

Do you think this matches the test-only pattern you had in mind?

merelcht:
Hi @tomatotomata, I'm getting the sense I may be talking to an LLM rather than a person. We do accept LLM-assisted contributions, but we ask that contributors engage with us as humans.

We put real effort into curating good first issues so that people new to open source have a welcoming place to start, and some are deliberately simple to help people learn. We'd like to keep those interactions human. Our maintainers also spend significant time reviewing and managing contributions, so please be mindful of that.

I'm closing this PR to give a human the chance to contribute instead. 

If I've misread the situation and you are a person, I apologise. Feel free to re-open the PR and I'm happy to work with you to get your contribution merged.

PR #1423, opened 2026-06-09
feat(datasets): Add `os.PathLike` support for Spark datasets

GDaamn:
## Description
This PR closes part of issues #1316 #1317 by adding support for `os.PathLike` objects as the filepath argument to the Spark datasets, in addition to strings. The PR also adds related tests for this change.

## Development notes
- Updated the filepath parameter type from str to str | os.PathLike for:
    * `SparkDataset`
    * `SparkStreamingDataset`
    * `DeltaTableDataset`
- Updated the corresponding docstrings to document that filepath can be either a `str` or an `os.PathLike` object.
- Added tests verifying that pathlib.Path objects can be used successfully:
    * `test_pathlike_filepath` for SparkDataset
    * `test_pathlike_filepath` for SparkStreamingDataset
    * `test_pathlike_filepath` for DeltaTableDataset

## Checklist

- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Updated `jsonschema/kedro-catalog-X.XX.json` if necessary
- [ ] Added a description of this change in the relevant `RELEASE.md` file
- [ x] Added tests to cover my changes
- [ ] Received approvals from at least half of the TSC (required for adding a new, non-experimental dataset)

PR #1493, opened 2026-08-11
feat(datasets): add os.PathLike support for SparkDatasetV2

JokeGbenro:
Fixes part of #1316.

### DIFF
--- a/kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
+++ b/kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
@@ -124,7 +124,7 @@ class SparkDatasetV2(AbstractVersionedDataset):
     def __init__(  # noqa: PLR0913
         self,
         *,
-        filepath: str,
+        filepath: str | os.PathLike,
         file_format: str = "parquet",
         load_args: dict[str, Any] | None = None,
         save_args: dict[str, Any] | None = None,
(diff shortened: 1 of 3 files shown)


kedro-datasets/kedro_datasets/spark/spark_dataset_v2.py
```diff
@@ -170,6 +171,9 @@ def __init__(  # noqa: PLR0913
             metadata: Any arbitrary metadata.
                 This is ignored by Kedro, but may be consumed by users or external plugins.
         """
+        # Normalise PathLike to string so helpers can call str methods
+        filepath = str(filepath)
```
merelcht: Why does the filepath need to be normalised as string? The aim of this task is to test that `os.PathLike` objects work as well, ideally without any string casting.

merelcht (changes requested):
Hi @JokeGbenro, thanks for the contribution! 🙂 I left a question on the normalisation.

JokeGbenro:
Thanks for the review, @merelcht. You're right — casting to `str` in `__init__` meant the `os.PathLike` object was never actually exercised.

I've reworked it in `42968f5`:
- Removed the `filepath = str(filepath)` normalisation from `SparkDatasetV2.__init__`; the original `os.PathLike` value is now kept as-is.
- `parse_spark_filepath` and `validate_databricks_path` (the only helpers that call string methods like `.startswith()`) now accept `str | os.PathLike` and normalise internally via `os.fspath`.
- The test now also asserts `dataset._filepath` is a `PurePosixPath` and round-trips to `filepath.as_posix()`, matching the `SparkDataset` PathLike test in #1423.

The PathLike value now flows all the way through `SparkDatasetV2` without any upfront string casting.

merelcht (approved):
Thanks @JokeGbenro, this looks good to me now 👍

SajidAlamQB (approved):
Perfect thank you @JokeGbenro

PR #1503, opened 2026-08-26
feat(datasets): add Apache Iceberg dataset support for polars

### LINKED ISSUE #1030: `kedro-datasets`: Add dataset for Iceberg Tables
## Description
Add native support for Iceberg tables to `kedro-datasets`

Follow up from documenting Kedro + Iceberg in https://github.com/kedro-org/kedro/pull/4521

## Context
We don't have any datasets that support Iceberg tables, the documentation added in https://github.com/kedro-org/kedro/pull/4521 is fairly minimal and has limitations:
- Only works for `pandas` dataframes
- Works with `pyiceberg` behind the scenes, which doesn't support the full range of features you can leverage for Iceberg tables
- Is custom implementation

I also want to get more feedback from the community about what level of support and features they would expect from this/these dataset/s. Also, would like to hear from users about how the currently use Iceberg tables with Kedro.

## Possible Implementation
- Extend the custom example from docs
- Use other libraries in the backend
- Spark + Iceberg dataset

Saurav-Gupta-9741:
## Description & Motivation
Closes #1030.

This PR introduces native **Apache Iceberg** dataset support for **Polars** in `kedro-datasets`:
- `polars.IcebergDataset` (`kedro_datasets.polars.IcebergDataset`)

### Key Capabilities:
- **Polars Native Delegation**: Prioritizes `polars.scan_iceberg(table, **load_args).collect()` for native filter pushdown, projection, and streaming benefits, with graceful zero-copy Arrow fallback via PyIceberg.
- **Catalog & Storage Support**: Seamless integration with REST, AWS Glue, Hive, and SQL catalogs via `pyiceberg.catalog.load_catalog()`.
- **Secure Credentials Handling**: Supports `credentials` parameter for catalog and storage authentication (e.g. tokens, access keys) while safely excluding credentials from `_describe()`.
- **Atomic Commits & Schema Enforcement**: Supports both `overwrite` (default) and `append` commit modes for eager `pl.DataFrame` and `pl.LazyFrame`.

## How Has This Been Tested?
- Added 12 mock-based unit tests covering:
  - Catalog instantiation, loading, and error handling
  - Credentials merging and exclusion from `_describe()`
  - Saving eager DataFrame and LazyFrame in overwrite/append modes
  - Table creation when not existing
  - Specific exception handling in `_exists()` (`NoSuchTableError`, `NoSuchNamespaceError`)
  - Graceful dependency checks for optional dependencies
- Formatted and linted with `ruff`.

## Checklist
- [x] Tested changes locally with pytest
- [x] Linting and code formatting checks pass (`ruff check`, `ruff format`)
- [x] Added docstrings with YAML and Python catalog examples
- [x] Added release notes entry in `RELEASE.md`


deepyaman:
@Saurav-Gupta-9741 please reduce this to just one library--which library are you primarily working with?

@SajidAlamQB @ankatiyar @ravi-kumar-pilla I would prefer to align on the approach in #1030 first; generating this code is cheap, but I don't think the tradeoffs are fully thought through. (@Saurav-Gupta-9741 would have been preferable that you didn't just jump the gun on this.)

Saurav-Gupta-9741:
Hi @deepyaman Sir
Thank you for the feedback and guidance! I apologize for jumping ahead with both libraries at once.
Following your recommendation, I have reduced this PR to focus exclusively on **Polars** (`polars.IcebergDataset`), which provides native zero-copy Arrow integration and aligns with the direction discussed in #1030.
All 59 CI checks and unit tests are green. I'm very happy to iterate on the API design or discuss any specific tradeoffs you'd like to refine!

ravi-kumar-pilla:
Hi @Saurav-Gupta-9741 ,

Thanks for putting this together, and for narrowing it down to Polars after Deepyaman’s feedback. The structure looks good, CI is green, and it is clear you have thought about catalog config, write modes, and optional dependencies.

Before we merge though, I want to align with the direction Deepyaman raised in [#1030](https://github.com/kedro-org/kedro-plugins/issues/1030). I think the goal should be a thin Kedro wrapper around Polars native Iceberg APIs, not a PyIceberg shim inside `polars.IcebergDataset`.  @ankatiyar do you agree on that ?

---

Right now `_load()` goes through `load_catalog()` → `table.scan()` → `to_polars()`, and `_save()` goes through Arrow → append/overwrite. That is essentially the same pattern as the custom PyIceberg dataset in our docs, just moved into `kedro-datasets`. The approach we discussed on the issue was closer to:

**Load:** delegate to `pl.scan_iceberg(table_identifier, catalog=..., **load_args)` and collect if we want an eager DataFrame

**Save:** delegate to `data.to_iceberg(...)` or `lazy_frame.sink_iceberg(...)` with the appropriate mode

That keeps filter pushdown, streaming, and future Polars Iceberg improvements on the native path instead of maintaining a parallel PyIceberg layer in Kedro.

---

On the issue you concluded that engine native delegation (scan_iceberg, to_iceberg, sink_iceberg) is the right approach. The current PR goes back to a PyIceberg direct implementation instead. Can you explain what changed? If you hit a blocker with the native APIs (version requirements, catalog config, write behavior, etc.) please share that here. If not, I think we should refactor to match what you proposed on #1030.

---

**A few other things to tidy up regardless of the above:**

- PR title and description still mention pandas and 18 tests. Please update to reflect Polars only and the actual test count.

- Docstring / YAML example: the example uses load_args: columns: [...], but PyIceberg scan uses selected_fields. If we go native, load args should match what scan_iceberg accepts. Worth double checking the example matches the final API.

- `RELEASE.md`: please add an entry under upcoming release for `polars.IcebergDataset` and credit yourself under community contributions.

- Tests: mock tests are a good start. It would help to have at least one integration test with PyIceberg’s SqlCatalog and SQLite, similar to what you outlined on the issue. That gives more confidence than mocks alone.

- `_exists()`: catching all exceptions and returning False makes a bad catalog config look the same as "table does not exist". Worth letting real connection or auth errors surface.

Let me know if refactoring to native Polars APIs raises any concerns on your side. Thanks again for the contribution.

ravi-kumar-pilla (commented):
Left some comments here - https://github.com/kedro-org/kedro-plugins/pull/1503#issuecomment-5460043443

Saurav-Gupta-9741:
Hi @ravi-kumar-pilla Sir,

Thank you so much for the detailed and thoughtful review! I have addressed all your points and updated the PR accordingly:

### 1. Polars Native Delegation (`_load` & `_save` Tradeoffs)
- **`_load()`**: Refactored to prioritize Polars native `pl.scan_iceberg(table, **load_args).collect()` whenever available. This ensures native predicate pushdown, projection, and streaming benefits. If native scan encounters an engine-specific mismatch or is unavailable in older Polars versions, it falls back to PyIceberg's `table.scan().to_polars()` / Arrow zero-copy path.
- **`_save()`**: We evaluated `sink_iceberg` / `to_iceberg`. Currently, native Polars write APIs (`sink_iceberg`) are experimental/in active development across Polars versions and depend on specific catalog backend configurations. Writing via PyIceberg's Arrow interface (`table.append()` / `table.overwrite()`) provides stable snapshot isolation, atomic commits, partition spec compliance, and schema enforcement across all catalog types (Glue, REST, SQL, etc.).

### 2. Specific Exception Handling in `_exists()`
- Updated `_exists()` to catch only `NoSuchTableError` and `NoSuchNamespaceError`. Genuine authentication, networking, or configuration errors (e.g. `ConnectionError`, `PermissionError`) will now surface directly instead of silently returning `False`.

### 3. Docstrings & YAML Examples
- Updated docstring and YAML examples to use `selected_fields` instead of `columns`, matching PyIceberg and scan API conventions.

### 4. `RELEASE.md` & PR Metadata
- Added `polars.IcebergDataset` entry under Major Features in `RELEASE.md` and credited community contribution.
- Updated the PR title and description to reflect the Polars-focused scope and updated test suite.

All **59 CI checks** are currently green and passing across Python 3.10–3.14 on both Linux and Windows. Looking forward to your review!

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,199 @@
...
+
+    """
+
+    DEFAULT_WRITE_MODE = "overwrite"
+    ACCEPTED_WRITE_MODES = ("overwrite", "append")
+
+    DEFAULT_LOAD_ARGS: dict[str, Any] = {}
+    DEFAULT_SAVE_ARGS: dict[str, Any] = {"mode": DEFAULT_WRITE_MODE}
+
+    def __init__(  # noqa: PLR0913
```
ankatiyar, line 63: Thanks @Saurav-Gupta-9741 for the PR, I noticed that there is no credentials argument and no credential handling. How is authentication handled for this dataset? If it is through `catalog_properties` then any keys will be visible in the `_describe()` which might be a security issue!

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,199 @@
...
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
+        """Creates a new instance of ``IcebergDataset`` pointing to an Apache Iceberg table.
+
+        Args:
+            table_name: Table identifier (e.g. ``"namespace.table_name"`` or ``"table_name"``).
+            catalog_name: Name of the Iceberg catalog to load. Defaults to None.
+            catalog_properties: Properties required to instantiate the catalog (e.g.
+                ``{"type": "rest", "uri": "https://..."}`` or ``{"type": "glue"}``).
+            load_args: Additional scan/read options such as ``selected_fields``,
```
ankatiyar: Double check if these arguments exist in `polars.scan_iceberg` function - i think there's some mis match

ankatiyar:
Could you also update the PR description to just focus on the one dataset?

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Thanks @ankatiyar Ma'am! I have added dedicated `credentials` argument handling in `__init__`. 

Credentials (e.g. tokens, AWS/GCP keys from `credentials.yml`) are now stored separately in `self._credentials` and merged into catalog properties when connecting to the catalog in `_get_catalog()`. They are completely excluded from `_describe()` so secrets will not leak in descriptions or logs. I've also added unit tests covering this behavior in commit `c34cfd2`.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: I have updated the docstring and YAML examples. `load_args` now explicitly documents scan options accepted by `polars.scan_iceberg` (such as `snapshot_id`, `storage_options`, `reader_override`) as well as PyIceberg scan fallback options (`selected_fields`, `row_filter`, `limit`).

ravi-kumar-pilla (commented):
Thanks for the updates. Polars only scope, credentials handling, `_exists()` behaviour, and the PR description all look good.

Two things to fix before merge:

1. Module and class docstrings overstate Polars native coverage. Save goes through PyIceberg, not Polars write APIs.
2. `load_args` mixes Polars and PyIceberg scan options in one dict. Invalid keys on the native path get silently swallowed.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,204 @@
+"""``IcebergDataset`` loads and saves data from/to Apache Iceberg tables
```
ravi-kumar-pilla: The module and class docstrings say load and save use Polars native APIs. That is only partly true.

Load tries `scan_iceberg` but falls back to PyIceberg. Save does not use `write_iceberg` or `sink_iceberg` today. It goes through PyIceberg directly.

Please update both docstrings. Something like:

```python
"""``IcebergDataset`` loads and saves Apache Iceberg tables.

Loads via Polars ``scan_iceberg`` when available, with a PyIceberg scan fallback.
Saves commit data to Iceberg tables. Catalog access is handled through PyIceberg.
"""
```

Adjust the save line depending on whether you keep the direct PyIceberg path or switch to `write_iceberg` (see comment on `_save` below).

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Args:
+            table_name: Table identifier (e.g. ``"namespace.table_name"`` or ``"table_name"``).
+            catalog_name: Name of the Iceberg catalog to load. Defaults to None.
+            catalog_properties: Properties required to instantiate the catalog (e.g.
+                ``{"type": "rest", "uri": "https://..."}`` or ``{"type": "glue"}``).
+            credentials: Authentication credentials or secrets (e.g. tokens, AWS/GCP keys).
+                These are merged with ``catalog_properties`` when connecting to the catalog
+                and are safely excluded from ``_describe()`` to avoid leaking secrets.
+            load_args: Additional scan/read options passed to ``polars.scan_iceberg``
```
ravi-kumar-pilla: `load_args` documents options for two different APIs in one dict.

`polars.scan_iceberg` accepts `snapshot_id`, `storage_options`, `reader_override`, etc. PyIceberg `table.scan` accepts `selected_fields`, `row_filter`, `limit`. These are not interchangeable.

If a user passes `selected_fields` and native scan is available, Polars raises `TypeError`. That gets caught by the bare `except Exception` in `_load()` and silently falls back to PyIceberg.

I would restrict `load_args` to Polars native keys only. Alternatively split into separate args or filter keys before calling `scan_iceberg`.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+
+        catalog = self._get_catalog()
+        table = catalog.load_table(self._table_name)
+
+        # Prefer native Polars Iceberg scan when available
+        if hasattr(pl, "scan_iceberg"):
+            try:
+                lazy_df = pl.scan_iceberg(table, **self._load_args)
+                return lazy_df.collect()
+            except Exception:  # noqa: BLE001
```
ravi-kumar-pilla: The bare `except Exception: pass` in `_load()` can hide real errors. Bad `snapshot_id`, auth failures, and invalid `load_args` all get swallowed.

Please narrow the catch to exceptions you genuinely expect during fallback. Re raise on `TypeError` and `ValueError` from bad user args.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Fallback: PyIceberg scan → to_polars or Arrow zero-copy
+        scan = table.scan(**self._load_args)
+        if hasattr(scan, "to_polars"):
+            return scan.to_polars()
+
+        arrow_table = scan.to_arrow()
+        return pl.from_arrow(arrow_table)
+
+    def _save(self, data: pl.DataFrame | pl.LazyFrame) -> None:
```
ravi-kumar-pilla, line 159: For save, consider `data.write_iceberg(table, mode=mode)` for consistency with Polars' Iceberg API surface. Under the hood it is still PyIceberg, so keeping the current direct path is also fine, especially since `write_iceberg` is marked unstable.

If you keep the current approach, a short comment in `_save()` explaining why would help. Update the docstring save line to match whichever approach you go with.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+        load_args: dict[str, Any] | None = None,
+        save_args: dict[str, Any] | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
+        """Creates a new instance of ``IcebergDataset`` pointing to an Apache Iceberg table.
+
+        Args:
+            table_name: Table identifier (e.g. ``"namespace.table_name"`` or ``"table_name"``).
+            catalog_name: Name of the Iceberg catalog to load. Defaults to None.
+            catalog_properties: Properties required to instantiate the catalog (e.g.
```
ravi-kumar-pilla, line 79: `credentials` are correctly excluded from `_describe()`. Secrets in `catalog_properties` will still show up in `_describe()`. Worth a note on the `catalog_properties` arg: do not put secrets here, use `credentials` instead.

kedro-datasets/tests/polars/test_iceberg_dataset.py
```diff
@@ -0,0 +1,208 @@
+import sys
```
ravi-kumar-pilla, line 1: Mock tests are a good start. No explicit test for the native `scan_iceberg` path. `test_load_pyiceberg_scan` may hit the fallback depending on Polars version.

PR description says 12 tests, file has 13.

Makefile
```diff
@@ -81,6 +81,7 @@ dataset-doctest%:
 	  --ignore kedro_datasets/huggingface/transformer_pipeline_dataset.py \
 	  --ignore kedro_datasets/pandas/gbq_dataset.py \
 	  --ignore kedro_datasets/partitions/partitioned_dataset.py \
+	  --ignore kedro_datasets/polars/iceberg_dataset.py \
```
ravi-kumar-pilla, line 84: Please add a short comment on the `iceberg_dataset.py` ignore entry explaining why (optional deps not available in doctest env).

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Updated both module and class docstrings to accurately reflect: loads via scan_iceberg with PyIceberg fallback, saves via PyIceberg catalog operations.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Scoped load_args docstring to Polars native scan_iceberg arguments only. In the fallback path, only compatible keys (e.g. snapshot_id) are forwarded to table.scan().

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Narrowed to except (NotImplementedError, AttributeError). TypeError/ValueError from bad user args now propagate directly. Added test_load_invalid_load_args_raises_type_error to verify.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Added inline comment explaining why we use PyIceberg's Arrow interface directly rather than unstable write_iceberg/sink_iceberg.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Added note: "Do not pass secrets/tokens here as they are visible in _describe(); use credentials instead."

kedro-datasets/tests/polars/test_iceberg_dataset.py
Saurav-Gupta-9741: Added test_load_native_scan_iceberg, test_load_pyiceberg_scan_fallback, test_load_invalid_load_args_raises_type_error, and test_missing_polars_raises_error. Total: 16 tests.

Makefile
Saurav-Gupta-9741: Updated comment to mention optional dependencies (pyiceberg) as reason for the ignore.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+        # Prefer native Polars Iceberg scan when available
+        if hasattr(pl, "scan_iceberg"):
+            try:
+                lazy_df = pl.scan_iceberg(table, **self._load_args)
+                return lazy_df.collect()
+            except (NotImplementedError, AttributeError):
+                # Fall back to PyIceberg scan if native scan is not supported for this table format/engine
+                pass
+
+        # Fallback: PyIceberg scan -> to_polars or Arrow zero-copy
```
deepyaman, line 147: I don't like this. Datasets are simple wrappers; set the bound for when scan_iceberg was introduced in the dependencies; if the user has an older version, they can't leverage the dataset, and that's fine.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+            scan_args["snapshot_id"] = self._load_args["snapshot_id"]
+
+        scan = table.scan(**scan_args)
+        if hasattr(scan, "to_polars"):
+            return scan.to_polars()
+
+        arrow_table = scan.to_arrow()
+        return pl.from_arrow(arrow_table)
+
+    def _save(self, data: pl.DataFrame | pl.LazyFrame) -> None:
```
deepyaman, line 159: This is not correct.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+        """Saves a Polars DataFrame into the Apache Iceberg table."""
+        if hasattr(data, "collect"):
```
deepyaman, line 161: Do other Polars datasets handle both lazy and eager frames like this?

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+
+        save_args = deepcopy(self._save_args)
+        mode = save_args.pop("mode", self.DEFAULT_WRITE_MODE)
+        location = save_args.pop("location", None)
+        snapshot_properties = save_args.pop("snapshot_properties", None)
+
+        catalog = self._get_catalog()
+        arrow_table = data.to_arrow()
+
+        # We write via PyIceberg's Arrow interface (append / overwrite) directly
```
deepyaman, line 172: Disagree, same reason as above. If you don't feel the sink_iceberg/write_iceberg are ready for use, that should have come out of the discussion on the issue. The Polars Iceberg dataset defers to Polars.

Makefile
```diff
@@ -76,11 +76,12 @@ dataset-doctest%:
 	  exit 2; \
 	fi; \
     \
-	# The ignored datasets below require complicated setup with cloud/database clients or network model download which is overkill for the doctest examples.
+	# The ignored datasets below require complicated setup with cloud/database clients, optional dependencies (e.g. pyiceberg for iceberg_dataset), or network model download which is overkill for the doctest examples.
```
deepyaman, line 79: We don't exclude things from doctests due to optional dependencies? I don't understand what makes it hard for this to work.

deepyaman (changes requested):

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Agreed — removed the fallback entirely. _load() is now a single line: pl.scan_iceberg(table, **self._load_args).collect().

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Removed along with the entire fallback path.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Updated to accept only pl.DataFrame, consistent with EagerPolarsDataset. Removed the hasattr(data, "collect") check.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
Saurav-Gupta-9741: Refactored to use data.write_iceberg(table, mode=mode). Removed all PyIceberg direct write logic and table auto-creation.

Makefile
Saurav-Gupta-9741: Removed the --ignore kedro_datasets/polars/iceberg_dataset.py entry and reverted the comment to its original text. Docstring examples already use # doctest: +SKIP.

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
deepyaman: What was the rationale behind this choice?

kedro-datasets/kedro_datasets/polars/iceberg_dataset.py
deepyaman: You misdiagnosed the issue. Study other dataset implementations; your training data seems to be outdated, and you're relying on AI without having studying other dataset implementations (e.g. pandas, Ibis), or having read the guide on creating a custom dataset.

